PR #17, opened 2016-03-22
Add systems.csv file to act as a registry for systems providing GBFS

jcn:
Based on the list originally from @mplsmitch I'd like to propose this format for our "system registry" CSV file. I like CSV on Github because it's human-readable and searchable. You can see what this file will look like at https://github.com/jcn/gbfs/blob/systems-list/systems.csv

Looking for comments and/or validation from the other contributors. I will merge this into master if I do not get any comments by the end of the week.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,4 +21,4 @@ More information:
 
 * [systems.csv](systems.csv)
 
-If you would like to add a system, please fork this repository and submit a pull request. Please keep this list alphabetized by system name.
+If you would like to add a system, please fork this repository and submit a pull request. Please keep this list alphabetized by country and system name.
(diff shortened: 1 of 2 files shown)


fkh:
+1 to csv.

CSV contributions are easy via github (crazy but true).

On Tue, Mar 22, 2016 at 5:16 PM, Jesse Chan-Norris <notifications@github.com

bcbeta:
Nice! Would be helpful to have geolocation field so it would be easy to find nearby systems.

mplsmitch:
this seems like the ideal place for system_id  from system_information.json - then new systems could check the csv ensure their proposed system_id is globally unique

jcn:
> Nice! Would be helpful to have geolocation field so it would be easy to find nearby systems.

I'm trying to keep the barrier to entry pretty low for this file, and it seems like geolocation might start making that more of a pain, but I'd be open to it if people really wanted it.

> this seems like the ideal place for system_id from system_information.json

I do think this makes a lot of sense - I'll add this, with the caveat that inclusion in this file is not mandatory for systems that implement GBFS, so it's not a guarantee that there are no other GBFS systems out there that might have this identifier.

jcn:
As there have been no additional comments on this, I will be merging this in to master.

Please note that this file is not part of the official spec and as such, I believe modifications to it can be done much easier.

PR #27, opened 2016-05-11
Specify format for WGS84 latitude and longitude

barbeau:
- Require decimal degrees (as opposed to degrees, minutes, and seconds)
- Allows clients to assume the value can be directly mapped to a numeric variable such as an integer, instead of potentially being a String.


jcn:
Is this a functional change or a clarification? i.e. given all of the current implementations from the existing vendors in `systems.csv`, do all of them use the same format right now?

barbeau:
This is a clarification - I just scanned all the `station_information.json` endpoints for the vendors in `systems.csv` to confirm and so far everyone is using decimal degrees.

barbeau:
BTW, is there a GBFS validation tool?

jcn:
> BTW, is there a GBFS validation tool?

There is no GBFS validation tool.

jcn:
This looks good to me.

PR #29, opened 2016-05-12
Add BCycle systems

jasongdove:
These are the BCycle systems that are currently opted into GBFS.


fruminator:
Awesome thanks guys!

PR #32, opened 2016-06-06
Moved field defintions to correct heading

serialc:
Moved field definitions from File Requirements to Field Definitions.
Updated station_id link to Field Definitions.


jcn:
This seems reasonable. Please move the "Time stamp" definition to `Field Definitions` as well since it also defines a data format and then I'll merge this in.

PR #38, opened 2016-07-12
Update system.csv

colinmcglynn:
Add three Social Bicycle systems.
1. VeloGo
2. University of South Florida
3. University of South Florida

PR #39, opened 2016-07-12
Remove duplicate USF entry

barbeau:
Duplicate was added in https://github.com/NABSA/gbfs/pull/38.


jcn:
@barbeau good catch, thanks.

PR #40, opened 2016-07-13
Remove duplicate Apple Pay entry

ghost:
Under the rental_methods for station_information.json, 'APPLEPAY' was listed twice. This commit removes the second occurrence of that payment method.


jcn:
Thanks for catching that.

PR #51, opened 2016-08-24
Update systems.csv

cubbi:
Buffalo bike share rebranding, new location of GBFS file.


jcn:
This brings up an interesting question regarding the spec. The spec indicates that the system identifier not change. 

If this was just a rename of the system (but with things like stations and users the same as the old buffalo system) then it might be good to keep the old identifier for consistency of the data if, say, some consumer was recording historical data and keying off of that identifier. 

If reddy is considered a brand new system and if consumers should treat their data sets as staring from scratch then it should definitely get a new identifier. 

@cubbi

jcn:
@cubbi regardless, the Reddy entry should be moved to the appropriate spot in the list alphabetically.

cubbi:
@jcn per your request I moved it to correct position in the file.  In regards to the identifier of the system, we will be using new one from now on.

PR #56, opened 2017-02-02
Added Abu Dhabi

serialc:
Also a syntax fix for Monash/Melbourne


jcn (changes requested):
Looks like the system ID is `ABU` not `abu_dhabi`: http://api-core.bikeshare.ae/gbfs/ar/system_information.json

PR #57, opened 2017-02-04
Orlando Juice no longer provides GBFS format feed

serialc:
Either their server has some problems or it's abandoned:
https://orlando.socialbicycles.com/opendata/gbfs.json


jcn:
Seems to have just moved: https://juicebikeshare.com/opendata/gbfs.json

jcn:
Added the updated URL in #58

PR #59, opened 2017-02-08
Clarify availability of GBFS feeds from NABSA members

barbeau:
* Per SoBi (@cubbi), Relay bikeshare (http://relaybikeshare.com/) in Atlanta, GA does not have a GBFS feed.  This patch clarifies that while all NABSA members have committed to implement the GBFS standard, this does not necessarily mean it will be available at all NABSA member deployments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Documentation for the General Bikeshare Feed Specification, a standardized data 
 
 General Bikeshare Feed Specification, known as GBFS, is the open data standard for bikeshare. GBFS will make real-time data feeds publicly available online in a uniform format so that map and transportation based apps can easily incorporate this data into their platforms.
 
-Under the North American Bikeshare Association’s leadership, bikeshare industry leaders including 8D Technologies, PBSC Urban Solutions, Social Bicycles, BCycle, Smoove, and Motivate developed the GBFS over the past year. All have agreed to implement the standard by the end of 2015.
+Under the North American Bikeshare Association’s leadership, bikeshare industry leaders including 8D Technologies, PBSC Urban Solutions, Social Bicycles, BCycle, Smoove, and Motivate developed the GBFS over the past year. All have agreed to implement the standard by the end of 2015, although the availability of a GBFS feed for each bikeshare deployment may vary.
 
 More information:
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
This is a bummer. There's quite a bit of interest in multi-modal transit for OneBusAway on iOS, and not having access to a GBFS feed for Atlanta will impact our ability to deliver this to our users.

barbeau:
Thanks to SoBi and @cubbi and team, Relay bikeshare in Atlanta now has a live GBFS feed!  Thanks SoBi and @cubbi!

I just opened a PR to add Relay to the list:
https://github.com/NABSA/gbfs/pull/60

I believe this PR is still valid, though.  The impression that I and it seems many others had is that GBFS was being deployed at all NABSA vendor locations.  But, it seems that isn't the case.  I certainly understand that it's up to the vendor to create whatever business model they want surrounding GBFS feeds, but I'd like to see this more clearly indicated in the README here so that cities and app developers don't assume they are getting something and then realize later that they aren't.

mplsmitch:
Generally speaking, the owner of a data feed is going to be the system owner which is typically a municipality, not a vendor. While vendors may agree to implement GBFS and offer it as a service to their clients, it's up to the individual client cities to opt in to publishing their system data.  Our challenge is to persuade cities that are on the fence about making their data public that GBFS will benefit them. I'll work on clarifying this in the README.

barbeau:
Thanks @mplsmitch!  Just curious - what reasons have cities given for opting out of publishing GBFS data?  

I'd certainly be interested in helping make the argument that it's worthwhile.  I'm wondering if it's similar to what we've seen in the past with GTFS data sharing.  @antrim and I wrote a paper for TRB that has a section on "GTFS Data Dissemination" (page 5 - http://bit.ly/TRB2017-GTFS) that might be relevant.  It cites some of the past fears that transit agencies had related to sharing data in GTFS format, and how practice and research has shown that the benefits strongly outweigh the risks.

cc'ing others that are interested in widespread availability of GBFS data as well - @antrim, @LeoFR, @kariwatkins.

antrim:
@mplsmitch : One thing we discovered with GTFS was that demonstrating compelling benefits (3rd party applications) encouraged agencies to open up their data. Keep us posted. I am curious how this develops. Creating some sort of information site for GBFS might help to make a clear case for open GBFS.

PR #61, opened 2017-04-06
Modify Toronto GPBS discovery to point to new URL

dsgermain:
Migration from Motivate to PBSC.

### DIFF
--- a/systems.csv
+++ b/systems.csv
@@ -2,7 +2,7 @@ Country Code,Name,Location,System ID,URL,Auto-Discovery URL
 AE,ADCB Bikeshare,"Abu Dhabi, AE",ABU,https://www.bikeshare.ae/,https://api-core.bikeshare.ae/gbfs/gbfs.json
 AU,Curtin University,"Curtin University, Perth, WA",curtin_university,http://curtinbikeshare.com/,http://curtinbikeshare.com/opendata/gbfs.json
 AU,Monash Bike Share,"Monash University, Melbourne, AU",monash_bike_share,https://monashbikeshare.com/,https://monashbikeshare.com/opendata/gbfs.json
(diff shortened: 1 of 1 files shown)


AntoineGiraud:
Hello @dsgermain,
I get a `Access-Control-Allow-Origin` issue when i try to access Toronto's GBFS from a web application (AJAX).
Can you enable it ? it's available for the others GBFS.
Thanks

jcn:
@dsgermain I already merged this in, but I realized that the `system_id` is not consistent with the `system_id` in the GBFS feed. Can you change one or the other to make them match each other and then submit a new PR? Thanks!

dsgermain:
@AntoineGiraud will look into it.
@jcn will do, thanks.

dsgermain:
@jcn since this is supposed to be constant over the lifetime of the system, we will change it in our feed during the next deployment.

jcn:
> @jcn since this is supposed to be constant over the lifetime of the system, we will change it in our feed during the next deployment.

@dsgermain To confirm, you will be changing the `system_id` that is actually sent in the gbfs feed to be `bike_share_toronto` on the next deployment? Thanks.

dsgermain:
@jcn that is correct. These feeds were previously wrapped by Motivate, but they are now being exposed directly which is why this ID changed (they were overriding our ID). We will update the feed to match the previously exposed public IDs.

jcn:
@dsgermain It also looks like your GBFS feed does not 100% conform to the GBFS spec. Specifically, you are using true/false values in a lot of places where GBFS specifies 1/0.

dsgermain:
@jcn Yes, that was reported to us. We will also address in the next release.

jcn:
> We will also address in the next release.

:+1:

PR #68, opened 2017-06-28
Remove BABS, add Ford GoBike

afischer:


jcn (changes requested):
Looks like your GBFS URL is not correct.

afischer:
Whoops, not sure how that happened.

PR #72, opened 2017-08-22
Remove USF Share-A-Bull feed

barbeau:
USF Share-A-Bull bikeshare has been merged into Coast Bike Share, so the Share-A-Bull bikes are now available via the Coast Bike Share GBFS feed.  The Share-A-Bull GBFS feed is now empty.


jcn:
For reference, here's a post noting the change: https://www.facebook.com/ShareABullBikes/photos/a.865815990260141.1073741829.831585480349859/865486890293051/?type=3&theater

PR #73, opened 2017-11-15
Update systems.csv

dsgermain:
Adding PBSC systems: Aruba, Aspen, Chattanooga, Guadalajara, Kona, Honolulu, Detroit, Louisville, Reykjavik, Recife

### DIFF
--- a/systems.csv
+++ b/systems.csv
@@ -11,7 +11,6 @@ CZ,Velonet,"Prague - Brno, CZ",velonet_cz,http://velonet.cz/,http://velonet.cz/o
 IS,WOW citybike,"Reykjavik, IS",wow_citybike,https://wowcitybike.com/,https://rey.publicbikesystem.net/ube/gbfs/v1/
 MX,Mibici Guadalajara,"Guadalajara, MX",mibici_guadalajara,https://www.mibici.net/en/,https://guad.publicbikesystem.net/ube/gbfs/v1/
 US,ArborBike,"Ann Arbor, MI",bcycle_arborbike,http://arborbike.org,https://gbfs.bcycle.com/bcycle_arborbike/gbfs.json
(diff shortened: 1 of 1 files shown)


jcn:
@dsgermain Thanks for these updates. I noticed some issues / inconsistencies between the `System ID` field in the csv and the actual `system_id` being reported by these GBFS feeds.

The [spec](https://github.com/NABSA/gbfs/blob/master/gbfs.md#system_informationjson) states:

"This should be globally unique (even between different systems) and it is currently up to the publisher of the feed to guarantee uniqueness. In addition, this value is intended to remain the same over the life of the system"

But in your update:

* In some cases (like Biki), the system is just returning a blank `system_id` which isn't valid as per the spec
* In other cases (like Arbua), the `system_id` reported by the GBFS feed is a UUID, but the ID you've included in the csv is something like "greenbike_aruba" (the Chattanooga GBFS feed used to return "bike_chattanooga" but now returnings a UUID)

Can the system GBFS feeds be change to match the values you've included in the CSV? Those are definitely more user-friendly than the UUIDs and would conform to the spec. 

We could also just remove the System ID from the CSV, but that's a different issue for discussion, and doesn't change my comments on the current PR.

Thanks!

maduprasPBSC:
This has been fixed in version 4.8.0 of the PBSC Backend:
- Aruba: green_bike_aruba
- Aspen: we-cycle
- Chattanooga: bike_chattanooga
- Chicago: divvy
- Columbus: cogo
- Detroit: mogo
- Guadalajara: mibici
- Kona: bike_share_hawaii
- Louisville: louvelo
- Reykjavik: wow-bikes
- SBU: sbu
- Toluca: huizi
- Toronto: bike_share_toronto

jcn:
> This has been fixed in version 4.8.0 of the PBSC Backend

@maduprasPBSC Thank you for the follow-up and for fixing the backend.

PR #79, opened 2017-12-18
Remove Bike Chattanooga in Chattanooga, TN from the available feed list.

tague:
The Chattanooga system still seems to exist but the GBFS feed seems to be gone.  Not affiliated with the system, so I can't speak to exactly what is going on.


jcn:
The operator changed so the feed url changed. Updated in #73.

PR #80, opened 2017-12-18
Update Chattanooga GBFS feed url

serialc:


jcn:
This change was merged in #73

PR #85, opened 2018-02-07
Update to README

mplsmitch:
Updates README with revised description of project and addtional clarification of change process and guiding principles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,21 +3,59 @@
 Documentation for the General Bikeshare Feed Specification, a standardized data feed for bike share system availability.
 
 ## What is GBFS?
+The General Bikeshare Feed Specification, known as GBFS, is the open data standard for bikeshare. GBFS makes real-time data feeds in a uniform format publicly available online, with an emphasis on findability.
 
-General Bikeshare Feed Specification, known as GBFS, is the open data standard for bikeshare. GBFS will make real-time data feeds publicly available online in a uniform format so that map and transportation based apps can easily incorporate this data into their platforms.
+Under the [North American Bikeshare Association’s](http://www.nabsa.net) leadership, GBFS has been developed by public, private sector and non-profit bike share system owners and operators, application developers, and technology vendors.
 
-Under the North American Bikeshare Association’s leadership, bikeshare industry leaders including 8D Technologies, PBSC Urban Solutions, Social Bicycles, BCycle, Smoove, and Motivate developed the GBFS over the past year. All have agreed to implement the standard by the end of 2015, although the availability of a GBFS feed for each bikeshare deployment may vary.
+ GBFS is intended as a specification for real-time, read-only data - any data being written back into individual bikeshare systems are excluded from this spec.
 
-More information:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,21 +3,59 @@
...
 
 * [Current spec](gbfs.md)
 * [Original draft spec in a Google doc](https://docs.google.com/document/d/1BQPZCKpem4-n6lUQDD4Mi8E5hNZ0-lhY62IVtWuyhec/edit#heading=h.ic7i1m4gcev7) (reference only)
 
+## Overview of the Change Process
+GBFS is an open specification, developed and maintained by the community of producers and consumers of GBFS data.
+The specification is not fixed or unchangeable. As the bikeshare industry evolves, it is expected that the specification will be extended by the GBFS community to include new features and capabilities over time. To manage the change process, the following guidelines have been established.
+
+The general outline for changing the spec has 4 steps:
+1.	Propose a change by opening an issue at the GBFS GitHub repository.
```
barbeau, line 28: IMHO this should be:

>Propose a change by opening a ***pull request*** at the GBFS GitHub repository

Based on my experience with GTFS and GTFS-realtime, it's far more productive to have a discussion on a pull request than an issue.  This is because the current proposal at any point in time is very clear in the form of a diff (i.e., "Files changed") in the PR.  And, as the discussion evolves, you can push more commits to the same PR to change it based on feedback.  As a result, you can clearly see the history of the proposal by looking at the commits in the PR.  And, when the proposal is finalized, you can squash and merge into one commit into the main repository.

jcn: @barbeau That's really interesting. Do you have an example of that kind of discussion that you could point at? I'm curious how it plays out in the real world.

barbeau: Sure, see https://github.com/google/transit/pull/64 as an example.

thekaveman: @jcn (this discussion, right now!)

Just kidding. Here's a simple example that went through a similar process (though it is code, not a spec), from a project I work on: CityofSantaMonica/SODA.NET#46

Here's another example from the same repository that hasn't been merged yet: CityofSantaMonica/SODA.NET#58

On the *Conversation* tab, you can see that after the PR was started, I added a few comments that triggered more code changes. These comments are hidden under the "Show Outdated" label, because the state of the code changed since I made the original comment. 

The current state of the changes can always be viewed on the *Commits* or *Files changed* tabs.

The one difference is, we didn't take the approach of squashing the merged PR into a single merge commit and instead opted to keep the full history in `master`.

mplsmitch: My intent is to keep this process as open as possible.  I'd like it to be possible for people without any specific technical knowledge who may own or operate the systems publishing under the spec to be able to contribute to it's development . I'm fully in favor of discussion of PRs as they take place but I don't want to limit proposed changes to people who are able to make a commit. Better that they should be able to bring their needs or ideas to the group and we help get them to the PR stage.

barbeau: @mplsmitch Sure, that's definitely a valid concern.  We wrestled with the same dilemma with GTFS/GTFS-realtime.  FYI, you can edit files, commit, and open a PR directly from the Github website, so you don't really need to understand Git or need to install software on your machine to open a PR - see https://help.github.com/articles/editing-files-in-another-user-s-repository/.

Issue #65, opened 2017-05-17
Polygon coordinates for stationless bikes


mplsmitch:
I'd like to get a discussion going around adding polygon coordinates to describe parking areas and service area boundaries for free floating bikes.  Cities are beginning to wrestle with how to regulate stationless bikes and it seems like there's an opportunity to address this via GBFS.  Since I run a station based system I'm not up to speed on how these data are currently described.  This issue also comes up with the proposed "super hubs" in Portland. Is there anyone from SoBi  or with GIS expertise that would be interested in working on this?

barbeau:
+1

We'd be interested in being a GBFS consumer for floating bike areas, both in [OneBusAway](http://onebusaway.org/) and our [USF Maps App](https://maps.usf.edu/#layers/bike_stations).  I actually just had this discussion with TriMet in Portland OR last week and I know they are interested in this for their multimodal app as well as part of their FTA MOD Sandbox project.  ccing @fpurcell from TriMet.

Right now for the USF Share-A-Bull system in Tampa, FL SoBi is publishing the drop off zones as individual virtual stations, defined as points in the existing `station_information.json` endpoint:
http://usf.socialbicycles.com/opendata/station_information.json

To give a concrete proposal to this discussion - how about a new endpoint?

### free_bike_zones.json

Field Name        | Required  | Defines
------------------| --------- | ----------
zone_id         | Yes       | ID field - identifier for this free bike drop-off zone. This should be globally unique (even between different systems) and it is currently up to the publisher of the feed to guarantee uniqueness. In addition, this value is intended to remain the same over the life of the system.  Each drop-off zone should be disjoint - there should be no geographic overlap between two different zones.
zone_area              | Yes       | A valid [GeoJSON `Polygon`](http://geojson.org/) that describes the area where free bikes can be dropped off.

I think our only realistic option to describe the geographic properties of zones is GeoJSON, as they can be complex.  Here's what the allowed drop off zone currently look like at USF (from https://usf.socialbicycles.com/):

![image](https://cloud.githubusercontent.com/assets/928045/26165892/9b997dd4-3aff-11e7-9936-1b321e1bf4e7.png)

serialc:
Sounds like a good idea with the growing number of such systems.

A few comments on refining the above definitions.

[...]

PR #86, opened 2018-02-07
GBFS V1.1 Adding support for electric bikes and geofenced bike return…

dsgermain:
… areas

Tried to decouple geofencing zones from stations as much as possible, while maintaining the ability to link them as that exists in some systems. 
free_bike_status.json can now be used to determine the number of each type of bike available per station and per geofencing zone. 

- Add geofencing_zone_information.json to list low volatility information relating to geofencing zones, such as their position and which station, if any, they are linked to.

- Add geofencing_status.json to list transient information relating to geofencing zones, such as whether the zone is allowing returns and the number of bikes it contains.

- Modify free_bike_status.json to include the bike type, as well as a way to differentiate whether bikes are geofenced at a station, or geofenced in a free floating system.

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -12,6 +12,8 @@ This document explains the types of files and data that comprise the General Bik
     * [system_information.json](#system_informationjson)
     * [station_information.json](#station_informationjson)
     * [station_status.json](#station_statusjson)
+    * [geofencing_zone_information.json](#geofencing_zone_statusjson)
+    * [geofencing_zone_status.json](#geofencing_zone_statusjson)    
     * [free_bike_status.json](#free_bike_statusjson)
     * [system_hours.json](#system_hoursjson)
     * [system_calendar.json](#system_calendarjson)
@@ -21,6 +23,7 @@ This document explains the types of files and data that comprise the General Bik
 * [Possible Future Enhancements](#possible-future-enhancements)
 
 ## Revision History
+* 07/02/2018 - GBFS V1.1 Adding support for electric bikes and geofenced bike return areas
 * 11/05/2015 - GBFS V1.0 Adopted by NABSA board
 * 08/2015 - Latest changes incorporated and name change to GBFS (comments from Motivate, 8D, others)
 * 06/2015 - Proposed refinements (prepared by Jesse Chan-Norris on behalf of Motivate)
@@ -42,7 +45,9 @@ File Name                   | Required      | Defines
 gbfs.json                   | Optional      | Auto-discovery file that links to all of the other files published by the system. This file is optional, but highly recommended.
 system_information.json     | Yes           | Describes the system including System operator, System location, year implemented, URLs, contact info, time zone
 station_information.json    | Yes           | Mostly static list of all stations, their capacities and locations
-station_status.json         | Yes           | Number of available bikes and docks at each station and station availability
+station_status.json         | Optional      | Number of available bikes and docks at each station and station availability. Optional as system can be free floating.
+geofencing_zone_information.json    | Optional   | Mostly static list of all geofencing zones, their capacities, locations and areas
+geofencing_zone_status.json      | Optional      | Defines geofencing zones available in the system and their link to physical stations if any. Optional because system might be based on physical stations only.
 free_bike_status.json       | Optional      | Describes bikes that are available in non station-based systems
 system_hours.json           | Optional      | Describes the hours of operation for the system
 system_calendar.json        | Optional      | Describes the days of operation for the system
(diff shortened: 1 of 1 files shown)


gbfs.md
```diff
@@ -207,6 +229,15 @@ stations              | Yes       | Array that contains one object per station i
...
 \- last_reported       | Yes       | Integer POSIX timestamp indicating the last time this station reported its status to the backend
 
+### geofencing_zone_status.json
+Field Name            | Required  | Defines
+--------------------- | ----------| ----------
+geofencing_zones      | Yes       | Array that contains one object per geofencing zone in the system as defined below
+\- geofencing_zone_id  | Yes       | Unique identifier of a geofencing zone (see geofencing_zone_information.json)
+\- num_bikes_available | Yes       | Number of bikes available for rental
+\- num_bikes_disabled  | Optional  | Number of disabled bikes in the geofencing zone. 
+\- is_returning        | Yes       | 1/0 boolean - is the geofencing zone accepting bike returns (business rules, such as city regulation may affect whether we are allowed to return in this zone or not during certain time slots)
```
barbeau, line 239: I'd recommend calling this field `accepting_returns`.

barbeau:
@dsgermain Thanks for putting this together!  One comment in-line, but generally speaking it looks good to me.

jcn:
@dsgermain I might have missed this in the discussion (#65) but why did you decide to implement the  geofenced return area as a separate feed rather than as different type of station? This seems like it significantly duplicates the `station_information` feed - couldn't `station_information` just have a new required field added called "station_type" with the possible values of "fixed" or "geofenced" (with additional types in the future as appropriate)?

gbfs.md
dsgermain: OK! I mirrored it on the naming already present in the station feed to not have different naming conventions. I didn't like the name either, but figured uniformity was better. Thoughts?

dsgermain:
@jcn I hesitated a while before choosing to go this way. My thinking is that geofencing_zones are a separate concept from stations, with different properties, parameters and business logic for front end applications. They can be **tied** to stations, but also have specific paradigms like the ability to overlap and be embedded in one another. I think my implementation of them is incomplete and will get extended as the use cases grow, and those would end up bloating up the station endpoint.

It's more of a philosophical issue, but I opted for separation of concerns. I'm opened to reconsider if you guys think it would be better in a merged solution.

nbdh:
Both big operators of bike sharing systems in Germany, DB CallABike (e.g. LIDL-BIKE in Berlin) and nextbike have real stations and "virtual stations" (also called "return zones") in many cities. You would probably classify the latter as geofenced bike return areas. As mentioned in #89 it's almost always possible to return bikes next to real stations (with docks) if there are no docks available and just lock it to itself.
To cover this scenario consequently their GBF would have to contain additionally to every real station a circular shaped geofenced area.
However from my point of view geofenced areas are conceptually really close to real stations and I think a what you called "merged solution" would be be sufficient to bear relevant information. It seems the separate-endpoint-concept would complicate things quite a lot. Even the official apps / websites are not showing the exact boundary of the area and as far as I know both aforementioned operators just "merge" all bikes within a certain radius around the "station" coordinate into the station.
I'd like to gain a better understanding of the @motivateco system's and how they look (also for #65).

dsgermain:
@nbdh how would you treat systems that have no stations at all, such as free floating with drop off zones, if both endpoints are merged?

nbdh:
How many zones are there and how large are they? Do you have an example?

I'm not sure I understand your question correctly. I'm proposing that small geofenced return zones (virtual stations) without docks should fit under the GBFS definition of station. A field in station_information as proposed by @jcn to be able to distinguish between real and virtual station seems to suffice. Consequently bikes parked at a return zone would not be listed in free_bikes_status. 
As your geofencing_zone_status.json is basically a station_status with less attributes the only new field is zone_area.
I know only systems with two types of "return zones":
1. systems with very few but very large return areas (that's what I'd consider free floating)
2. systems with many rather small return zones which the operator considers as "dockless stations"  (usually it's not even communicated to the user that there is no real station with docks and the 'station' is still represented by a single icon and not an polygon in the front-end, but as I said basically all bikes inside of a 25-75ish meter circle will be considered as being parked at the station.

In mixed systems those dockless return zones can lie within or outside of the free floating return areas
I was under the impression that we discussed how this large return areas of free floating systems can be embedded in GBFS in #65.

dsgermain:
@nbdh We have clear hybrids of physical stations with defined zones near them, or around them. 
Is your argument around complexity of implementation only? I don't think it is very hard to add two endpoints to define zones, and it covers all cases and future extensions. Architecturally, I think it is cleaner.
The geofenced zones have a whole bunch of properties that are distinct from stations. They are not always centered on the station, and they have coordinates that can be defined by the city or the operator beyond a simple circle around a station. They also have properties that are distinct from the station, such as operation hours when you can return to them, which are distinct from station hours. Maybe you haven't seen these usecases yet, but we have.

I think all systems can work with my proposal, but some systems can't work with yours.

If you want, please propose a clear counter-proposal pull request that arrives at the same objectives and provides future extensibility. We could then review in more detail. If not, I propose we move forward with this one because we will start implementing soon.

Thanks!

nbdh:
@dsgermain As I mentioned I'd first like to understand the systems you are talking about better, thus I asked for examples. How are the return zones displayed in the front end right now?
You mention different properties and yet the only additional field in the new end point compared to station_information is zone_area. Which else do you have in mind?
My concern is also about complexity as it should remain as easy as possible to understand GBFS and  implement GBFS into clients. To me it also seams like a breaking proposal because if an publisher formerly treated geofenced return zones as stations and moves them to the new endpoint they'll disappear in clients that are not consuming the new endpoint.
Additionally your proposal breaks the free_bikes_status. In the description it says 'that are not at a station' but yet there is a field "station_id". 
If I get it right you intended to list all free (in sense of lendable) bikes in this endpoint, which makes sense to transport additional bike data . But it is again a breaking proposal as old GBFS clients will start showing not only the station but also all lendable bikes at the station or must implement a filter logic to exclude bikes from the free_bikes_status that have set a station_id or geofence_zone_id.

About the operation hours: Even though it's not in the specification yet to specify operation hours for single stations I'll try to take that into account when creating a counter-proposal.

dsgermain:
@nbdh 

> As I mentioned I'd first like to understand the systems you are talking about better, thus I asked for examples. How are the return zones displayed in the front end right now?

There are not current systems using this. We are proposing this because we are adding these new features to our systems this year. The frontend will display stations as before. Optionally, there will be zone(s) near or around the stations where bikes can be returned, only when the station is full. These zone(s) can also have schedules associated to them, allowing or disallowing returns. This is to match city regulation. For instance, a city might have garbage trucks or sidewalk cleaning equipment going through on Thursdays between 1 and 3PM. The zone would not allow parking during that timeframe.

> You mention different properties and yet the only additional field in the new end point compared to station_information is zone_area. Which else do you have in mind?

As mentioned, these zone(s) are not necessarily at the same coordinates as the station, or even attached to a station. The coordinates are thus distinct. The capacity, the schedule and the defining polygon are new properties. 


> To me it also seams like a breaking proposal because if an publisher formerly treated geofenced return zones as stations and moves them to the new endpoint they'll disappear in clients that are not consuming the new endpoint.


I don't see how this is breaking. There is nothing mentioning zones in the current station implementation. If they were using this feed in an undocumented way, and we introduce a formal way of doing things, that is not a breaking change. Nothing prevents the frontend implementer of continuing to use things in an undocumented way as they did before and display zones around their stations. We could add the feed version number to GBFS.json so that frontend clients know how to handle different versions. If the spec can't change because people are using it in undocumented way, it is not worth much to have a spec!

> Additionally your proposal breaks the free_bikes_status. In the description it says 'that are not at a station' but yet there is a field "station_id".

Agreed. The goal here was to allow easy aggregation of the bike count for a zone nearing a station with that station's bike count, but that could be done through latitude/longitude proximity instead and this new field could be removed.

----

The separation of both concepts (station and zone) also allows multiple distinct dropoff zones near a station. Would you address this by adding a list of optional zones to each station? And then supplementing station_status with a list of statuses for these different zones? I find that bloats the structures.

Anyone else wants to chyme in with an opinion? I'd like to move forward one way or another so that we have clarity on how this should be implemented. We are starting implementation on this in about 3-4 weeks.

Thanks,
David

dsgermain:
What is the merge process for pull requests? Do we need some sort of consensus / who has final approval?

dsgermain:
Closing as I need to add more fields for ebikes. Will open another pull request.

Issue #94, opened 2018-04-13
Number of bikes and docks available during non service hours or out of service stations


Noe-Santana:
Hello everyone.

There are a some systems that doesn't work the whole day, in these cases, What should be the behaviour of the num_bikes_available during the non service hours? 
 
There are two values that indicate if a station is renting or receiving bikes, additionally, The number of bikes and docks should change to "0" on each case?, I can see that not all the systems are managing these values the same way so I would like to find a standarization.
 
In our experience, mantain the value of bikes and docks available at the station as an independent value from the renting and returning flags can be much more valuable for analisys and developing purposes, also, as an end user is very functional to know where the bikes are, it allows the user to plan where to go and take a bike at the very first hours of service, ultimately, there are a lot of options like color, size, shape, symbols, legends, etc. to represent the functional status and the occupation of a station at the same time.

mplsmitch:
During non-service hours num_bikes_available would be "0". This probably wouldn't be true of num_docks_available since most systems allow returns outside of business hours.

Noe-Santana:
Thanks for your response Mitch.

Doo you think that the "num_bikes_available" on the feed should change when there is a "is_renting" flag?, if a developer of an end user tool based on the GBFS decides that it should work the way you propose it still possible by using this flag.

On the other hand, if there is a purpose of analysis of the distribution of the bikes, having the numbers of bikes at the stations at all time is something indispensable, in fact is a number that tends to be more accurate when the system is not in service for the end users. 

The value always available can reflect the work of the operators at night, either for the community interested on bikeshare systems or for the regulators of the service like in some cases the goverment.

barbeau:
IMHO `num_bikes_available` should definitely be the the number of bikes available to rent right now - consumers shouldn't need to check the `is_renting` flag to see what this field means.  A different default interpretation based on `is_renting` would break existing consumers.  The [GBFS Guiding Principles](https://github.com/NABSA/gbfs#guiding-principles) say:

>GBFS is a specification for real-time or semi-real-time, read-only data. The spec is not intended for historical or archival data such as trip records. The spec is about public information intended for bikeshare users.

While being able to analyze service would be nice, IMHO this needs to be in separate fields dedicated to this concept, if they should be in GBFS at all.

jcn:
Doesn't the consumer always have to check is_renting - otherwise, what's the point of it? The explanation of that field is:

[...]

PR #95, opened 2018-04-21
Add "num_usable_bikes" and "num_usable_spaces" to station_status.json

albloptor:
This change tries to fix the limitation the current spec has, where it's not possible to know how many bikes are there at a station when `is_renting = 0`.

In my opinion, as exposed in issue #94, `is_renting` should be independent to `num_bikes_available`. In the same way, `is_returning` should be independent to `num_docks_available`.

Because changing the spec would break existing clients implementing it, I propose adding `num_usable_bikes` and `num_usable_spaces` to the spec.

These two new fields, in conjunction with `is_renting` and `is_returning`, arguably make `num_bikes_available` and `num_docks_available` redundant, but keeping them guarantees backward compatibility.

What we achieve with this spec change is to know how many bikes and spaces there are in a station regardless of the `is_renting` and `is_returning` flag. 

This allows clients to:
- Keep all the current GBFS functionality, but providing the extra features below. 
- Plan their journeys before the beginning of a station's renting hours. If a station starts renting at 7am, they will be able to see how many usable bikes there are at the station before that time. The current spec would display 0 bikes until 7am.
- Visualize number of bikes in each station on a map, regardless of the renting/returning hours.

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -198,6 +198,8 @@ Field Name            | Required  | Defines
 --------------------- | ----------| ----------
 stations              | Yes       | Array that contains one object per station in the system as defined below
 \- station_id          | Yes       | Unique identifier of a station (see station_information.json)
+\- num_usable_bikes    | Yes       | Number of usable bikes at the station. A bike is usable if it's not broken and could be rented if `is_renting` was true.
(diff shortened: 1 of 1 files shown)


jcn:
Closing this as there appears to be consensus in #94 that adding additional fields does not actually solve the underlying problem, and that the better course of action is to fix/clarify the existing fields.

Issue #96, opened 2018-04-25
Freefloating bicycles


sven4all:
First well done with creating this standard!

I want to use this standard to make positions of freefloating bicycles available available as open data. In the standard https://github.com/NABSA/gbfs/blob/master/gbfs.md#files is specified that station_information and station_status are required. In a freefloating bicycle system there are no stations and therefor these files are not relevant. 

Another suggestion that I have is adding a type of system (free floating/ station_based maybe some other options renting a bike at a station and having to bring it back to the same station) to system information https://github.com/NABSA/gbfs/blob/master/gbfs.md#system_informationjson 

I am interested in your opinion. Based on that I could propose formal changes.

mplsmitch:
@sven4all You are correct - in a free-floating system, `station_infomation.json` and `station_status.json` would not be required end points but `free_bike_status.json` would go from optional to required. What do you think would be the best way to express that? I'm reluctant to make every end point optional.

`system_type` sounds interesting. The challenge is in clearly defining the types. I could see something like:

- station based

- free floating

- hybrid (both stations and free floating)

Going beyond that would require a way to describe the various business rules or city regulations. For example in a free floating system are users required to lock the bikes to a physical object (lock to requirement), park them in designated spaces ( virtual station), park anywhere legal or some combination?

sven4all:
@mplsmitch Thank you for your input!

I think it should be made semi-optional. You have to provide at least station_information.json and station_status.json or free_bike_status.json. If neither of this two is provided it's completely useless. Do you agree?

The second part is a bit more difficult. I don't have an overview of all different policies that are implemented world wide (in the Netherlands an effort is made to create one policy that is applied by all municipalities). Let's start with `system_type` and later think about how to implement all those rules, hopefully there will some form of standardisation by governments on those rules as well.

barbeau:
>I think it should be made semi-optional. You have to provide at least station_information.json and station_status.json or free_bike_status.json. If neither of this two is provided it's completely useless.

We used the term "Conditionally required" in [GTFS-realtime v2.0](https://medium.com/@sjbarbeau/whats-new-in-gtfs-realtime-v2-0-cd45e6a861e9), and then explained the conditions under which the field was required in the "Description" section.

sven4all:
@barbeau Good addition!

tomschenkjr:
Great conversation. 

[...]

PR #102, opened 2018-06-14
Make station_information.json, station_status.json and free_bike_status.json conditionally required.

sven4all:
This change is based on the discussion we had in #96. I used the suggestion of @barbeau to use the term conditionally required. 

This what is changed:
station_information.json and station_status.json are
now conditionally required instead of required.
free_bike_status.json is conditionaly required
instead of optional.

Station based systems can still use the same structure as before.
But free floating systems don't have to provide a station_information.json
and station_status.json file anymore if they provide the free_bike_status.json
file.

There are already several implementations that implement this see the following urls for example: https://ddot.dc.gov/page/dockless-api
https://mobike.com/us/gbfs/v1/free_bike_status
https://web.spin.pm/api/gbfs/v1/free_bike_status

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -41,9 +41,9 @@ File Name                   | Required                |       Defines
 --------------------------- | ----------------------- | ----------
 gbfs.json                   | Optional                | Auto-discovery file that links to all of the other files published by the system. This file is optional, but highly recommended.
 system_information.json     | Yes                     | Describes the system including System operator, System location, year implemented, URLs, contact info, time zone
-station_information.json    | Conditionally required  | Mostly static list of all stations, their capacities and locations. Either (station_information.json and station_status.json) or free_bike_status.json should be provided.
-station_status.json         | Conditionally required  | Number of available bikes and docks at each station and station availability. Either (station_information.json and station_status.json) or free_bike_status.json should be provided.
-free_bike_status.json       | Conditionally required  | Describes bikes that are available in non station-based systems. Either (station_information.json and station_status.json) or free_bike_status.json should be provided.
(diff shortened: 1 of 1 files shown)


mplsmitch:
I'm very supportive of this change but I think we need to do some work on the definitions. As @barbeau described in #96, if something is "Conditionally required" the definitions need to cover the conditions under which they are (or aren't) required.  Without that all three files are effectively optional.

Here's a first take at that - please feel free to refine/iterate.  


station_information.json    | Conditionally required  | Mostly static list of all stations, their capacities and locations. Required of systems utilizing docks. 
station_status.json         | Conditionally required  | Number of available bikes and docks at each station and station availability. Required of systems utilizing docks.
free_bike_status.json       | Conditionally required  | Describes bikes that are available for rent.   Required of systems that don't utilize docks or offer bikes for rent outside of stations.


Questions I have:  
* Does this adequately cover systems with both docked and dockless bikes ( Hybrids)?  
* Do the definitions need more detail to accommodate systems like BIKETOWN in Portland that don't use docks but have designated racks and lock-to requirements?

sven4all:
I think this are good refinements. Will change them in a few days, maybe there are improvements of other people.

> Does this adequately cover systems with both docked and dockless bikes ( Hybrids)?

I think so

> Do the definitions need more detail to accommodate systems like BIKETOWN in Portland that don't use docks but have designated racks and lock-to requirements?

I think to cover this we maybe have to introduce a station_type in station_information. Values could be for example station_type ["DOCKING", "RACK", "VIRTUAL", "MANNED"]. I will create a separate proposal for that.

PR #110, opened 2018-09-02
Fix administrative area code for nextbike Canada

iOS4ever:
Updates the administrative area code for bike share in Victoria as it is in British Columbia, or BC.


iOS4ever:
Huh, just noticed that the cron job is using the country code for the administrative code. I think a more extensive fix needs to be in place.

yocontra (approved):

PR #120, opened 2018-11-21
Add new JUMP systems - San Diego and Seattle

cubbi:
Additionally, change marketing URL for jump systems.


jcn:
I'm going to merge this in, but in the future this should have been 2 commits: one to update the existing system URLs and one to add the systems. It would make it easier to catch errors like this. 

Thanks!

PR #135, opened 2019-01-31
remove mountain ride bike share

yocontra:
domain expired and is now a godaddy landing page, has been in this state for 2 days so assuming the program is defunct


jcn:
I have been in touch with Mountain Rides who say

> Our bike share program is still operational. We have closed up shop for the winter months and need to update our website.

I am waiting to hear whether their domain has changed or if the expiration was an oversight.

yocontra:
@jcn I called them on the phone before I opened the ticket and they said the same thing, but I got the impression the employees answering phones/emails aren't closely related/knowledgeable in relation to the bikeshare system. I asked about the domain and they said they don't manage it, and that isn't their website.

jcn:
I've given them a chance to respond and for the domain to return, but it seems like it's gone for the time being. If it comes back, we can re-add it.

Issue #105, opened 2018-07-20
Shared scooters


Dehorser:
Will gbfs be expanded to cover shared scooters?

michalnaka:
Yeah I was wondering the same thing^

mplsmitch:
There's been some discussion about how might be done in #81  - As far as I know it's already being used for scooters in some cities but I haven't found the data source, here's an example:  http://bikes.oobrien.com/washingtondcskip/#zoom=10&lon=-77.1224&lat=38.7025

sven4all:
Would be interesting, I am working on a concrete proposal. I think it's a good idea to take scooters into account. https://github.com/openbikeshare/gbfs/blob/bike_types/gbfs.md#bike_typesjson

heidiguenin:
Hey all! PR #136 responds to this request, please take a look and provide any comments/concerns there. Once folks have had a chance to check it out, I'd love to go ahead and close this issue.

@Dehorser @michalnaka @sven4all

Issue #37, opened 2016-06-30
Provide JSON Schema for GBFS Documents


afischer:
I'm not a huge fan of [the JSON schema](http://json-schema.org/), but there are quite a few libraries [(some examples)](https://github.com/json-schema-org/JSON-Schema-Test-Suite#who-uses-the-test-suite) that use or support schema validation.

I'm currently trying to validate geoJSON, and I am shocked that there is no official schema. I think it would be helpful to developers to provide the schema ourselves.

barbeau:
+1 for this.  It does make it easier to generate client-side code used to parse feeds, especially if you have a number of optional fields that some producers may not implement.

thzinc:
I'd love some feedback on my PR to add schema files to the spec. :)

antrim:
@thzinc: Could I close this issue so that we guide all discussion towards your open pull request?

thzinc:
@antrim, that’s fine by me. (It’s not my issue, but I support it)

antrim:
@afischer May I close this issue (to continue discussion in PR #70)?

afischer:
Sure!

PR #136, opened 2019-02-19
Add vehicle type definitions

evansiroky:
Hello there, I am Evan Siroky and I have been working on improving [OpenTripPlanner](https://github.com/opentripplanner/OpenTripPlanner/) to be able to plan multi-modal trips with one-way car rentals (Car2Go and ReachNow). In this process, I copy-pasted a lot of the code for the bikeshare stuff and realized that it would be great if all this common code could be generalized to accomodate any kind of vehicle rental system.

This spec is a great standard in use by many bikeshare systems, but it can kind of be generalized to other modes. This repository already has a ticket about adding scooters (#105) and there's also some discussion about bike types (#81). It also looks like there are 2 other places where people have began work on proposing a change to add e-bikes (see @sven4all's changes [here](https://github.com/openbikeshare/gbfs/blob/bike_types/gbfs.md#bike_typesjson) and @dsgermain's changes [here](https://github.com/NABSA/gbfs/pull/92)).

I've done a lot of thinking about this and am going to go ahead and throw this PR out there with proposed changes that will hopefully allow this specification to accomodate basically any vehicle that someone could rent. My approach to adding these fields comes from a trip planning software development perspective, so some of the highly technical bits of information like the cruising speed, weight, power output and energy potential are things that I'm asking for in order to be able to calculate how far a vehicle could travel before running out of juice and how well the vehicle could climb inclined roads.

<strike>

## High level changes

* Renames `GBFS (General Bikeshare Feed Specification)` to `GVFS (General Vehicleshare Feed Specification)` to reflect the specification's possibility of defining a vehicle share system with any type of vehicle. 🤣 Not sure how far I'm going to get with this suggestion, so I'm totally fine with changing things back to GBFS.
* Renames file `gbfs.json` to `gvfs.json`.  🤣
* Renames file `free_bike_status.json` to `free_vehicle_status.json`

## Changes to `system_information.json`

Adds a required field `vehicle_types`. This value of this key would contain an array of objects with the following keys: `vehicle_type_id`, `name`, `description`, `vehicle_url`, `image_small`, `image_large`, `propulsion_type`, `cruising_speed`, `range_with_full_energy_potential`, `vehicle_weight`, `power_output`, `passenger_capacity` and `enclosed`.

## Changes to `station_information.json`

Adds a required field `allowed_vehicle_types` which would contain a list of strings that are `vehicle_type_id`s that are allowed to use the station.

## Changes to `station_status.json`

* Renames field `num_bikes_available` to `num_vehicles_available`
* Renames field `num_bikes_disabled` to `num_vehicles_disabled`
* The fields `num_vehicles_available`, `num_vehicles_disabled`, `num_docks_available` and `num_docks_disabled` are all changed to have an object with a key representing the vehicle_type_id and a numeric value indicating the number for that vehicle type.

## Changes to `free_bike_status.json`

* File is renamed to `free_vehicle_status.json`
* root key name `bikes` is renamed to `vehicles`
* subfield `bike_id` is renamed to `vehicle_id`
* fields `vehicle_type_id` and `range_with_current_energy_potential` are added

## Changes to `system_alerts.json`

* field `vehicle_type_ids` is added

# 2019-09-19 Update

After discussions occurring as a part of ongoing work related to https://github.com/NABSA/gbfs/issues/137, I am significantly revising and simplifying this PR in hopes of coming to a consensus on a minimum viable product. Furthermore, I have reimplemented this PR in a non-breaking way. See below for a list of proposed changes:

## Addition of `vehicle_types.json` file

This optional file can be used to desribe different vehicle types that an operator is renting out. The current fields allowed are `vehicle_type_id`, `form_factor`, `propulsion_types`, `description` and `max_range`.

## Changes to `station_information.json` file

The field `vehicle_type_capacity` can now be added to specify capacity by vehicle type at a station. A note of clarification has been added to the existing `capacity` field in such a way that existing feeds won't break.

## Changes to `station_status.json` file

The fields `vehicle_docks_available`, `vehicle_docks_disabled`, `vehicles_available` and `vehicles_disabled` have all been added. Clarification notes and changes in requirements have been added to the `num_bikes_available` and `num_docks_available` in such a way that existing feeds won't break with these changes.

## Changes to `free_bike_status.json` file

The fields `vehicle_type_id` and `range` have been added in such a way that existing feeds won't break.

# 2019-10-07 Update

After discussions occurring at the GBFS developer workshop, I have made two small changes that should now make this PR non-breaking. See below for a list of all proposed changes for this PR:

## Addition of `vehicle_types.json` file

This optional file can be used to describe different vehicle types that an operator has available for rent. The current fields allowed are `vehicle_type_id`, `form_factor`, `propulsion_types`, `max_range_meters` and `name`.

## Changes to `station_information.json` file

The field `vehicle_type_capacity` can now be added to specify capacity by vehicle type at a station. A note of clarification has been added to the existing `capacity` field in such a way that existing feeds won't break.

## Changes to `station_status.json` file

The fields `vehicle_docks_available`, `vehicle_docks_disabled`, `vehicles_available` and `vehicles_disabled` have all been added. Clarification notes and changes in requirements have been added to the `num_bikes_available` and `num_docks_available` in such a way that existing feeds won't break with these changes.

## Changes to `free_bike_status.json` file

The fields `vehicle_type_id`, `last_reported` and `current_range_meters ` have been added in such a way that existing feeds won't break.</strike>

# 2019-11-23 Update

See below for a list of all proposed changes for this PR:

## Addition of `vehicle_types.json` file

This optional file can be used to describe different vehicle types that an operator has available for rent. The current fields allowed are `vehicle_type_id`, `form_factor`, `propulsion_type`, `is_carryable`, `max_range_meters` and `name`.

## Changes to `station_information.json` file

The field `vehicle_type_capacity` can now be added to specify capacity by vehicle type at a station. A note of clarification has been added to the existing `capacity` field in such a way that existing feeds won't break.

## Changes to `station_status.json` file

The fields `vehicle_docks_available`, `vehicle_docks_disabled` and `vehicles` have all been added. Clarification notes and changes in requirements have been added to the `num_bikes_available`, `num_bikes_disabled`, `num_docks_available` and `last_reported` in such a way that existing feeds won't break with these changes.

## Changes to `free_bike_status.json` file

The fields `vehicle_type_id`, `last_reported` and `current_range_meters ` have been added in such a way that existing feeds won't break.


sven4all:
Interesting  proposal, maybe it's time for a bigger update for GBFS/GVFS  to include all the developments that have been going on over the past 4 years (The proposal that you are doing now is breaking existing api's, that requires a 2.0 update). 

This is a wish list I created last year https://github.com/openbikeshare/gbfsplus. It would be nice if as much business rules as possible can be generalised in a standard. That's the only way accurate planning algorithms can be build.

gvfs.md
```diff
@@ -160,18 +161,80 @@ The following fields are all attributes within the main "data" object for this f
...
+        "vehicle_url": "https://www.inboardtechnology.com/products/the-glider",
+        "image_small": "http://cdn.shopify.com/s/files/1/1136/4406/files/glider_nav_1_x150.png?11016692238788873189",
+        "image_large": "http://cdn.shopify.com/s/files/1/1136/4406/products/glider_pdp_4_super_duper_extended_2_de0ab82e-e603-4455-8395-dd6ec3b28cbb_375x375_crop_center.png?v=1542056407,%20//cdn.shopify.com/s/files/1/1136/4406/products/glider_pdp_4_super_duper_extended_2_de0ab82e-e603-4455-8395-dd6ec3b28cbb_750x750_crop_center.png?v=1542056407%202x",
+        "propulsion_type": "ELECTRIC",
+        "maximum_cruising_speed": 25,
+        "range_with_full_energy_potential": 19.3,
+        "vehicle_weight": 15.88,
+        "power_output": 350,
+        "passenger_capacity": 1,
+        "enclosed": 0
```
sven4all: Maybe add a field for if and what kind of driver license is needed according to local rules.

gvfs.md
```diff
@@ -160,18 +161,80 @@ The following fields are all attributes within the main "data" object for this f
...
+  "last_updated": 1434054678,
+  "ttl": 12345,
+  "data": {
+    "system_id": "inboard",
+    "language": "en",
+    "name": "Inboard",
+    "short_name": "ib",
+    "url": "https://www.inboardtechnology.com/",
+    "timezone": "America/Los_Angeles",
+    "vehicle_types": [
```
sven4all: Is there a reason you prefer this information within system.json? 

A separate file for vehicle_types is more in the spirit of the standard until now in my opinion. But it doesn't really matter to me.

gvfs.md
```diff
@@ -160,18 +161,80 @@ The following fields are all attributes within the main "data" object for this f
...
+      {
+        "vehicle_type_id": "m1",
+        "name": "M1 Electric Skateboard",
+        "description": "Experience the most advanced e-board deck ever developed. A custom designed composite deck that combines an inverted 3D Poplar wood core with full sandwich PU sidewalls and wrapped in the same fiberglass top sheet found in premium snowboards and skis. Integrated electronics, reinforced truck mounting points, and resilient TPU nose and tail bumpers at the rails yield a remarkably light and durable platform optimized for a more responsive ride.",
+        "vehicle_url": "https://www.inboardtechnology.com/products/m1-electric-skateboard",
+        "image_small": "http://cdn.shopify.com/s/files/1/1136/4406/files/Skateboard_x150.png?11016692238788873189",
+        "image_large": "http://cdn.shopify.com/s/files/1/1136/4406/products/m1_pdp_fullBleed_6_9e6614b5-c3e4-4dd1-b18b-fd51451d0c17_1600x.jpg?v=1535665708",
+        "propulsion_type": "ELECTRIC",
```
sven4all: maybe add a field voor vehicle_type as well?

bike/scooter/car/motor

sven4all (commented):
Already some comments.

yocontra:
IMO with scooters entering the mix this is a solid move to start thinking about this. We work with all of the GBFS feeds daily and noticed everyone is starting to deviate from the spec to accommodate dockless bikes, custom fields for scooter battery levels, vehicle type fields all named differently, etc. - and once everyone starts to deviate from the spec for their own use cases the spec doesn't mean much anymore.

To pull off a big change like this I think it needs to have an accompanying JSONSchema + a CLI or web tool to validate a URL against the spec. This will make it much easier for people to migrate to this next iteration. I can assist anywhere needed for this, so let me know if you need a hand with it.

As one point of feedback though - I think it would be nice to use objects instead of arrays of objects with key/value pairs (`vehicle_types`). Much simpler to reason about and work with.

gvfs.md
evansiroky: Is there some data standard that we can reference to add this data?

gvfs.md
evansiroky: I put it in here because I thought it made sense for the vehicles to be associated with a certain system or operator.

gvfs.md
evansiroky: I intentionally avoided this kind of field because there are already so many kinds of micromobility vehicles being made that may not be classifiable in a commonly understood way. In the example json in this PR, I included an electric skateboard which already falls outside the proposed types you have listed. This kind of classification of the vehicle type I believe is best left for interpretation by the gvfs-consuming application. In OpenTripPlanner I forsee adding updaters that would parse each vehicle type and assign it a various operational characteristics that would determine how the vehicle would be able to travel through the street network in the routing algorithms.

evansiroky:
I very much like the idea of a CLI and/or validation tool, but do not have the capacity for contributing towards building one at this time. Looks like someone may already be doing this in https://github.com/PierrickP/gbfs-validator?

afischer:
@evansiroky it seems like simple JSON schema validation would be perfect for validation across languages/implementations. #37 was opened for that a while back.

gvfs.md
sven4all: I think it doesn't hurt to add a vehicle_type for all mainstream modes. If new mainstream modes appear they can be added as well to the specification without breaking previous api's. We could also specify 'other' for all other modes that should be derived.

gvfs.md
sven4all: Within the European Union we do have a standardised classification https://en.wikipedia.org/wiki/European_driving_licence, I don't know if there is any data standard for this.

efel85:
@evansiroky good idea! I've been working the past few months on generalizing the GBFS specs to fit any transport mode. We have a large Mobility as a Service (also called mobility on demand) deployment in the Netherlands at the moment, I'm trying to facilitate a number of transport operators (bike & scooter sharing, car sharing and public transport) into creating a common API. 

We have a working group for standardisation set up within the MaaS-alliance, a global public-private collaboration of stakeholders. We are aiming to set up our work on https://github.com/maas-alliance/apis. I've been working from Swaggerhub until now (https://app.swaggerhub.com/apis/efel85/GTOAS/1.0.4), but will be merging everything on there to the maas-alliance github. I would be very interested in hearing what you think and if we can collaborate!

mfp22:
Today I have to start creating TypeScript interfaces to use for both bikes and scooters. This PR is very interesting to me, but like @contra says, we will have to deviate from the GBFS standard because this isn't ready yet. But I think we will look into following this PR unless anybody knows of a better approach.

mplsmitch:
This is a great discussion which highlights some much needed updates to the specification. I'm working with NABSA on a proposal to hire a consultant to help manage this project. If approved, NABSA will issue an RFP to develop a governance model and versioning scheme that will help keep things moving forward. In addition to better governance, GBFS needs to better reflect changes in the mobility industry.  NABSA has now broadened it's scope to include micro-mobility services beyond shared bicycles and the specification should support those services. What else should go into a scope of work for this RFP?  I'll also post this on the Issues page.

evansiroky:
I have significantly refactored this PR. Please see the updated PR description.

gbfs.md
```diff
@@ -173,6 +175,43 @@ email             | Optional  | A single contact email address for customers to
...
+\- description | Optional | A free text description of the vehicle
```
kanagy: Can you add what is description meant to be used for?

gbfs.md
```diff
@@ -173,6 +175,43 @@ email             | Optional  | A single contact email address for customers to
 timezone          | Yes       | The time zone where the system is located. Time zone names never contain the space character but may contain an underscore. Please refer to the "TZ" value in https://en.wikipedia.org/wiki/List_of_tz_database_time_zones for a list of valid values
 license_url       | Optional  | A fully qualified URL of a page that defines the license terms for the GBFS data for this system, as well as any other license terms the system would like to define (including the use of corporate trademarks, etc)
 
+### vehicle_types.json
+The following fields are all attributes within the main "data" object for this feed.
+
+ield Name        | Required  | Defines
```
kanagy: Nit: missing 'F'

gbfs.md
```diff
@@ -218,6 +325,38 @@ bikes             | Yes       | Array that contains one object per bike that is
 \- lon             | Yes       | Longitude of the bike. The field value must be a valid WGS 84 latitude in decimal degrees format. See: http://en.wikipedia.org/wiki/World_Geodetic_System, https://en.wikipedia.org/wiki/Decimal_degrees
 \- is_reserved     | Yes       | 1/0 value - is the bike currently reserved for someone else
 \- is_disabled     | Yes       | 1/0 value - is the bike currently disabled (broken)
+\- vehicle_type_id | Conditionally Required | The vehicle_type_id of this vehicle as described in [vehicle_types.json](#vehicle_typesjson). This field is required if the [vehicle_types.json](#vehicle_typesjson) is defined.
+\- range | Optional | If the vehicle has a motor, this value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel.
+
```
kanagy, line 461: Can we add a last_reported field here, which would optionally specify the timestamp when the bike was last reported (defaulting to feed last_updated)? It would be consistent with station_status.json, which already has last_reported.

gbfs.md
```diff
@@ -218,6 +325,38 @@ bikes             | Yes       | Array that contains one object per bike that is
 \- lon             | Yes       | Longitude of the bike. The field value must be a valid WGS 84 latitude in decimal degrees format. See: http://en.wikipedia.org/wiki/World_Geodetic_System, https://en.wikipedia.org/wiki/Decimal_degrees
 \- is_reserved     | Yes       | 1/0 value - is the bike currently reserved for someone else
 \- is_disabled     | Yes       | 1/0 value - is the bike currently disabled (broken)
+\- vehicle_type_id | Conditionally Required | The vehicle_type_id of this vehicle as described in [vehicle_types.json](#vehicle_typesjson). This field is required if the [vehicle_types.json](#vehicle_typesjson) is defined.
+\- range | Optional | If the vehicle has a motor, this value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel.
```
kanagy: current_range_meters would be a more explicit name. "current" makes it clear what range is being referred to and its nice to make the units explicit. Similarly, I'd prefer if max_range was renamed to max_range_meters.

kanagy (changes requested):

gbfs.md
evansiroky: Good idea. I added more detail.

gbfs.md
evansiroky: Oops.

gbfs.md
evansiroky: I added this field and marked it as required. I believe this might introduce a breaking change, so if I need to mark it as optional in order to be backwards-compatible, I will do so.

gbfs.md
evansiroky: Sounds good, I implemented this as you recommended.

heidiguenin:
Wanted to make sure that the folks who were engaged in this conversation earlier have the chance to see the changes @evansiroky made and provide feedback.

@sven4all @contra @afischer @efel85 @mfp22 @mplsmitch

gbfs.md
```diff
@@ -173,6 +175,43 @@ email             | Optional  | A single contact email address for customers to
...
 license_url       | Optional  | A fully qualified URL of a page that defines the license terms for the GBFS data for this system, as well as any other license terms the system would like to define (including the use of corporate trademarks, etc)
 
+### vehicle_types.json
+The following fields are all attributes within the main "data" object for this feed.
+
+Field Name        | Required  | Defines
+------------------| --------- | ----------
+vehicle_types          | Yes       | Array that contains one object per vehicle type in the system as defined below
+\- vehicle_type_id      | Yes       | Unique identifier of a vehicle type. See [Field Definitions](#field-definitions) above for ID field requirements
+\- form_factor      | Yes       | An enumerable describing the vehicle's general form factor. <br /><br />Current valid values are:<br /><ul><li>`bicycle`</li><li>`scooter`</li><li>`car`</li></ul>
```
yocontra: I think there may be a missing `moped` value here for providers like Revel/Scoot.

gbfs.md
```diff
@@ -190,22 +229,90 @@ stations          | Yes       | Array that contains one object per station in th
...
+```
 
 ### station_status.json
 
 Field Name            | Required  | Defines
 --------------------- | ----------| ----------
 stations              | Yes       | Array that contains one object per station in the system as defined below
 \- station_id          | Yes       | Unique identifier of a station (see station_information.json)
-\- num_bikes_available | Yes       | Number of bikes available for rental
+\- num_bikes_available | Conditionally required       | Number of vehicles of all types available for rental. This field is not required if the `vehicles_available` field has been defined.
```
yocontra: Thoughts on deprecating this + num_docks_available since it is bike specific? Not sure if making breaking changes are out of scope here, since versioning is still up for discussion. Seems like a lot to have two ways of specifying the same thing, but one is clearly better than the other.

yocontra:
@HeidiMG Left a few minor comments, overall looks great and a solid improvement.

gbfs.md
```diff
@@ -190,22 +229,90 @@ stations          | Yes       | Array that contains one object per station in th
...
-\- capacity        | Optional  | Number of total docking points installed at this station, both available and unavailable
+\- capacity        | Optional  | Number of total docking points installed at this station, both available and unavailable, regardless of what vehicle types are allowed at each dock. This field should only be used if every dock at the station is able to accept every vehicle type used in the system.
+\- vehicle_type_capacity        | Optional  | An object where each key is a vehicle_type_id as described in [vehicle_types.json](#vehicle_typesjson) and the value is a number representing the total docking points installed at this station, both available and unavailable for the specified vehicle type.
```
antrim: Do we need to enable a hierarchy or groups of vehicle types? e.g. are there cases where there are various defined electric bikes with different amenities, and we just need to specify capacity for electric bikes in general?

evansiroky: I don't think this is necessary as it is possible to determine the form factor and additional information about a vehicle type by looking up the data associated with a particular vehicle type ID.

gbfs.md
evansiroky: I do like the idea of versioning and would change this and rename everything that says "bike" to "vehicle", but will await guidance from the community.

gbfs.md
evansiroky: I went ahead and added `moped`. I also added the value `other` in order to allow the definition of any vehicle form factor that is not able to be categorized by any of the pre-defined form factors.

sven4all:
Good improvements, I would like to support this PR.

gbfs.md
```diff
@@ -173,6 +175,43 @@ email             | Optional  | A single contact email address for customers to
...
+{
+  "last_updated": 1434054678,
+  "ttl": 0,
+  "data": {
+    "vehicle_types": [
+      {
+        "vehicle_type_id": "abc123",
+        "form_factor": "bicycle",
+        "propulsion_types": ["human"],
+        "description": "a bicycle without a motor"
```
sven4all: I think it would be really nice if it's possible to add a link to a photo/graphic of the vehicle type. Is that something we can specify as optional field as well well? 

If we add it, I think it's a good idea to specify the resolution and format of the image.

gbfs.md
mplsmitch: Recently and number of cities ( DC, Bay Area, Portland etc.) have taken steps to introduce various types of adaptive cycles to their bike share fleets.  Additionally there are a number of examples of shared cargo bikes now in Europe. How would these types of vehicles be differentiated within the form factor and vehicle type? Existing examples are:

- Upright handcycle
- Recumbent handcycle
- Recumbent leg trike
- Recumbent trike tandem
- Side-by-side tandem bicycle
- Cargo bicycle
- Cargo tricycle

evansiroky: I think the intent of the _form_factor_ field is to have a very general classification and avoid making up a potentially endless hyper-specific list of enumerations of what a vehicle's "type" is. Therefore, all of those vehicles @mplsmitch listed would each have a _form_factor_ value of `bicycle`. For now, the place to further explain details about the vehicle would be in the _description_ field of a vehicle type object in the vehicle_types.json file. In my initial PR, I didn't include this field, but instead proposed to add many more fields with additional places to differentiate vehicles (such as adding pictures which @sven4all [brought up again](https://github.com/NABSA/gbfs/pull/136#discussion_r328438810)). However, after a discussion with @antrim and @kanagy, I removed those other fields and added the _form_factor_ field in order to try to make a minimum viable pull request to merge in.

heidiguenin:
This was one issue discussed during an in-person GBFS developers workshop held before the NABSA 2019 conference (workshop [agenda](https://docs.google.com/document/d/13uXgMF5LWYvKX7vw5Y6_SZsBmtd32VS0opEfDJFjJYo/edit?usp=sharing) and [notes](https://docs.google.com/document/d/1TXf3U7gbrJrS_r2PmTrfbJspNydAyd42xU3ctNgzNLM/edit?usp=sharing)).

Participants discussed fields that we’d like to see included in the spec in the future, but there was general agreement that this proposal is a valuable addition that makes the spec better able to represent options available without foreclosing on future additions to the spec. Other issues that came up for future work include the creation and maintenance of a centralized vehicle type definition repository and traveler privacy implications of increasingly specific vehicle attributes. 

A couple of changes proposed during the workshop include: making free_bike_status.json last_reported optional and changing "description" to "name." The intention is for last_reported to become required in the future, but we want to get the non-breaking changes in motion as soon as possible.

mattdsteele:
Coming here from another issue (#81) - are the "~High Level Changes~" in the original post on the PR up to date?

evansiroky:
I have some more incoming changes and will update the PR description after that.

evansiroky:
Ok, the PR description has been updated with the latest round of changes. I have:

- removed the `vehicle_types#description` field in favor of the `vehicle_types#name` field.
- made the `free_bike_status#last_reported` field optional
- made the `vehicle_types#max_range_meters` field conditionally required
- made the `free_bike_status#current_range_meters` field conditionally required

gbfs.md
```diff
@@ -173,6 +175,44 @@ email             | Optional  | A single contact email address for customers to
...
+\- propulsion_types | Yes | An array consisting of enumerables that describe the propulsion type of the vehicle. <br /><br />A device may have one or more values from the propulsion_type, depending on the number of modes of operation. For example, a scooter that can be powered by foot or by electric motor would have the propulsion_type represented by the array ['human', 'electric']. A bicycle with pedal-assist would have the propulsion_type represented by the array ['human', 'electric_assist'] if it can also be operated as a traditional bicycle. A car with an internal combustion engine would be represented by the array ['combustion']<br /><br />Current valid values are:<br /><ul><li>`human` _(Pedal or foot propulsion)_</li><li>`electric assist` _(Provides power only alongside human propulsion)_</li><li>`electric` _(Contains throttle mode with a battery-powered motor)_</li><li>`combustion` _(Contains throttle mode with a gas engine-powered motor)_</li></ul> This field was copied from [City of Los Angeles Mobility Data Specification](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/73995a151f0a1d67aab3d617a4693f8f81967936/provider/README.md#propulsion-types)
```
jcn: Is the "electric assist" value supposed to be `electric_assist` or `electric assist`? I see that it's inconsistent between the field definition and the example. I would suggest going with the underscore just to avoid ambiguity (i.e. with a space there's more of a possibility of it being confused for 2 different values).

gbfs.md
```diff
@@ -173,6 +175,44 @@ email             | Optional  | A single contact email address for customers to
...
+\- max_range_meters | Conditionally Required | If the vehicle has a motor (as indicated by having one or more values other than `human` in the `propulsion_types` array), this field is required. This represents the furthest distance in meters that the vehicle can travel without recharging or refueling when it has the maximum amount of energy potential (for example a full battery or full tank of gas).
+\- name | Optional | The public name of this vehicle type.
```
jcn: @evansiroky I'm curious the rationale behind making the `name` optional. My initial thought was that this might complicate things for consumers who will then need to invent language around displaying this data, but as a consumer you probably have much better insights into whether this would actually a problem or not.

gbfs.md
evansiroky: Good catch, it should be `electric_assist`. Fix incoming.

gbfs.md
evansiroky: As a consumer, I would prefer this field to be required. However, it seems that a number of systems don't have names for their vehicles. Take for example [Pacers Bikeshare](https://www.pacersbikeshare.org/how-it-works). I can't find any info on the models they call their bikes - even though they have both regular and e-bikes. So the thought was that maybe some producers don't call their bike models anything and maybe this field should be optional to reflect that. I'm open to making this required.

gbfs.md
```diff
@@ -190,22 +230,90 @@ stations          | Yes       | Array that contains one object per station in th
...
 
 ### station_status.json
 
 Field Name            | Required  | Defines
 --------------------- | ----------| ----------
 stations              | Yes       | Array that contains one object per station in the system as defined below
 \- station_id          | Yes       | Unique identifier of a station (see station_information.json)
-\- num_bikes_available | Yes       | Number of bikes available for rental
+\- num_bikes_available | Conditionally required       | Number of vehicles of all types available for rental. This field is not required if the `vehicles_available` field has been defined.
 \- num_bikes_disabled  | Optional  | Number of disabled bikes at the station. Vendors who do not want to publicize the number of disabled bikes or docks in their system can opt to omit station capacity (in station_information), num_bikes_disabled and num_docks_disabled. If station capacity is published then broken docks/bikes can be inferred (though not specifically whether the decreased capacity is a broken bike or dock)
```
kanagy: Do you need to update this description to make it clear that this is the number of disabled vehicles of any type? I.e. what you did for num_docks_available.

That said, I'm not sure why num_bikes_disabled and num_docks_disabled (and now vehicle_docks_disabled and vehicles_disabled) are part of the consumer GBFS spec. Users wouldn't care how many bikes/docks are disabled, only how many are functional. Any idea why this is in the spec (and does it make sense to make it clear in description)?

gbfs.md
```diff
@@ -218,6 +326,41 @@ bikes             | Yes       | Array that contains one object per bike that is
 \- lon             | Yes       | Longitude of the bike. The field value must be a valid WGS 84 latitude in decimal degrees format. See: http://en.wikipedia.org/wiki/World_Geodetic_System, https://en.wikipedia.org/wiki/Decimal_degrees
 \- is_reserved     | Yes       | 1/0 value - is the bike currently reserved for someone else
 \- is_disabled     | Yes       | 1/0 value - is the bike currently disabled (broken)
+\- vehicle_type_id | Conditionally Required | The vehicle_type_id of this vehicle as described in [vehicle_types.json](#vehicle_typesjson). This field is required if the [vehicle_types.json](#vehicle_typesjson) is defined.
+\- last_reported | Optional       | Integer POSIX timestamp indicating the last time this vehicle reported its status to the backend
```
kanagy: Nit: "to the operator's backend", to make clear what backend this is.

gbfs.md
evansiroky: Yes, I should do that. Change incoming.

gbfs.md
evansiroky: Ok, I'll add this.

gbfs.md
```diff
@@ -190,22 +230,90 @@ stations          | Yes       | Array that contains one object per station in th
...
+\- vehicle_docks_disabled  | Optional  | An object consisting of keys that are vehicle_type_ids as described in [vehicle_types.json](#vehicle_typesjson) and values that represent the number of empty but disabled dock points for vehicles of the respective vehicle type at the station.
+\- vehicles_available | Conditionally Required       | This field is required if the [vehicle_types.json](#vehicle_typesjson) file has been defined. This field's value is an object consisting of keys that are vehicle_type_ids as described in [vehicle_types.json](#vehicle_typesjson) and values that represent the number of vehicles of the respective vehicle type available for rental
```
gcamp: There's no way to know the `current_range_meters` for vehicles in a station. We can potentially use the `max_range_meters` from the vehicle but that might be inaccurate.

gbfs.md
```diff
@@ -190,22 +230,90 @@ stations          | Yes       | Array that contains one object per station in th
...
 \- is_returning        | Yes       | 1/0 boolean - is the station accepting bike returns (if a station is full but would allow a return if it was not full then this value should be 1)
-\- last_reported       | Yes       | Integer POSIX timestamp indicating the last time this station reported its status to the backend
+\- last_reported       | Yes       | Integer POSIX timestamp indicating the last time this station reported its status to the operator's backend
+\- vehicle_docks_available | Conditionally Required       | This field is required if the [vehicle_types.json](#vehicle_typesjson) file has been defined and the `num_docks_available` field is not defined. This field's value is an object consisting of keys that are vehicle_type_ids as described in [vehicle_types.json](#vehicle_typesjson) and values that represent the number of docks accepting vehicle returns of vehicles of the respective vehicle type. If a single dock can accept multiple types, these should be added to the count of available docks for each applicable vehicle type.
```
gcamp: We wouldn't use it ourselves in Transit but I know compagnies that use GBFS for statical analysis and prediction of dock availability. If a single dock is included in the count of multiple types that will be hard/impossible to analyze in the same way. Not a deal breaker for us but could be for other companies. I know that [Quicit](https://en.qucit.com/qucit-bike) did something similar.

gbfs.md
evansiroky: You're right. This seems like a critical design flaw. I'll work on revising my PR in light of this.

evansiroky:
Thanks, @gcamp for bringing up the matter about not having information about vehicle range in the station_status file. I have refactored the design of the station_status file in order to make it possible to provide this information.

heidiguenin:
Does # of passengers belong in MVP?

https://techcrunch.com/2019/06/04/bird-is-launching-a-two-seater-electric-vehicle-to-become-more-than-a-kick-scooter-startup/

antrim:
Is this currently an implementation candidate? Are there producers and/or consumers that are ready to implement this?

evansiroky:
I just reviewed my PR again and added in some clarifying comments about when certain fields are required. I think this PR is at a good point to be an implementation candidate. We at IBI Group would be interested in being a consumer that could implement this proposal.

antrim:
Ok. @evansiroky I'd support calling a vote if you want to go ahead and do that. I just did that on the "versioning" proposal: https://github.com/NABSA/gbfs/pull/188#issuecomment-554215761

evansiroky:
Since some of my PR includes JSON examples, maybe it would be best to wait until the feed versioning PR has been merged in, so the example JSON can be updated with versioning information.

antrim:
I'd suggest calling the vote. We're going to have a lot of overlapping and interlinked PRs right now while we're catching up with previous GBFS proposals. We can make editorial changes to the specification that don't change its meaning to sort changes into the versioning scheme.

evansiroky:
Should we still call a vote since next week has a major holiday? Maybe we should wait until the week after Thanksgiving to call a vote.

heidiguenin:
I'm in favor of calling the vote. What do you think about extending the window 2 extra business days out to accommodate the U.S. holiday?

sven4all:
I am still in doubt about "propulsion_types" I think allowing multiple values here is making things unnecessary complex. I think it's better to let the operator decide which `human`, `electric assist`, `electric` or 'combustion`. There is normally only one way the vehicle is meant to be used.

I made yesterday some integration between different systems and decided to not allow multiple values for "propulsion_type" in my own system. Because I have no idea how I should present that to the user. Also I think that this field will be unambiguously used. A scooter is electric but you can humanly power it, but that is not the way the scooter is intended to be used. Maybe one operator assigns ['human', 'electric'] and another operator assigns ['electric'] while they are using the same hardware. I think this leads only to confusion and making it simpler by having to select the main 'propulsion_type' it's more clear.

evansiroky:
That field was copied directly from the [mobility data specification's definition](https://github.com/openmobilityfoundation/mobility-data-specification/blob/dev/provider/README.md#propulsion-types). From a trip-planning perspective I think the way I would use this array is to indicate if it is possible to walk with a motorized vehicle in the event that a user may want to traverse some stairs or if they are able to push a vehicle up a short hill in the event that it does not have sufficient power to climb the hill. For example, if `['human', 'electric']` were provided, then I think a trip planner would allow traversal of stairs. But I definitely can see your point about how it could lead to confusion and inconsistencies. It seems like a likely scenario that most operators would merely choose one field. Furthermore, SAE recently released [Taxonomy and Classification of Powered Micromobility Vehicles J3194_201911](https://www.sae.org/standards/content/j3194_201911/) where they only classify a "power source" as either "electric" or "combustion".

It seems like a good idea to be consistent with other standards, but I'm open to making the propulsion type have a single value to indicate the primary power source and also adding a boolean field to indicate whether it would be feasible to travel with a vehicle over stairs or up short steep hills.

barbeau:
>I'm in favor of calling the vote. What do you think about extending the window 2 extra business days out to accommodate the U.S. holiday?

GTFS follows [the rule](https://github.com/google/transit/blob/master/gtfs/CHANGES.md) "Vote lasts the minimum period sufficient to cover 7 full calendar days and 5 full Swiss business days. "

So yes, I'd suggest we follow this.

sven4all:
> It seems like a good idea to be consistent with other standards, but I'm open to making the propulsion type have a single value to indicate the primary power source and also adding a boolean field to indicate whether it would be feasible to travel with a vehicle over stairs or up short steep hills.

I think that this is a more elegant solution! With this suggestion you make the usecase far more explicit (clearer for operators but also for consumers of GBFS).

evansiroky:
Ok, I just changed the `propulsion_types` field to `propulsion_type` and added the `is_carryable` field. I also updated the PR description to list all changes at this point in time.

evansiroky:
I just updated some of the example files I made to include the now required `version` field in each file.

I hereby call a vote on this proposal. Since this week has some holidays in certain countries, voting will be open until 11:59PM UTC on December 6th.

Please vote for or against this proposal to allow vehicle type definitions, and include the organization for which you are voting in your comment.

kanagy:
Just one other find: Is vehicle_docks_disabled ever useful for anything compared to just having existing num_docks_disabled? First wondering why consumer apps would need to know about disabled docks. Second, not sure why splitting that up by vehicle type is useful (especially since the split up amounts can add up to more than num_docks_disabled).

kanagy:
One other suggestion: can you keep num_docks_available and num_bikes_available required (meaning total number of vehicles / docks at the station)? 

First it would keep the spec simpler and easier to understand what is required. Second, making it conditionally required might be breaking compatibility: e.g. existing producers can start setting vehicle_docks_available and stop setting num_docks_available; that's correct according to the spec, but it will break consumers who don't implement using vehicle_docks_available yet.

madupras:
+1 from PBSC

gbfs.md
```diff
@@ -269,28 +245,34 @@ Example:
...
-\- num_docks_available | Yes       | An object consisting of keys that are vehicle_type_ids and values that represent the number of docks accepting vehicle returns of vehicles of the respective vehicle type
-\- num_docks_disabled  | Optional  | An object consisting of keys that are vehicle_type_ids and values that represent the number of empty but disabled dock points for vehicles of the respective vehicle type at the station. This value remains as part of the spec as it is possibly useful during development
+\- vehicle_docks_available | Conditionally Required       | This field is required if the [vehicle_types.json](#vehicle_typesjson) file has been defined and the `num_docks_available` field is not defined. This field's value is an object consisting of keys that are vehicle_type_ids as described in [vehicle_types.json](#vehicle_typesjson) and values that represent the number of docks accepting vehicle returns of vehicles of the respective vehicle type. If a single dock can accept multiple types, these should be added to the count of available docks for each applicable vehicle type.
```
ahyatt: It seems like this change makes it difficult for clients to understand how many total docks are available.  For example, let's say the station accepts bikes, e-bikes, and scooters.  There are 5 bike docks, 5 e-bike docks, and 5 scooter docks available.  So are there 5 available docks if you have a bike or e-bike, or 10?  How would you know that the docks are specific in what they can accept or not?

Additionally, this requires knowledge from another feed, which makes it especially difficult to understand the dock situation.

antrim: If the optional `station_information`.`capacity` field was provided, then consumers could know the total capacity. I think  appropriate spec here depends on the needs of consumers. If consumers need more precision, then the proposed `vehicle_docks_available` object could instead include an array of `vehicle_type_id`.

What do you think?
@kanagy @gcamp

evansiroky: I've also been wondering this and do believe it would be challenging to make a UI that shows dock availability according to vehicle type. Overall it does seem like it would be useful to provide an overall metric of the total number of docks available, but I still think it's critical to include vehicle-specific dock availability info as well. 

Looking at how I worded the spec, I think it would make sense to make a change to always require the `station_information.json#capacity` and `station_status.json#num_docks_available` fields and indicate that it always represents the total amount of docks regardless of whether certain docks only accept certain vehicle types. In a system that omits the vehicle_types.json file, this would indicate the number of docks at the station (according to `station_information.json#capacity`) and those that are currently available (according to `station_status.json#num_docks_available`). And in systems that include the vehicle_types.json this could represent the total number of docks/current availability of docks regardless of the compatibility of each dock with each vehicle type.

evansiroky:
@kanagy

> Just one other find: Is vehicle_docks_disabled ever useful for anything compared to just having existing num_docks_disabled? First wondering why consumer apps would need to know about disabled docks. Second, not sure why splitting that up by vehicle type is useful (especially since the split up amounts can add up to more than num_docks_disabled).

I was wondering the same. We don't use information about the disabled docks or bikes.

evansiroky:
@kanagy 

> One other suggestion: can you keep num_docks_available and num_bikes_available required (meaning total number of vehicles / docks at the station)?
> 
> First it would keep the spec simpler and easier to understand what is required. Second, making it conditionally required might be breaking compatibility: e.g. existing producers can start setting vehicle_docks_available and stop setting num_docks_available; that's correct according to the spec, but it will break consumers who don't implement using vehicle_docks_available yet.

See my other comment here about proposing some changes to include this info. https://github.com/NABSA/gbfs/pull/136#discussion_r353937159 Also, in an email sent out recently, a roadmap suggested that this PR would be included as part of a 2.0 release. Therefore, I don't think we have to worry so much about making breaking changes.

gbfs.md
ahyatt: I agree with making `station_status.num_docks_available` always required.  In addition, you may want to change `vehicle_docks_available` so you can represent exclusivity better.  For example: `vehicle_docks_available: [{'vehicle_type': ['ebike', 'bike'], 'count': 5}]`.

kanagy:
> @kanagy
> 
> > Just one other find: Is vehicle_docks_disabled ever useful for anything compared to just having existing num_docks_disabled? First wondering why consumer apps would need to know about disabled docks. Second, not sure why splitting that up by vehicle type is useful (especially since the split up amounts can add up to more than num_docks_disabled).
> 
> I was wondering the same. We don't use information about the disabled docks or bikes.

Just noticed that the existing spec for num_docks_disabled says "This value remains as part of the spec as it is possibly useful during development". This sounds pretty vague to replicate it per-vehicle. Hence I would remove vehicle_docks_disabled from the proposal unless there's someone who explicitly sees a use for it.

evansiroky:
Seeing that there is only 1 vote in favor and a few people requesting changes, I am canceling the voting so I can implement the requested changes and resolve a merge conflict.

evansiroky:
All right, I have updated how information about capacity and availability of vehicles is modeled at stations. It seems like it would also make sense to incorporate the language and field requirements that are proposed in https://github.com/NABSA/gbfs/pull/195. I'm inclined to merge that branch into this one before it gets merged to master but will wait for a thumbs up or something.

evansiroky:
Also just made the field `num_bikes_availabe` required again as per @kanagy's suggestion.

evansiroky:
Also, to through a curveball out there, since the GBFS now has versioning and this PR will likely end up in a new major release, would this possibly be a good time to rename all `bike` references to `vehicle`? ie `num_bikes_available` becomes `vehicles_available` and `free_bike_status.json` becomes `vehicle_status.json` or possibly even `GBFS` becomes `GVFS`?

heidiguenin:
Thanks, Evan - that's a good question! We had a workshop for European GBFS stakeholders this past week and got some feedback on both the minimum viable proposal and the longer term needs for vehicle type definitions. We'll pull together all of those notes and share them here in the coming days, too!

ahyatt:
Evan, the new `vehicle_docks_available` looks much better, thanks for making the change.  I don't think I have any significant objections at the moment.

kartofel:
Thanks for making the changes Evan. I'm happy with starting the vote again.

Note: you might want to remove stations.vehicles.bike_id given the sentiment in https://github.com/NABSA/gbfs/pull/147 Also, its not really needed for this proposal.

kanagy:
Thanks for making the change Evan. This looks good to me and can do another vote.

Note: you might want to remove stations.vehicles.bike_id given the sentiment in https://github.com/NABSA/gbfs/pull/147 Also, it is not really needed in this proposal.

Empty2k12:
In addition to `current_range_meters`, I would also propose `battery_available`. Both should be optional. This field could be used when `current_range_meters` can not accurately be calculated or is not available. This info could be displayed to the user directly and be shown in a UI, or be used to make assumptions about the remaining range or defects about of the bike.

I am also strongly in favor of keeping `stations.vehicles.bike_id`, since there's privacy concern with station-based systems and this info can be useful for consuming applications. A local provider in my city added `bike_id`s to their station endpoint and started showing it in their user facing application.

evansiroky:
> In addition to `current_range_meters`, I would also propose `battery_available`. Both should be optional. This field could be used when `current_range_meters` can not accurately be calculated or is not available. This info could be displayed to the user directly and be shown in a UI, or be used to make assumptions about the remaining range or defects about of the bike.

For trip planning purposes the `current_range_meters` is critical for knowing the distance a vehicle could travel. However, I can understand some operators may not want to over or under-report this field. If this field is optional, I would imagine that to be safe some trip planners would assume the vehicle has a range of 0 since they wouldn't want to offer a trip on a vehicle where there is uncertainty about how far the vehicle could travel.

I don't see what value `battery_available` adds here given that there is already the `vehicle_types#propulsion_type` field. In the `station_status#num_bikes_available` it states that this value represents the "Number of functional bikes physically at the station." Perhaps some additional language should be added to the free_bike_status.json file to indicate that each entry is supposed to be a functional vehicle.

NeoLegends:
> I don't see what value battery_available adds here given that there is already the vehicle_types#propulsion_type field.

I think there is a misunderstanding here. `battery_available` would show the charging percentage of the battery and not whether the vehicle is battery-powered or not. IMO the name of the field is a little unclear. The field would indeed be very useful in the case where the range estimation is inaccurate (e. g. in a mountainous city where the maximum travel distance greatly depends on the route).

> I am also strongly in favor of keeping stations.vehicles.bike_id, since there's privacy concern with station-based systems and this info can be useful for consuming applications. A local provider in my city added bike_ids to their station endpoint and started showing it in their user facing application.

@Empty2k12 it's a little unclear to me why you mention privacy concerns but advocate for __keeping__ the field (IMO privacy concerns would advocate for removing the field). Could you elaborate further?

Empty2k12:
> @Empty2k12 it's a little unclear to me why you mention privacy concerns but advocate for keeping the field (IMO privacy concerns would advocate for removing the field). Could you elaborate further?

I just noticed a type in my response. There is **no** privacy-concern with docked systems since only the riding history of the bike, not the rider could be inferred.

Empty2k12:
> IMO the name of the field is a little unclear. The field would indeed be very useful in the case where the range estimation is inaccurate (e. g. in a mountainous city where the maximum travel distance greatly depends on the route).

This is what I meant, sorry for choosing such an unclear variable name. Maybe we should call it `remanining_battery_percent` or something?

evansiroky:
I understand now. Thanks for the clarification. 

In the proposed `vehicle_types.json` file, there is already a conditionally required field called `max_range_meters`. It was assumed that this field could act in conjunction with the `station_status.json#stations#vehicles#current_range_meters` and `free_bike_status.json#bikes#current_range_meters` fields in order to determine a "percent energy potential remaining" statistic. Perhaps a clarifying note could be made that the range fields assume travel along flat pavement at cruising speed?

Empty2k12:
Ah that makes sense. But why have a difficult calculation for battery percentage when these values exist already in the backends of e-mobility providers and could be easily supplied to GBFS? 
This would also make sense when modeling other means of transportation like E-Scooters with `free_bike_status.json`. Users already use the battery percentage provided by bike rental providers or e-scooter companies to estimate the range. If anything this additional information can be used together with `station_status.json#stations#vehicles#current_range_meters` to give an even better estimation.

I am not sure if assuming flat pavement at cruising speed is a good idea, since there's so many more factors to an accurate range calculation. How often does the user stop / begins riding again? What's the elevation of a chosen route? How does the outside temperature affect battery? ...

For the e-bike company we deal with, as a user it's typically better to use the battery percentage and guess. 80%: Enough for multiple rides between stations, 40%: Good for at most two rides, 20%: Maybe not take this bike on a long ride. Same applies to e-scooters.

So I still think there should be 

`station_status.json#stations#vehicles#remanining_battery_percent`
`free_bike_status.json#bikes#remanining_battery_percent`

in addition to

`station_status.json#stations#vehicles#current_range_meters`
`free_bike_status.json#bikes#current_range_meters`

All 4 being optional at the discretion of the mobility provider.

evansiroky:
I understand the argument and would like to offer a friendly amendment to your proposal.

First off, I think we should rename `remanining_battery_percent` to `remaining_energy_potential_percent`. This way, the field is agnostic to the energy source that the vehicle has and this standard will be able to make sense for either vehicles with a combustion engine or electric engine or who knows what future energy source. 

I'd also like to push back a bit on not providing an estimate of the remaining range. Most countries have laws that require car manufacturers to publish info about miles per gallon (or liters of fuel per kilometer) that a user can expect a car to experience during regular use. Certainly, some users will get worse miles per gallon and others may get better results, but at the very least it's a good rough estimate. It seems like this should be possible to do with battery-powered vehicles as well. 

As a GBFS data consumer wanting to reliably calculate trip plans with this data standard, having a way to calculate the remaining range of a vehicle is a must-have feature. If neither field is provided, there is no way to know how far a vehicle could travel and to play it safe, the trip planner might not even show a vehicle as an option since it could have no more fuel/battery. 

Another option to consider would be to keep just the `remaining_energy_potential_percent` field, remove the `cur/max_range_meters` fields and then modify the `vehicle_types.json` file to include more detailed information about the vehicle's total possible energy potential. Fields to add could be `gas_tank_size_liters` and `meters_per_liter` gas consumption for vehicles with combustion engines or `battery_watt_hours` and `meters_per_watt_hour` for vehicles with electric engines respectively. This could be simplified to be the fields `maximum_energy_potential` and `meters_per_energy_potential_unit` which would have corresponding units assigned based on the `propulsion_type` value.

So in summary here are the changes I'm proposing to this current PR:

- Remove the field `vehicle_types.json#max_range_meters`
- Add a new conditionally-required field called `maximum_energy_potential` to `vehicle_types.json`. This field is required if the `propulsion_type` field is one of `electric_assist`, `electric` or `combustion`. If the `propulsion_type` field is `electric` or `electric_assist` the unit type for this field is watt-hours. If the `propulsion_type` field is `combustion` then the unit type is liters.
- Add a new conditionally-required field called `meters_per_energy_potential_unit` to `vehicle_types.json`. This field is required if the `propulsion_type` field is one of `electric_assist`, `electric` or `combustion`. If the `propulsion_type` field is `electric` or `electric_assist` the unit type for this field is meters per watt-hour. If the `propulsion_type` field is `combustion` then the unit type is meters per liter.
- Remove the field `station_status.json#stations#vehicles#current_range_meters`
- Add a conditionally required field `station_status.json#stations#vehicles#remaining_energy_potential_percent`. This field is required if the vehicle type's `propulsion_type` field is one of `electric_assist`, `electric` or `combustion`.
- Remove the field `free_bike_status.json#bikes#current_range_meters`
- Add a conditionally required field `free_bike_status.json#bikes#remaining_energy_potential_percent`. This field is required if the vehicle type's `propulsion_type` field is one of `electric_assist`, `electric` or `combustion`.

barbeau:
>I'd also like to push back a bit on not providing an estimate of the remaining range. Most countries have laws that require car manufacturers to publish info about miles per gallon (or liters of fuel per kilometer) that a user can expect a car to experience during regular use. Certainly, some users will get worse miles per gallon and others may get better results, but at the very least it's a good rough estimate. It seems like this should be possible to do with battery-powered vehicles as well.

I agree that producers providing an estimate of remaining range is important. The's the primary piece of vehicle info that Google Maps is surfacing at the moment in their Lime integration:

image

I think this is important for two reasons:
1. Converting between remaining charge and range isn't trivial - As a user you need to have a good amount of personal experience with specific types of vehicles to translate remaining charge to range. And consumers would also need a good amount of information about the vehicle (age of battery, etc.) to estimate this correctly. IMHO the producer knows the most about their vehicles and is in the best position to provide some kind of estimate of range (even if it's just on flat terrain). Which leads to...
1. Consistency between consumers - In GTFS there was an analogous situation where initially transit agencies only wanted to publish scheduled arrival times at timepoint stops, and let consumers figure out when the bus was scheduled to arrive at the in between stops. This became a problem when you opened different apps and they told you different scheduled times when the bus was scheduled to arrive. Similarly, I would think that GBFS producers would want estimated range information about their vehicles to be consistent between apps.

antrim:
Are there data consumers that would use the energy potential, available energy, and meters per energy unit fields in any way other than to calculate the available range?

On their own, those fields still wouldn't solve the problem of knowing how terrain changes range. Consuming applications would need more information about vehicle characteristics.

If the additional fields are not necessary right now, the simpler approach (a range field) might be optimal for now. Other fields could always be added later -- and this discussion is really useful to chart that trajectory in advance.

What should we include in a minimum viable proposal that is ready to be implemented in software?

yocontra:
@antrim Correct me if I'm wrong (not an expert on this at all) but:

- If you have the total energy available in joules + the vehicle weight you could get the range estimate. This depends on the added weight of the rider + anything they are carrying as well, but it would be close enough if you used an average weight add in the formula.
- If you know those two variables, you can also take any arbitrary path (distance + elevation) and calculate the amount of work (in joules) needed to travel that route. If a user is trying to route from A to B a mapping app could filter out vehicles that don't have the available energy automatically.

This is how some electric cars give their range estimates based on your route if you are using GPS, I think the same could work if all of those pieces of information are present. By using joules as the measurement instead of breaking it down into a bunch of different pieces that have loose relation (percentage, mAh, motor efficiency, etc.) you let the vendor be concrete about the actual energy capacity, and this could adjust (from the operators side) based on the specific motor or age characteristics of a vehicle. This boils down to these fields:

- `energy_capacity` (Total joules possible)
- `energy_available` (Total joules available for use)
- `weight` (Weight of the vehicle in kg)
- `range_estimate` (Estimated range in meters at 0 elevation gain with an average weight add, this would basically exist only for convenience if you don't want to use the other three attributes for calculating this yourself for a given route)

This is probably out of scope for this ticket, but thought I would mention it while the topic is being discussed.

mplsmitch:
I'm also not an expert on this but I can tell you that in Minneapolis, the range of electric vehicles in winter is about half of what it is in the summer.

kanagy:
My 2c: I'm not super keen on consumers having to calculate the battery range themselves based off different factors. Other than this complicating GBFS (now we have to all agree on how to model calculation of battery ranges and this proposal drags on), I think provider would have a more accurate idea of what's the current range (more inputs and data; e.g. it can be seasonal or also depend on how old the battery is and how well it will hold capacity).

Second issue is that ideally the battery range that shows up in a consumer app (e.g. Google Maps) should match the battery range in the provider app for the same vehicle. Otherwise we're confusing user with 2 different numbers (if we're doing the calculation ourselves they won't agree).

yocontra:
@kanagy Yep, I agree - the other three attributes exist solely for displaying charge %, eliminating scooters that are not viable for a given route, and some other use cases.

I think for this proposal everything is out of scope except for the range estimate, the others are nice to have for more advanced routing cases. Not trying to cram more into this proposal, a separate one can exist for other range related information if there is demand for it IMO.

evansiroky:
I'd like to echo what others have said about the calculations of vehicle range. Huge variations can occur in three broad categories: 

1. Variations in vehicle quality (such as drivetrain efficiency, vehicle weight, aerodynamics, and tire quality)
2. Environmental factors (such as gradients, road quality, and atmospheric conditions like wind and air pressure)
3. User preferences (such as a user's desired speeds, a user's aerodynamics, and route choice preferences)

Taking all these factors into consideration can result in very different energy uses and vehicle ranges. Therefore, I think it is best to have a consistent experience among consumers to be provided with a generic range per vehicle. I am fine with this PR as is, but I will point out that in the proposal to add the `meters_per_energy_potential_unit` and `maximum_energy_potential` fields if these fields are multiplied together, it should result in the same value that the `max_range_meters` field would have provided. 

An added benefit of having additional information about the total energy and energy efficiency is that it could prove useful to certain consumers wanting to show the energy-efficiency of vehicles. For example, I can imagine a car rental agency using these fields to publish information about the fuel/energy efficiency of cars. Typically in the United States, people who rent cars will need to pay costs in addition to rental costs to refuel/recharge a vehicle, so it can be nice for end-users of applications built by GBFS consumers to have some kind of expectation about how much fuel/energy might be used in the course of using a rental.

Empty2k12:
> I am fine with this PR as is, but I will point out that in the proposal to add the `meters_per_energy_potential_unit` and `maximum_energy_potential` fields if these fields are multiplied together, it should result in the same value that the `max_range_meters` field would have provided.

I think adding those values would be a good thing for this PR, as all current use cases (displaying percentage where it available today, and displaying distance where it is available) could then be modeled. 

For reliable distance calculations, we should have another discussion and figure out a good way of providing all those additional details to consumers.

evansiroky:
I have modified the PR as proposed in https://github.com/NABSA/gbfs/pull/136#issuecomment-582573214. I hope this is well received by all interested parties and that providers are ok with providing estimates of energy efficiency for their vehicles.

In order to move on this proposal, I think we should try to move forward with the current state of this PR as the minimum viable proposal to include in the GBFS specification.

I hereby call a vote on this proposal. Voting will be open for 7 full days, until 11:59PM UTC on February 28th.

Empty2k12:
I vote in favor of this proposal for [MIA map intelligence agency](https://github.com/mapintelligenceagency/).

sven4all:
I vote in favor for this proposal on behave of [CROW](https://www.crow.nl/english-summary).

kanagy:
I vote in favor on behalf of Google Maps. But I would like to hear from producers whether modelling max and current vehicle distance using the energy potential fields makes sense (and would be implemented). If it doesn't make sense, we'd prefer having two separate "current_range_meters" and "current_energy_potential_pct" fields.

Nit: should station_status.vehicle_docks_available.vehicle_types be called vehicle_type_ids to match the naming of vehicles.vehicle_type_id and the referred field name?

gcamp:
I would have the same question as @kanagy regarding the range vs energy potential fields. 

Did anyone commit to implement those changes? We will implement those but there are few fields that we won't use that maybe shouldn't be in the proposition (mainly thinking of `is_carryable`).

heidiguenin:
To the question @gcamp raised around `is_carryable` - we did hear from several stakeholders during our European GBFS developers' workshop that it should be removed from MVP.

evansiroky:
Seems like the vote for this has failed since no producers voted yes on this. I have removed the `is_carryable` field. What else needs to happen to get the support of producers for this change to the specification?

heidiguenin:
Not to call y'all out, but we could use your input! Can we get some feedback on the question @evansiroky raised above?

@bhandzo @charlesjump @johnpena @rowillia @madupras @maduprasPBSC @lindarand @dsgermain

bhandzo:
I agree with others who've posted above that providers know best how to calculate remaining range for a vehicle under varying conditions, and should be getting better at this over time. So leaving range to a simpler value of 'range remaining' makes more sense to me than having others do their own calculations using a potential energy value.

kanagy:
Nit: if you plan to keep remaining_energy_potential_percent, can you clarify whether this should be a value between 0 and 100 or value between 0 and 1?

evansiroky:
I have gone ahead and reverted back to using `vehicle_types.json#max_range_meters` and `station_status.json#stations#vehicles#current_range_meters`/`free_bike_status.json#current_range_meters`. Also, I have incorporated @kanagy's [suggestion](https://github.com/NABSA/gbfs/pull/136#issuecomment-590651465) to rename to the field `station_status.json#stations#vehicle_docks_available#vehicle_type_ids`.

If any other producers can chime in with any other feedback, that would be appreciated.

Empty2k12 (approved):
This looks good to me now 🚀

evansiroky:
Seeing no further feedback, I think it is a good time to call a vote again to try to see if this proposal is can be accepted.

I hereby call a vote on this proposal. Voting will be open for 7 full days, until 11:59PM UTC on April 21st.

yocontra:
Voting +1 for Stae - great work @evansiroky 👍

Empty2k12:
I vote in favor of this proposal for [MIA map intelligence agency](https://github.com/mapintelligenceagency/).

kanagy:
Voting +1 for Google Maps. Thanks for the work Evan.

madupras:
+1 from PBSC

cmonagle:
Voting +1 on behalf of Transit

bhandzo:
Voting +1 on behalf of Bird. Thanks @evansiroky for pushing this across the line.

gbfs.md
```diff
@@ -256,6 +259,53 @@ Field Name | Required | Type | Defines
...
+---|---|---|---
+`vehicle_types` | Yes | Array | Array that contains one object per vehicle type in the system as defined below
+\- `vehicle_type_id` | Yes | ID | Unique identifier of a vehicle type. See [Field Definitions](#field-definitions) above for ID field requirements
+\- `form_factor` | Yes | Enum | The vehicle's general form factor. <br /><br />Current valid values are:<br /><ul><li>`bicycle`</li><li>`car`</li><li>`moped`</li><li>`other`</li><li>`scooter`</li></ul>
+\- `propulsion_type` | Yes | Enum | The primary propulsion type of the vehicle. <br /><br />Current valid values are:<br /><ul><li>`human` _(Pedal or foot propulsion)_</li><li>`electric_assist` _(Provides power only alongside human propulsion)_</li><li>`electric` _(Contains throttle mode with a battery-powered motor)_</li><li>`combustion` _(Contains throttle mode with a gas engine-powered motor)_</li></ul> This field was insipred by, but differs from the propulsion types field described in the [City of Los Angeles Mobility Data Specification](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/73995a151f0a1d67aab3d617a4693f8f81967936/provider/README.md#propulsion-types)
```
charlesjump: Please update this description from City of Los Angeles Mobility Data Specification --> Open Mobility Foundation and the URL updated to https://github.com/openmobilityfoundation/mobility-data-specification

charlesjump (changes requested):

charlesjump:
Wouldn't this be the time to change from `bike_id` to `vehicle_id`?

heidiguenin:
@charlesjump I agree that it's time to vote on changing bike_id to vehicle_id, but right now this PR represents a non-breaking change.  Is it worth making this whole PR a breaking change to make the transition to vehicle_id now, or could we open a separate PR for that breaking change and go ahead and get this one in process?

charlesjump:
@heidiguenin which version of GBFS will this change be included?

heidiguenin:
@charlesjump - Both #219 (Geofencing, virtual station, and dockless support) and vehicle type definitions, if passed, would be included in v2.1.  I believe that they're mostly already implemented out in the field, too, so they should be able to move from release candidate to official fairly quickly.

Changing bike_id to vehicle_id, once passed in a vote, would be added to the v3.0 release candidate that currently includes #182  - requiring license_url.

heidiguenin:

Votes in favor:
Stae
MIA
Google Maps
PBSC 
Transit
Bird

No votes against.

We'll be merging and creating v2.1RC in the coming days. If you're implementing this currently or planning on it in the future, please let us know. Thanks all!

heidiguenin:
As we're getting ready to tag the new release - is there anyone who is implementing this PR yet?

kanagy:
> As we're getting ready to tag the new release - is there anyone who is implementing this PR yet?

Google is starting implementation with Lime. We plan to migrate our custom fields to use the official vehicle_types.json for free floating bikes and scooters.

PR #141, opened 2019-03-30
update nicridemn gbfs root

yocontra:
SSL certificate and domain expired on the current one - this new endpoint is the one used on their website to display stations on the map and seems to be current.

As a note - their system is closed for the season, so the endpoints don't have any data in them at the present.


jcn:
Looks like niceridemn.com is the new domain, so I just switched over to that instead of using the 8d.com one. Thanks for the PR.

PR #143, opened 2019-04-03
added bikesharing services from urbansharing

hpl002:
Added
 - Bergen City Bike
 - Oslo City Bike
 - Trondheim City Bike
 - Just Eat Cycles, Edinburgh

CSV validated in https://csvlint.io/


jcn (changes requested):
The `System ID` column should be the `system_id` from the `system_information` GBFS feed for each system, not "urbansharing." Please update and I will re-review.

jcn:
Thanks!

PR #147, opened 2019-04-30
Rotate bike_id on free_bike_status

morganherlocker:
This PR changes the description of bike_id to a random id, rotated after each trip. This is a non breaking change, as some providers already rotate to a new random id daily.


jcn:
Regardless of where we end up on the issue of `bike_id`, I would argue that this is a breaking change to the spec as it means that vendors that do _not_ rotate their ids will then be out of compliance.

yocontra:
If the primary purpose of having a bike_id is for booking handoff, wouldn't it be clearer to rename this field to `booking_token` or similar to be clearer about its purpose and allow providers to have their own hashing/rolling method, with a baseline being at least once per trip? This also clarifies to consumers that this should not be used as an index or unique vehicle ID, and is ephemeral without having to read the spec deeply.

heidiguenin:
This was one issue discussed during an in-person GBFS developers workshop held before the NABSA 2019 conference (workshop [agenda](https://docs.google.com/document/d/13uXgMF5LWYvKX7vw5Y6_SZsBmtd32VS0opEfDJFjJYo/edit?usp=sharing) and [notes](https://docs.google.com/document/d/1TXf3U7gbrJrS_r2PmTrfbJspNydAyd42xU3ctNgzNLM/edit?usp=sharing)).

During the workshop, we discussed GBFS dataset producer and consumer uses of bike_id and if/how bike_id is currently rotated. We learned that some producers currently rotate bike_id; we learned of one example where the rotation of bike_id was required by a city as a requirement of permitted operations. We also heard about GBFS producers removing bike_id from their GBFS datasets and being required by cities to add them back in as a requirement of their permits. There was speculation about why producers may have done this, and privacy was discussed as a likely factor.

Identified uses of bike_id include:
Booking - bike_id is used for in-app booking
Operations - bike_id can be used to identify stale bikes
Regulatory compliance - bike_id may be used for counting bikes in support of regulatory enforcement

Because GBFS is concerned with traveler-facing information, we are primarily concerned with the booking use case, though participants did discuss how to address the other two uses cases assuming that bike_id would be rotated regularly in the future. Ultimately, the inconvenience of changing current use cases needs to be weighed against current privacy concerns, and participants generally agreed that the rotation of bike_id was the simplest solution for now. 

For booking, the main purpose of bike_id is as a hand-off token. With the deeplinks PR #25  assumed to be moving forward in the near future, it’s possible that bike_id may not even serve that purpose in the future. Participants discussed making bike_id an optional field, but ultimately to reduce the possibility of this being a breaking change, the consensus was to rotate bike_id *at least* after every trip (some operators suggested they’d prefer to rotate more often for operational reasons) and to provide best practice guidance suggesting methods for rotation and for reversibility for operators.

This change is intended to reduce the exposure of private data & does not guarantee anonymization. This proposal is part of the series of the minimum viable proposals we’re hoping to move forward in the coming weeks to get the spec better in line with industry practices and concerns, but we anticipate the community continuing to examine this issue in more depth on an on-going basis.

sven4all:
One of the use cases of GBFS is monitoring of bike-sharing operators by
municipalities (It's for example required to provide a GBFS in the MDS
specification
https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/provider
).

I respect the wish of bike-sharing operators to not communicate their
bike_ids in public feeds. But it would be a good idea to have the
possibility to communicate a static bike_id in a GBFS feed that is private
for municipalities / governments to not break with current implementations
of monitoring software.

Op di 8 okt. 2019 om 00:33 schreef Heidi G <notifications@github.com>:

barbeau:
@sven4all In the workshop we reinforced the division of responsibilities between GBFS and MDS.  GBFS is intended to be used for traveler-facing information, and MDS is intended to be used for regularly monitoring and enforcement. Some may that have used GBFS for regulatory purposes, but going forward privacy concerns outweigh alternate use cases like this for GBFS that fall outside its primary purpose.  Hence the need to rotate bike_ids.

If MDS doesn't fit the exact use case you're looking at for monitoring, I'd suggest bringing that up on the MDS repository as a new issue - https://github.com/CityOfLosAngeles/mobility-data-specification.

In the workshop we did discuss fleshing out this PR with the exact expected steps for rotating IDs to make expectations clearer to producers and consumers, and that work remains to be done.

morganherlocker:
@HeidiMG thanks for the informative summary for those of us following along online. I have committed the change in language to specify that bike_id needs to be rotated after every trip, _"at minimum"_.

heidiguenin:
@morganherlocker Want to go ahead and call a vote on this now that versioning has passed?

tmontes:
> @sven4all In the workshop we reinforced the division of responsibilities between GBFS and MDS. GBFS is intended to be used for traveler-facing information, and MDS is intended to be used for regularly monitoring and enforcement. Some may that have used GBFS for regulatory purposes, but going forward privacy concerns outweigh alternate use cases like this for GBFS that fall outside its primary purpose. Hence the need to rotate bike_ids.

IIUC, @sven4all could be referring to authenticated (and encrypted?) access to `free_bike_status.json` feeds. Under those circumstances, mobility providers do know who they're "talking to" and could (should, IMHO) be ready to provide more precise information.

In the case of the City of Lisbon, most `free_bike_status.json` feeds are both authenticated and transferred over HTTPS such that the city can monitor public space utilisation.

Regarding privacy, and the particularly strict requirements of GDPR in Europe, there should be no limitations given that the City itself **is GDPR compliant** in the ways it processes and stores information.

My 2c.

sven4all:
@tmontes that is exactly what I mean. I think that it would be helpful if GBFS supports and standardise this usecase as well (inclusive authentication and encryption). We use GBFS really like a MDS light variant (it's easier for operators to implement and less data has to be handed over). If it's decided that free_bike_status should have rotating bike ids it will lead to a lot of confusion for many operators when governments are still requiring static bike_ids in the free_bike_status.json endpoint. 

Would it be an idea to introduce a monitor_bike_status.json exclusively for government access inspired by free_bike_status.json?

jcn:
In the interest of not duplicating too much data between MDS and GBFS, while respecting the use cases of both (and noting that MDS currently requires providers to produce a GBFS feed), I wonder if a solution might be to leave `bike_id` as the booking token (which is how people are generally using it), and adding something like a `immutable_bike_id` which would only be provided for authenticated feeds. It must be excluded in public feeds.

I would only propose that something like that even be considered if a producer was willing to actually test it out, but it could serve both needs in a relatively lightweight way.

heidiguenin:
We need to have a broader community conversation about how to avoid breaking the existing use cases of bike_id and what guidance to provide around when feed authorization is appropriate and how fields might be different in that case. But we also need updated spec documentation as soon as possible to ensure that we create a new understanding for operators and regulators that **publishing stable bike_id in a publicly available GBFS dataset is not GBFS compliant.**  Everything that we have heard from GBFS consumers and producers indicates that there is a shared understanding about the risks of publicly publishing stable bike_id, and that we just need to come up with the right solutions for how to address it and still meet industry needs.

With other GBFS enhancements we have started with minimum viable proposals, and we should do the same here. What do others think of this plan forward? 
1) For this proposal, we also add "for publicly published GBFS data sets" to the definition of the field?  
2) We open another new issue to discuss bike_id more broadly - considering both publicly published and authenticated feeds?

barbeau:
@heidiguenin I think that path forward sounds reasonable.

I agree with @jcn that a new field for an immutable `immutable_bike_id` that is only allowed in authenticated feeds is probably the best solution.

I agree that we need better language around what is allowed in a public feed for `bike_ids` ASAP - this issue surfaced in https://github.com/NABSA/gbfs/issues/201 yesterday.

barbeau:
@heidiguenin Well, thinking more, I'm going to walk back a thumbs up for:

>For this proposal, we also add "for publicly published GBFS data sets" to the definition of the field?

I think the above text should be part of the broader issue of private feeds and immutable bike IDs in these feeds.

The [GBFS Guiding Principles](https://github.com/NABSA/gbfs#guiding-principles) say (emphasis mine):

>* GBFS is a specification for real-time or semi-real-time, read-only data. The spec is not intended for historical or archival data such as trip records. **The spec is about public information intended for bikeshare users.**
>
>* GBFS is targeted at providing transit information to the bikeshare end user. It’s primary purpose is to power tools for riders that will make bikesharing more accessible to users. **GBFS is about public information.** Producers and owners of GBFS data should take licensing and discoverability into account when publishing GBFS feeds.

IMHO adding explicit text saying "for publicly published GBFS data sets" implies that privately published GBFS data sets are also encouraged/supported, which seems to go directly against the above guiding principles.

So I would propose in this issue we focus only on providing guidance for handling bike IDs in public GBFS feeds, and leave all implicit and explicit modifications related to private feeds for another issue.

gbfs.md
```diff
@@ -213,7 +213,7 @@ Describes bikes that are not at a station and are not currently in the middle of
 Field Name        | Required  | Defines
 ------------------| ----------| ----------
 bikes             | Yes       | Array that contains one object per bike that is currently docked/stopped outside of the system as defined below
-\- bike_id         | Yes       | Unique identifier of a bike
+\- bike_id         | Yes       | Identifier of a bike, rotated to a random string, at minimum, after each trip to protect privacy
```
heidiguenin:
```suggestion
\- bike_id         | Yes       | Identifier of a bike, rotated to a random string, at minimum, after each trip to protect privacy. Note: Persistent bike_id, published publicly, could pose a threat to individual traveler privacy.
```

heidiguenin (commented):
Added language to provide some context for the change, as it will break certain use cases.

gbfs.md
morganherlocker: Looks great 👍

barbeau:
I talked to @heidiguenin further about this issue offline. 

Given the critical privacy issues at stake, and that we want to follow the community process for changes to GBFS that meet the needs of the entire community, I'd like to propose that we try to address both issues (privacy and IDs in private feeds) in this proposal in order to try and move forwards with a viable solution quickly.

I think the simplest path forward is adding the immutable_bike_id that @jcn proposed earlier that is used only in feeds with authentication (i.e., not open publicly available feeds).

So, we would redefine `bike_id` as already proposed (with additional text):

Field Name        | Required  | Defines
------------------| ----------| ----------
\- bike_id         | Yes       | Identifier of a bike, rotated to a random string, at minimum, after each trip to protect privacy. Note: Persistent bike_id, published publicly, poses a threat to individual traveler privacy. See immutable_bike_id for a secure way to publish persistent IDs.

and we would also add:

Field Name        | Required  | Defines
------------------| ----------| ----------
\- immutable_bike_id         | No       | Unique identifier of a bike. IMPORTANT: To protect the privacy of end users, this field **must not** be published in a publicly available open feed. Any feeds containing immutable_bike_id **must** be properly secured and protected by authentication.

@morganherlocker @tmontes @sven4all @jcn @heidiguenin What do you all think of this?

mplsmitch:
@barbeau  what do you think of the @sven4all idea of a separate file ` monitor_bike_status.json` to hold immutable ids so that all the non-pii data in free_bike_status can remain auth free?   That way we could make a clear distinction between feeds where authentication is appropriate and where it's not. I have a hunch that this won't be the last time that there's a valid argument for including data that shouldn't be fully public.

barbeau:
A separate monitor_bike_status.json could work too, although this obviously duplicates a lot of data. But I do see that given authentication would happen at an endpoint level it would make sense to create an endpoint specifically for this use case. A producer would need to create a 2nd endpoint anyway if they were to provide both an open and auth'd version of the feed with immutable_bike_id. By officially spec'ing this it might avoid some producers omitting an open free_bike_status.json.

For everyone following this issue, could you please comment with a `+1 for monitor_bike_status.json` if you prefer spec'ing a separate feed concept, or `+1 for immutable_bike_id` if you prefer this solution?

This isn't an official vote but I'd just like to see what everyone thinks.

tmontes:
> Given the critical privacy issues at stake, and that we want to follow the community process for changes to GBFS that meet the needs of the entire community, I'd like to propose that we try to address both issues (privacy and IDs in private feeds) in this proposal in order to try and move forwards with a viable solution quickly.

The idea of finding a viable solution quickly sounds good to me. The idea of making it sustainable and consistent is important too.

> I think the simplest path forward is adding the immutable_bike_id that @jcn proposed earlier that is used only in feeds with authentication (i.e., not open publicly available feeds).
> 
> So, we would redefine `bike_id` as already proposed (with additional text):
> 
> Field Name	Required	Defines
> - **bike_id**	(required)	Identifier of a bike, rotated to a random string, at minimum, after each trip to protect privacy. Note: Persistent bike_id, published publicly, poses a threat to individual traveler privacy. See immutable_bike_id for a secure way to publish persistent IDs.
> and we would also add:
> 
> Field Name	Required	Defines
> - **immutable_bike_id**	(optional)	Unique identifier of a bike. IMPORTANT: To protect the privacy of end users, this field **must not** be published in a publicly available open feed. Any feeds containing immutable_bike_id **must** be properly secured and protected by authentication.
>
> @morganherlocker @tmontes @sven4all @jcn @heidiguenin What do you all think of this?

Maybe we can do better than this, in a few aspects.

### The identifier concept

If `bike_id` is to be rotated, at minimum, once after each trip, then that means that two successive observations of `free_bike_status.json` are bound (and desired, per this update) to have different values. Thoughts:

* What is the use case for such a non-stable `bike_id`? I can't foresee any.
* Such wording and semantics contradict the general **Field definitions** that currently state:

    > ID fields in the document should be represented as strings that identify that particular object. They:
    > * must be unique within like fields (bike_id must be unique among bikes) 
    > * do not have to be globally unique
    > * must not contain spaces
    > * should be persistent for a given object (station, plan, etc)

This definition is not only important but also pretty much inline with the common expectation among analysts and developers that an ID uniquely identifies one "thing", whichever it is (a bike, in this case).

Wondering:
* If there is no use case for such unpredictable and varying `bike_id`, wouldn't it be simpler to say it is optional? Maybe adding a note on possible use cases that would benefit from its inclusion, mandating that, when that is the case, privacy may be at stake and that providers/consumers must use authenticated and encrypted data transfers?

### Naming

Naming things "right" is important and not a easy task. If we go through with @jcn's proposal of having `bike_id` as a "random thing", and `immutable_bike_id` as the "actual bike id" then I would recommend reviewing the names.

Again, for consistency with the document's  **Field definitions** statements and, importantly, inline with the common analyst/developer expectations on what an ID is.

With that in mind, and if my suggestion of simply making `bike_id` optional does not serve the intended use cases, then I would probably:

* Rename the current `bike_id` to something else.
* Call the proposed `immutable_bike_id` `bike_id`.

### Parting thoughts

I do think that improving this spec does mean thinking hard about current and future use cases. Like I said, I see no use case for a `free_bike_status.json` feed with "random" `bike_id` values and, as such, maybe its simpler to keep saying that they are actual, stable IDs, but that they must be omitted from public feeds, on one hand, and that they can only be included when authenticated and encrypted data transfers are used, on the other.

My 2c.

sven4all:
Random bike_ids can be used for deeplinking in to applications and also for MaaS applications that can use the temporary identifier to reserve bicycles/scooters (this reservation can result in retrieving the static id). Also when the id is rotated only when a trip is ended, you can measure how long a bicycle is parked at a certain location and it's possible to do differential updates (by comparing previous and current file) instead of reading all the data in your database every time you pull the data. 

I am a strong advocate for a separate feed for monitoring purpose. In my opinion it helps to prevent a lot of confusion. If you find a monitor_bike_status.json without proper authentication you immediately know that there is something wrong. Also when you require as a government certain data it's easier to reference to 'monitor_bike_status.json from GBFS'  instead of 'free_bike_status.json from GBFS with authentication and static bike_ids'. 

Duplication of data is not a problem, it's not really different as two different free_bike_status.json endpoints, when producers of GBFS do this efficiently I don't foresee any problems.  

+1 for monitor_bike_status.json for me.

tmontes:
> Random bike_ids can be used for deeplinking in to applications and also for MaaS applications that can use the temporary identifier to reserve bicycles/scooters (this reservation can result in retrieving the static id). Also when the id is rotated only when a trip is ended, you can measure how long a bicycle is parked at a certain location and it's possible to do differential updates (by comparing previous and current file) instead of reading all the data in your database every time you pull the data.

Understood. The current wording does not support that use case. It says "rotated to a random string, at minimum, after each trip". This means that if a given bike is out in the public space, with no use for, say, one hour, three `free_bike_status.json` "reads" may identify that bike differently.

The only way to support that use case, which I've been working with for the City if Lisbon, is if, and only if, `bike_id` rotates **at most** once after each trip. Conflating that with the privacy related concerns, might get us to a wording that says **it must rotate exactly once** after each trip.

I still feel it is unnecessarily complicated. Like I read somewhere (not sure where now, sorry) in this context, the concept of "public" vs "private" data may need to be explicitly described and contextualized in the spec. From there, saying that a given field is mandatory/optional seems to i) keep changes to a minimum, ii) ensure diverse use cases are supported, iii) account for privacy related issues.

> I am a strong advocate for a separate feed for monitoring purpose. In my opinion it helps to prevent a lot of confusion. If you find a monitor_bike_status.json without proper authentication you immediately know that there is something wrong. Also when you require as a government certain data it's easier to reference to 'monitor_bike_status.json from GBFS' instead of 'free_bike_status.json from GBFS with authentication and static bike_ids'.
> 
> Duplication of data is not a problem, it's not really different as two different free_bike_status.json endpoints, when producers of GBFS do this efficiently I don't foresee any problems.

Understood. But I still disagree mostly on a conceptual basis: the spec's feeds should be as orthogonal as possible, from a data perspective. That makes the spec simpler to understand, reason about and, importantly, to change in the future.

And believe me, I have been in those shoes: "reviewing spec details with mobility providers". I honestly believe that such nuisances will become simpler as the spec evolves into something that is less ambiguous to all involved parties. :)

barbeau:
>What is the use case for such a non-stable bike_id? I can't foresee any.

Now that the deep links proposal has passed (https://github.com/NABSA/gbfs/pull/25), those fields should address any use cases related to deeplinking that requires an identifier. The only other use case I'm aware of for a non-stable bike_id is caching and optimization (e.g., to reduce jitter of a bike on a map due to GPS drift), but to my knowledge these aren't critical features and privacy concerns are likely more important. So yes, I agree that there may not be any critical use cases left for this field.

Does anyone object to removing `bike_id` entirely from free_bike_status.json in open GBFS feeds (assuming we come up with another field/feed for the authenticated use case)?

heidiguenin:
@madupras @mdwestervelt @charlesjump @johnpena

As producers, what do you all think about the proposed solutions so that we can get some closure on this issue?

wellorder:
At Bird, we currently rotate IDs.  Removing `bike_id` entirely makes sense; agreed that it's hard to identify a use case for them with free-bike-status.json.

madupras:
> @madupras @mdwestervelt @charlesjump @johnpena
> 
> As producers, what do you all think about the proposed solutions so that we can get some closure on this issue?

At PBSC we do not currently publish any bike information so we cannot help with this change.

johnpena:
Agree with the folks from Bird that removing `bike_id` makes sense as there isn't a great use case.

tmontes:
Chiming in with my clarification after seeing many parties stating that "there are no use cases" for `bike_id` in `free_bike_status.json`.

I do disagree with that and have a current use case in the City of Lisbon where bikes that are in the same place (modulo GPS errors) for a certain period of time (across multiple "samples" of `free_bike_status.json`) are deemed to not be in use. This allows the City to measure how much use a given provider's fleet has -- the figure should ideally be 100%. A large share of unused bikes means that the provider's fleet is just occupying public space with no meaningful purpose.

This use case is possible if either a) `bike_id` values are stable, per the current wording on the spec, or b) rotate at most once after each trip. If they rotate more than once after each trip, such use case is compromised.

AFAICT, @sven4all has a similar use case in the City of Rotterdam and other cities across the Netherlands.

barbeau:
@tmontes To clarify my earlier statement, I'm proposing that we allow bike_id in authenticated feeds (like a new monitor_bike_status.json file), but remove bike_id from unauthenticated public feeds. I believe the above comments were in context of use cases for rotated bike_id in unauthenticated feeds (assuming that a stable bike_id would exist in a separate authenticated feed).  As long as we allow stable bike_ids in authenticated feeds, I believe this meets your use case, right?

tmontes:
@barbeau, thanks for clarifying. Agreed. IMHO, given the tooling that the City already has in place, the best scenario would be:
* Public feeds: no `bike_id` in `free_bike_status.json`.
* Authenticated **and** encrypted feeds: mandatory stable `bike_id` in `free_bike_status.json`.

MuteQ:
Given the recent approval of the [deep linking PR](https://github.com/NABSA/gbfs/pull/25), Transit is OK with the removal of `bike_id` from public `free_bike_status.json` feeds.

The suggestion, however, is a breaking change so I’d recommend this wait for inclusion in 2.0. I want operators to have sufficient time to update their feeds, otherwise, anybody who includes `bike_id` in a public feed would suddenly be non-compliant.

kanagy:
To chime in for Google Maps. Removing bike_id (or rotating more often than after each trip) would break our current integration with Lime. Our use case is:
- We update our feed fairly often on the server.
- User issues a query and we show some bikes on a map. These can go stale.
- User can click on a bike on the map.
- We need a way to say whether that bike is still available after it was clicked. We use bike IDs for this (match that the bike is still available in the latest feed).

There are probably workarounds that we could do, but have to think more about it. Using e.g. lat/lngs wouldn't be reliable because of GPS jitter. I just wanted to point out that removing bike_id indeed breaks backwards compat and should be done in a separate version.

We have no problem with the idea of having bike_id only in authenticated feeds (we use auth) and removing it from publicly available feeds.

Empty2k12:
I also would like to leave some insight into how the local bike-sharing service in our city is handling bike ids:

Here, bike IDs are used to uniquely identify a bike for various purposes such as:
 * reporting problems with a bike
 * reporting improper use of a rider who is currently renting the bike (such as multiple people riding on the same bike)
 * identifying the bike model/generation 

Bike IDs have the format of a number from 1 to ~350 with some bike IDs having an `N_` prefix denoting generation two.

I don't see how these bike IDs can be used to track an individual traveler. At most, the station which a bike is currently docked in can be located and a bike travel history can be created. This feature has actually been helpful since a friend forgot a personal item in the bike basket and was able to locate the bike and retrieve the item. Since many people are sharing these bikes, the travel history might paint an interesting view of how much a bike is being used and which distances it is traveling, but individual riders cannot be identified. Even if a rider is taking the same bike on multiple consecutive trips (to ex. avoid paying a fee for extending 30 min ride time), this cannot be deduced from the bike history.

@kanagy I think we should strive for building an open standard that is valuable to everyone from private programmer to big corporations. Not everyone is Google and can get access to a authenticated feed. 

Open data should be open to everyone!

kanagy:
@Empty2k12 You mention a docked system. This proposal is about rotating free floating bike and scooter IDs. These are privacy sensitive because anyone could track user's destinations like this:
- Find the scooters/bikes parked outside the user's house.
- Use public GBFS to track that they disappear from your house and re-appear at destination X (which is potentially an incriminating destination). bike_id is what allows this.
- Do this multiple times to gain confidence it is indeed the user going to X.

Rotating bike ID once after each trip to a random string (as proposed) looks like a good balance of public feed and privacy. Why can't we just do that - is it the city monitoring use case? The only other approach so far is to not have it publicly available.

Empty2k12:
In #136 you say remove bike_id from station_status.json since the sentiment here is against it, here you say this talks about free_bike_status only. I think the use case is there for both docked and dockless systems, but removing it from station_status.json as well does not offer any added privacy.

I think the problem for dockless systems is also quite smaller in cities, where scooters typically operate (at least here in Germany). So tieing a starting point to a specific persion is again quite hard, as hundreds of people live in the same block and building.

I still think that bike_id should stay in GBFS as an optional field so that operators can choose to include it if their service area is dense enough and re-identification is not easily done.

Empty2k12:
Additionally, the precise location and the full scooter ID or parts thereof can still be obtained by reverse-engineering the mobile app from the scooter company and talking with the backend directly. Removing it from GBFS means people building legitimate applications will not be able to use the info easily, while still being available for anyone really interested in the data.

barbeau:
@Empty2k12 Let's focus on `free_bike_status.json` for this particular proposal. Would it negatively impact your city if we removed `bike_id` from open `free_bike_status.json` feeds? As mentioned above `bike_id` could still be provided in a separate authenticated feed like a new `monitor_bike_status.json`.

Empty2k12:
Authenticated feeds shift the problem somewhere else, now only select companies and cities can re-id riders. How can we trust they don't do that? If this is really about complete and utter user privacy, it should be removed from all feeds.

I can already see Cambridge Analytica type situations where the authenticated feed is abused. Making feeds authenticated only worsens the burden a malicious actor has to undertake. Make everything public or remove it everywhere.

> Would it negatively impact your city if we removed bike_id from open free_bike_status.json feeds?

It would negatively impact scooter feeds we (as Open Data activists) provide for our city, since the data is still available to us (as Open Data activists) but not developers and data scientists using our data feeds to build awesome applications or hack on improvements of public infrastructure.

barbeau:
>Authenticated feeds shift the problem somewhere else, now only select companies and cities can re-id riders. How can we trust they don't do that? If this is really about complete and utter user privacy, it should be removed from all feeds.

It's up to the city/jurisdiction to manage data sharing agreements and vet consumers, which is outside the scope of GBFS.  GBFS has [Guiding Principles](https://github.com/NABSA/gbfs#guiding-principles) that narrow it's focus to real-time passenger data which should not infringe on user privacy.

>I can already see Cambridge Analytica type situations where the authenticated feed is abused. Making feeds authenticated only worsens the burden a malicious actor has to undertake. Make everything public or remove it everywhere.

It's certainly possible for abuse to occur, but if the feed is authenticated at least the city/jurisdiction has legal recourse via agreements they can create. And it's clear to data producers that they should take rider privacy seriously.

>It would negatively impact scooter feeds we (as Open Data activists) provide for our city, since the data is still available to us (as Open Data activists) but not developers and data scientists using our data feeds to build awesome applications or hack on improvements of public infrastructure.

@Empty2k12 Could you provide an exact use case of bike_id in free_bike_status.json that removing bike_id from open feeds would break?  Note that a new monitor_bike_status.json feed would effectively be a copy of free_bike_status.json, just with authentication and bike_id added. And that the rest of free_bike_status.json and other feed URLs would remain open.

kanagy:
barbeau@ Can you clarify why rotating the bike_id after each trip is no longer considered?

barbeau:
@kanagy As @tmontes pointed out [earlier](https://github.com/NABSA/gbfs/pull/147#issuecomment-564488598), there doesn't seem to be a valid remaining use case for a rotated bike_id, now that the deep links proposal has passed (#25). The only other use case I'm aware of for a non-stable bike_id is caching and optimization (e.g., to reduce jitter of a bike on a map due to GPS drift), but to my knowledge these aren't critical features and privacy concerns are likely more important.

That being said, ultimately either solution (rotating the ID or removing it) is still on the table - there haven't been any official votes yet.

Empty2k12:
> And it's clear to data producers that they should take rider privacy seriously.

That's why `bike_id` should remain public and be set at the discretion of city/jurisdiction/operator! I don't see why educating these parties and putting best practice items in the Guidelines is not as good, if not better then completely removing. This way, safe applications can still benefit from bike_ids, while possibly insecure use cases (e.g. Scooters) chose to not include it.

> Could you provide an exact use case of bike_id in free_bike_status.json that removing bike_id from open feeds would break?

An application that uses public Scooter GBFS feed to display Mobility Locations on a map and displays Scooter ID. For your understanding, I attached a screenshot from an application running with one of our scooter feeds:
 
Screenshot 2020-01-08 at 14 53 17

morganherlocker:
Hey everyone, thank you for the excellent discussion. I really appreciate all the wonderful feedback and perspectives from you all. 🙇 

It sounds like we are at a consensus point where:

1. There is general agreement that stable bike_id does not meet the privacy goals of the spec
2. Rotating bike_ids or dropping bike_id both solve the immediate privacy/safety concerns
3. Dropping bike_id will not hurt any primary use cases, but could prevent certain quality improvements (reducing GPS drift)

Since this an active, widespread vulnerability, I would like to see us move to a vote and get this merged as quickly as possible. I'm supportive of the emerging regulatory use cases around authenticated feeds that could contain new metadata, and believe we should follow this up with separate formal proposals to shore up these ideas in the specification. These use cases are important and deserving of dedicated discussion and design process.

With expediency in mind, **I voice my support for merging bike_id rotation as the PR currently describes, since this position seems to have the fewest operational concerns and the broadest level of consensus, while addressing the user safety issue. I'll follow up this comment with a call for a formal vote, following the [GBFS governance guidelines](https://github.com/NABSA/gbfs/issues/163#issuecomment-559218562).** 

Thank you for taking the time to follow along and contribute to this process!

morganherlocker:
I hereby call a vote on this proposal. Voting will be open for 7 full days, until 11:59PM UTC on January 16.

tmontes:
> @kanagy As @tmontes pointed out [earlier](https://github.com/NABSA/gbfs/pull/147#issuecomment-564488598), there doesn't seem to be a valid remaining use case for a rotated bike_id, now that the deep links proposal has passed (#25). The only other use case I'm aware of for a non-stable bike_id is caching and optimization (e.g., to reduce jitter of a bike on a map due to GPS drift), but to my knowledge these aren't critical features and privacy concerns are likely more important.

Clarification:
* I see no use case for a rotated `bike_id` with the current wording: "rotated to a random string, at minimum, after each trip".
* I do see (and have) use cases for a rotated `bike_id` if it is "rotated to a random string exactly once after each trip".

I know a call for vote is ongoing and I'll review this with the City of Lisbon officials before casting any vote. Nevertheless, I suppose a better PR would say:
* No `bike_id` field in public feeds -- privacy 100% guaranteed.
* Mandatory and stable `bike_id` in private feeds -- private meaning authenticated and encrypted.
* I don't see the need for `bike_id` WRT to the recently integrated deep link capabilities given that its wording explicitly states "(...) Note that the URI does not necessarily include the bike_id (...)" -- whether those links are exploitable from a privacy standpoint is a whole other matter.

My 2c.

antrim:
Hi Tiago - Thanks for the thorough comments. Here are some notes for your and City of Lisbon's consideration.

(Responses inline)

> * I see no use case for a rotated `bike_id` with the current wording: "rotated to a random string, at minimum, after each trip".

I believe some operators intend to rotate bike_id after a bike is moved or after some other "non-trip" event. "…At a minimum" would explicitly allow that, and would not diminish functionality in traveler-facing applications.

> * Mandatory and stable `bike_id` in private feeds -- private meaning authenticated and encrypted.

The stated purpose of GBFS is for traveler-facing applications. GBFS has also been useful for the purposes of monitoring and oversight by cities. I believe there needs to be a more thorough process to determine best practices and standards for that purpose -- beyond this immediate privacy patch. Long-term, I think there is a broadly held intention that GBFS and associated best practices should offer sound guidance to cities looking to set their own policies and requirements related to data for traveler-facing and monitoring applications.

> * I don't see the need for `bike_id` WRT to the recently integrated deep link capabilities given that its wording explicitly states "(...) Note that the URI does not necessarily include the bike_id (...)" -- whether those links are exploitable from a privacy standpoint is a whole other matter.

Some consumers and producers have committed to implementing the deeplinks proposal (https://github.com/NABSA/gbfs/pull/25#issuecomment-562167173) but to my knowledge it hasn't yet been implemented. It takes time for new features to be implemented. The purpose of specifying bike_id rotation is that this is already current GBFS practice in many cases and easily implementable. It is an expedient way of solving a privacy issue that otherwise remains open. In future versions, the community will have the opportunity to consider some of the ideas you've put forward.

quicklywilliam:
Ride Report votes in favor of this change.

kanagy:
Google Maps votes in favor for the current proposal.

charlesjump:
Uber/JUMP votes in favor of this change.

rickbruce:
Ito World votes in favor of this change.

bhandzo:
Bird votes in favor of this change.

yocontra:
Stae votes in favor of this change.

MuteQ:
Transit votes in favor of this change.

heidiguenin:
Voting has closed, and this change has been passed with 7 votes in favor and 0 votes against:

Consumers (5): RideReport, Google Maps, Ito World, Stae, Transit 
Producers (2): Uber/JUMP, Bird

Next steps:
We'll merge the PR and add it to the v2.0 Major Release candidate.
@antrim @barbeau

heidiguenin:
We'd love to make this an official part of the spec, but first we need to see this change being implemented. Could you comment here if your organization has implemented this?

@MuteQ @contra @bhandzo @rickbruce @charlesjump @kanagy @quicklywilliam  Others?

mdwestervelt:
@heidiguenin Bird has implemented rotating IDs

heidiguenin:
I opened a new issue #237 to discuss options for maintaining the existing GBFS use cases that may have been broken by bike_id rotation.

PR #152, opened 2019-07-01
Added Chicago, IL Scooter feeds to systems.csv

nicklucius:


jcn (changes requested):
@nicklucius Hey there, it looks like whatever editor you used to make these changes have changed every line in the file, which makes it hard to actually review. Can you please make sure that only the lines that you're intending to add/remove are part of this pull request?

Thanks!

nicklucius:
@jcn - got it, will do. I fixed it locally but I don't think it will show up on this PR. I'll close and resubmit.

PR #158, opened 2019-07-18
Update README.md

heidiguenin:
Updating the readme with the beginning of the all-contributors spec. Will need to do some more leg-work to populate the table from historical contributions. In the future will be able to use the all-contributors bot.


heidiguenin:
Decided to try a different approach for the initial acknowledgements v.01 :)

PR #168, opened 2019-09-08
Update Chicago gruv Auto-Discovery URL

nicklucius:


jcn:
It appears that this feed is being returned as HTML as opposed to `application/json` and that the URLs in the auto-discovery file are being escaped and so aren't actually valid URLs. 

In addition, the `system_id`, while technically valid, is just an integer, which is probably going to cause some collisions in the future.

Do you have contacts at grüv who can address these issues?

nicklucius:
@jcn yes thank you, I will check in with grüv and follow up.

jcn:
@nicklucius I'm afraid this still does not look right. The gruv file is still returning bad information, and the Veoride URL has been moved from an auto-discovery URL to the free_bike_status url, which is also not what we're expecting here. Thanks for looking into this.

nicklucius:
@jcn Thanks. I'm still waiting on gruv, and will update Veoride with the correct url. I'll close this PR for now.

PR #174, opened 2019-09-26
Add Beryl feeds

seanwiseman:
Add https://beryl.cc/ feeds. 🚲 

(I am a dev there)

### DIFF
--- a/systems.csv
+++ b/systems.csv
@@ -61,6 +61,9 @@ ES,Bilbaobizi (Bilbao),"Bilbao, ES",nextbike_bo,https://www.bilbaobizi.bilbao.eu
 ES,Sitycleta (Las Palmas),"Las Palmas de Gran Canaria, ES",nextbike_el,https://www.sitycleta.com/xx/laspalmas/,https://gbfs.nextbike.net/maps/gbfs/v1/nextbike_el/gbfs.json
 FR,Libélo,"Valence, FR",libelo,http://www.citea.info/presentation/?rub_code=92,https://valence.publicbikesystem.net/ube/gbfs/v1/gbfs.json
 GB,BelfastBikes,"Belfast, GB",nextbike_bu,https://www.belfastbikes.co.uk/xx/belfast/,https://gbfs.nextbike.net/maps/gbfs/v1/nextbike_bu/gbfs.json
+GB,Beryl - BCP,"Bournemouth, Christchurch, Poole, GB",beryl_bc,https://www.beryl.cc,http://gbfs.basis-pdn.bike/BCP/gbfs.json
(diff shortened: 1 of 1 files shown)


systems.csv
```diff
@@ -61,6 +61,9 @@ ES,Bilbaobizi (Bilbao),"Bilbao, ES",nextbike_bo,https://www.bilbaobizi.bilbao.eu
 ES,Sitycleta (Las Palmas),"Las Palmas de Gran Canaria, ES",nextbike_el,https://www.sitycleta.com/xx/laspalmas/,https://gbfs.nextbike.net/maps/gbfs/v1/nextbike_el/gbfs.json
 FR,Libélo,"Valence, FR",libelo,http://www.citea.info/presentation/?rub_code=92,https://valence.publicbikesystem.net/ube/gbfs/v1/gbfs.json
 GB,BelfastBikes,"Belfast, GB",nextbike_bu,https://www.belfastbikes.co.uk/xx/belfast/,https://gbfs.nextbike.net/maps/gbfs/v1/nextbike_bu/gbfs.json
+GB,Beryl - BCP,"Bournemouth, Christchurch, Poole, GB",beryl_bc,https://www.beryl.cc,http://gbfs.basis-pdn.bike/BCP/gbfs.json
+GB,Beryl - Hereford,"Hereford, GB",beryl_he,https://www.beryl.cc,http://gbfs.basis-pdn.bike/Hereford/gbfs.json
+GB,Beryl - London,"London, GB",beryl_lo,https://www.beryl.cc,http://gbfs.basis-pdn.bike/London/gbfs.json
```
bddq, line 66: It's too bad that your GBFS files are not hosted under an HTTPS url.

seanwiseman: This is down to an additional load balancer being required in order to allow `ssl` redirects for the hosting of these data files. Should we need to in the future we can make the appropriate infrastructure changes.

If need be users can also get it via googles domain over `ssl` e.g.:

https://storage.googleapis.com/gbfs.basis-test.bike/BCP/free_bike_status.json

jcn (changes requested):
@seanwiseman The GBFS spec notes that `system_id` "should be globally unique (even between different systems) and it is currently up to the publisher of the feed to guarantee uniqueness."

Currently all of your systems report a system_id as `beryl_en` where they should instead be something like `beryl_bcp` and `beryl_london` to ensure that consumers can distinguish between the feeds internally.

seanwiseman:
> @seanwiseman The GBFS spec notes that `system_id` "should be globally unique (even between different systems) and it is currently up to the publisher of the feed to guarantee uniqueness."
> 
> Currently all of your systems report a system_id as `beryl_en` where they should instead be something like `beryl_bcp` and `beryl_london` to ensure that consumers can distinguish between the feeds internally.

Hi @jcn , this is now implemented.

jcn (approved):

Issue #15, opened 2016-01-15
Do we need to include GBFS version?


jcn:
@mdarveau and @dsgermain raise up a good point in #9 about versioning. While we've said that we would version the spec in the repository via git tags, we did not actually ever decide where to indicate the GBFS version in the feed. 

I can imagine two logical places for this:
1. In gbfs.json, in which case we would want to make this file required (though the version field would be the only file that was required - the auto-discovery mechanism would change to optional)
2. In system_information.json, but this doesn't really make sense since it's metadata about the feed, not about the system

I am leaning toward including it in gbfs.json but would like other people's thoughts.

emilesalem:
How about in the common header info.

jcn:
@E1000s That's an interesting idea, though I'd be concerned about a system deciding to mix and match their feeds. This shouldn't be supported (i.e. you shouldn't be able to have a 1.0 `station_information` but a 2.0 `station_status`) but if the version was in the header of each individual file this might implied.

jcn:
I would also love to hear from people who have implemented GTFS consumers and publishers. The GTFS spec actually indicates that changes should be backward compatible. This is probably something we should shoot for and versioning should be saved for major breaking changes. The problem of course with breaking changes is that they're a pain for everyone (consumers and publishers), so they should really be avoided.

dsgermain:
The gbfs.json should maybe list a correspondence between a version number and the URL path of the specified version ( /v1, /v2, etc.).
The other option is to pass the version number in headers and for the server to redirect to the proper endpoint. I prefer the first option. Thoughts?

mdarveau:
I agree we should aim for no breaking change. This does not mean we should not version the file so the consumer can know upfront what to expect.

Proposition:
- In `gbfs.json` we add a version field that respect http://semver.org/ and we increment the minor version in changes that are backward compatible.
- If we have a breaking change, we would introduce a new header so publisher can publish the old and new version. ie: `<link rel="gbfs-v2" type="application/json" href="http://mybikeshare.com/opendata/v2/gbfs.json" />`
  - This would allow breaking change in the `gbfs.json` itself

cubbi:
Shouldn't we simply link to other versions of the protocol from gbfs.json file? This would not allow for breaking changes in the gbfs.json, but in majority cases we will rather add fileds than remove, and would not require change of the link attribute.

Also maintaining few versions inside gbfs.json file will be more elegant than having 2 or 3 link tags in the header.

[...]

Issue #163, opened 2019-08-20
Governance pilot proposal & upcoming changes


antrim:
NABSA has selected MobilityData IO (@mobilitydata) to support them in their work to update and enhance GBFS, as mentioned in issue #137. MobilityData works to improve the coverage, completeness, and quality of mobility data standards around the world, including the General Transit Feed Specification (GTFS). GBFS has been modeled off of GTFS in many ways, and we'll continue to borrow relevant lessons from GTFS - including during this governance pilot.

Currently, GBFS' change process is not as detailed as may be necessary to effectively clarify and enhance the spec. In the coming weeks we'll be working to clarify and enhance GBFS. We propose piloting the [GTFS change framework](https://github.com/MobilityData/transit/blob/master/gtfs/CHANGES.md) within the GBFS repo as we do this work. (Some parts of the process, such as the CLA and the mailing list, are not relevant for GBFS and will not be used.)

In some cases, piloting the GTFS change framework may mean working with the authors of pull requests or issues to combine or split posts up. For example, pull request #92 may be a good candidate for moving some parts of the conversation to their own issue or pull request.

Please respond if you have a specific objection we can address or if you support this pilot governance proposal.

antrim:
Update on the governance pilot:

We have been following the _voting_ process of the GTFS change framework, which requires at least seven days of voting and unanimous yes votes, with at least three votes total from including at least one vote each from a producer and consumer.

In order to ensure practicality, the GTFS change framework also requires that any accepted specification change must first be implemented in at least one dataset and in at least one consuming application. Over the past 5 months, @MobilityData has _not_ strictly observed this other requirement in our facilitation of the GBFS change process because we believed it would greatly slow down the process of catching GBFS up with shared mobility industry practice. We have tried to ensure practicality by calling votes on "minimum viable proposals" and asking stakeholders to commit to implement the proposal on an unspecified timeline.

We propose to codify this process thus:

* Any change proposal should be open on GitHub for at least 7 days of discussion before voting.
* Each part of the proposal must have at least one organization committed to its implementation: either at least one organization committed to implementing the entire proposal or separate organizations committed to partial implementations.
* For changes to the specification at least 3 votes in favor are required for a proposal to pass.
* Voting is announced by a message that conforms to this template:

> I hereby call a vote on this proposal. Voting will be open for 7 full days, until 11:59PM UTC on X.
>
> Please vote for or against the proposal, and include the organization for which you are voting in your comment.
>
[...]

PR #182, opened 2019-10-09
Require license_url, add attribution fields

antrim:
`license_url` is an optional field in `system_information.json` (added in PR #49). This new pull request would make a backwards-compatibility breaking change to define `license_url` as mandatory. The outcome of the discussion related to versioning (issue #15) would define how such a change would be treated and incorporated into the spec.

This proposal would:
- Make `feed_license` a mandatory field.
- Add statement that a `license_url` being present but empty means the feed is released without a license (public domain).
- Add fields for attribution to the GBFS, partially borrowed from this [GTFS-Attributions](https://github.com/google/transit/issues/138) proposal.
- Add a limited inventory of licenses to the GBFS wiki, with implications of each.


yocontra:
Can we use [SPDX](https://spdx.org/licenses/) license IDs? This is what most OSS tools have converged around, and I think it makes sense here as well.

antrim:
@contra : Makes good sense. I'll revise the PR.

antrim:
**I've revised the PR:**
* **Added `license_id`.** This is a string that refers to the SPDX identifier.
* **`license_url` and `license_id` are both conditionally required.** Only one should be provided. If a standard license then feed publishers should specify `license_id`; if custom, then `license_url`.

Since the introduced requirement makes the change backwards-compatibility breaking, then this would need to be part of a MAJOR version. (see #188)

**What this PR does NOT currently do:**
* **Make any recommendation with regard to particular licenses or custom vs. standard licenses.** I was tempted to include a recommendation to use a standard license, but thought better of writing that sort of practice recommendation into the specification, at least before there is a more thorough research and stakeholder interview process.
* **Allow SPDX [License Expressions]((https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60)).** This would allow multiple standard licenses to be applied or a standard license to be applied with modifications. I figured this would just add complexity now, and could be added later if necessary. LMK if that's wrong and this would be used immediately.

antrim:
I hereby call a vote on this proposal. Because of the US Thanksgiving holiday, I propose we extend by 1 day. Voting will be open for 8 full days, until 11:59PM UTC on Tuesday, Dec 3.

Please vote for or against the proposal to add a data license, and include the organization for which you are voting in your comment.

barbeau:
@antrim I know in some states (including Florida) Thursday and Friday are both considered holidays:
https://www.dms.myflorida.com/workforce_operations/human_resource_management/for_state_personnel_system_hr_practitioners/state_holidays

...so I'd suggest giving two additional days to account for this.

antrim:
Following @barbeau 's suggestion, I'd like to extend the voting period until 11:59PM UTC on Dec 6 to accommodate the USA Thanksgiving holiday.

Please vote, and note if you can commit to implementing this proposal.

yocontra:
+1 from Stae

madupras:
+1 from PBSC

quicklywilliam:
+1 Ride Report

evansiroky:
+1 from IBI Group

antrim:
Voting has closed and this changed has been passed. Four votes were in favor:
* IBI Group (Developer of OpenTripPlanner, GBFS consumer)
* Stae (Vendor)
* PBSC (Vendor / GBFS producer)
* Ride Report (GBFS consumer)

@MobilityData is refining how we'll implement the versioning scheme (#188). We'll merge this once we arrive at a sensible release process (see #163).

antrim:
Merged - branch deleted.

heidiguenin:
We'd love to make this an official part of the spec, but first we need to see this change being implemented. Could you comment here if your organization has implemented this? [Sorry to those of you fabulously active community members who will see this message over and over again today!]

@evansiroky @quicklywilliam @madupras @contra Others?

madupras:
We still have multiple months before tackling this change at PBSC. It's in our "GBFS version 2" internal project which is planned for the Q3/Q4 2020.

PR #183, opened 2019-10-09
system_regions.json: deprecation candidate

antrim:
system_regions.json is currently optional.

At the GBFS Developers Workshop last week, there was only one known instance of a feed that provides system_regions.json, ([Capital Bikeshare](https://gbfs.capitalbikeshare.com/gbfs/gbfs.json)). It is not known how or if this data is being used.

**This pull request is to discuss deprecation candidacy for system_regions.json.** Is this data being used by applications? If not, should we remove system_regions.json?

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -15,7 +15,6 @@ This document explains the types of files and data that comprise the General Bik
     * [free_bike_status.json](#free_bike_statusjson)
     * [system_hours.json](#system_hoursjson)
     * [system_calendar.json](#system_calendarjson)
-    * [system_regions.json](#system_regionsjson)
     * [system_pricing_plans.json](#system_pricing_plansjson)
     * [system_alerts.json](#system_alertsjson)
 * [Possible Future Enhancements](#possible-future-enhancements)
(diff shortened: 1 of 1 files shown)


mwillmott:
Hello! Just getting stuck into various proposals in a bit more detail. I appreciate a defined operating area is less important for fully docked schemes but they are critically important for other operators (even if they aren't all making use of this feed).

At [Beryl](https://beryl.cc/bikeshare) we absolutely make use of `system_regions.json` and have several third-parties that rely on it... so i'm strongly against removing this. If it is removed it should be replaced with another way to describe the area in which bikes can be used.

testower:
If a company operates in several cities and provides a separate gbfs feed for each city, then system_regions seems to be the best way to describe each region. E.g. use region_id in system_information to point to a region in system_regions. If this is removed, then it seems necessary that we need some other way to describe the geographical restrictions of operation.

josee-sabourin:
Hi all! This doesn't seem ready for a change proposal since some stakeholders are clearly using the file. I'm going to close this PR and the discussion can continue over in the issue for the same topic (#183) to determine how to best approach improving system_regions.json

@mwillmott @testower

PR #188, opened 2019-11-06
GBFS documentation versioning and and feed conformance

antrim:
Issue #15 contains a discussion thread about versioning GBFS documentation and data. Versioning is needed to enable backwards-compatibility breaking changes.

This PR:
- Defines a semantic versioning approach for backwards-compatibility breaking (MAJOR) and backwards-compatible (MINOR) GBFS changes
- Includes release practice guidelines, including a long-term support (LTS) version and maximum number of MAJOR releases per year. New versions would be subject to community vote, so these are guidelines, not hard rules.
- Defines a backwards-compatible lookup mechanism to disseminate GBFS data feeds encoded in various versions of the specification.
- Does not yet include the `versionCode` concept (integers associated with each named version), but this could be added if requested.

Please comment and propose any changes. @MobilityData proposes to bring this to vote in 1 week or less so that GBFS can begin to incorporate necessary changes that break backwards-compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The general outline for changing the spec has 4 steps:
 4.	Submit a final request-for-comments on the proposed change to the issue discussion. If no outstanding issues are identified after one week’s time, and there is general agreement that the proposed change is worthwhile and follows the GBFS guiding principles outlined below, the proposal will be officially adopted.
 
 ## Specification Versioning
-To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. A git tag in the form of `X.Y` establishes semantic versions. Git tags batch pull requests into versions.
+To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. A git tag in the form of `X.Y` establishes semantic versions. Multiple changes (commits) may be batched into a single new release.
 
 A whole integer increase is used for breaking changes (MAJOR changes). A decimal increase is used for non-breaking changes (MINOR changes or patches).
 
@@ -57,9 +57,9 @@ Examples of non-breaking changes include:
 * Modifying documentation or spec language in a way that clarifies semantics or recommended practices
 
 ### Version Release Cycles
-* There is no strict limitation on the frequency of MAJOR releases, but the GBFS community aims to limit the MAJOR releases to 2 or less every 12 months. To limit releases, breaking changes can be batched together.
-* MINOR changes may be applied at any time. There is no guideline to limit the number of MINOR changes.
-* GBFS documentation will include a designated long-term support (LTS) branch. The LTS branch must maintain backwards compatibility for a period of 2 years. Non-breaking changes (MINOR) may be applied to the LTS branch.
+* There is no strict limitation on the frequency of MAJOR releases, but the GBFS community aims to limit the MAJOR releases to 2 or fewer every 12 months. To limit releases, breaking changes can be batched together.
+* MINOR changes may be applied at any time. There is no guideline to limit the number of MINOR changes. MINOR changes may be batched or released immediately, at the discretion of the pull request author and advocate.
+* GBFS documentation will include a designated long-term support (LTS) branch. The LTS branch would maintain its LTS status for at least 2 years, after which a new LTS release and branch would be designated. The LTS branch will be determined according to the GBFS voting process. Non-breaking changes (MINOR) will be applied to the LTS branch when relevant.
 
 ## Extensions Outside of the Specification ##
 To accommodate the needs of feed producers and consumers prior to the adoption of a change, additional fields can be added to feeds even if these fields are not part of the official specification. It's strongly recommended that these additional fields be documented on the wiki page in this format:
--- a/README.md
+++ b/README.md
@@ -40,6 +40,26 @@ The general outline for changing the spec has 4 steps:
 3.	Find at least one GBFS producer to implement and test the proposed change.
 4.	Submit a final request-for-comments on the proposed change to the issue discussion. If no outstanding issues are identified after one week’s time, and there is general agreement that the proposed change is worthwhile and follows the GBFS guiding principles outlined below, the proposal will be officially adopted.
 
+## Specification versioning
+To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. A git tag in the form of `X.Y` establishes semantic versions. Git tags batch pull requests into versions.
+
+A whole integer increase is used for breaking changes (MAJOR changes). A decimal increase is used for non-breaking changes (MINOR changes or patches).
+
+Examples of breaking changes include:
+
+* Adding or removing a required endpoint or field
+* Changing the data type or semantics of an existing field
+
+Examples of non-breaking changes include:
+ 
+* Adding or removing an optional endpoint or field
+* Adding or removing enum values
+* Modifying documentation or spec language in a way that clarifies semantics or recommended practices
+
+### Version release cycles
+* There is no strict limitation on the frequency of MAJOR releases, but the GBFS community aims to limit the MAJOR releases to 2 or less every 12 months. To limit releases, breaking changes can be batched together.
+* MINOR changes may be applied at any time. There is no guideline to limit the number of MINOR changes.
+* GBFS documentation will include a designated long-term support (LTS) branch. The LTS branch must maintain backwards compatibility for a period of 2 years. Non-breaking changes (MINOR) may be applied to the LTS branch.
 
 ## Extensions Outside of the Specification ##
 To accommodate the needs of feed producers and consumers prior to the adoption of a change, additional fields can be added to feeds even if these fields are not part of the official specification. It's strongly recommended that these additional fields be documented on the wiki page in this format:
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The general outline for changing the spec has 4 steps:
 4.	Submit a final request-for-comments on the proposed change to the issue discussion. If no outstanding issues are identified after one week’s time, and there is general agreement that the proposed change is worthwhile and follows the GBFS guiding principles outlined below, the proposal will be officially adopted.
 
 ## Specification Versioning
-To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. A git tag in the form of `X.Y` establishes semantic versions. Multiple changes (commits) may be batched into a single new release.
+To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. Semantic versions are established by a git tag in the form of `vX.Y` where `X.Y` is the version name. Multiple changes (commits) may be batched into a single new release.
 
 A whole integer increase is used for breaking changes (MAJOR changes). A decimal increase is used for non-breaking changes (MINOR changes or patches).
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ The general outline for changing the spec has 4 steps:
 3.	Find at least one GBFS producer to implement and test the proposed change.
 4.	Submit a final request-for-comments on the proposed change to the issue discussion. If no outstanding issues are identified after one week’s time, and there is general agreement that the proposed change is worthwhile and follows the GBFS guiding principles outlined below, the proposal will be officially adopted.
 
-## Specification versioning
+## Specification Versioning
 To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. A git tag in the form of `X.Y` establishes semantic versions. Git tags batch pull requests into versions.
 
 A whole integer increase is used for breaking changes (MAJOR changes). A decimal increase is used for non-breaking changes (MINOR changes or patches).
@@ -56,7 +56,7 @@ Examples of non-breaking changes include:
 * Adding or removing enum values
 * Modifying documentation or spec language in a way that clarifies semantics or recommended practices
 
-### Version release cycles
+### Version Release Cycles
 * There is no strict limitation on the frequency of MAJOR releases, but the GBFS community aims to limit the MAJOR releases to 2 or less every 12 months. To limit releases, breaking changes can be batched together.
 * MINOR changes may be applied at any time. There is no guideline to limit the number of MINOR changes.
 * GBFS documentation will include a designated long-term support (LTS) branch. The LTS branch must maintain backwards compatibility for a period of 2 years. Non-breaking changes (MINOR) may be applied to the LTS branch.
--- a/gbfs.md
+++ b/gbfs.md
@@ -149,11 +149,11 @@ Example:
       "feeds": [
         {
           "name": "system_information",
-          "url": "https://www.example.com/gbfs/2/en/system_information"
+          "url": "https://www.example.com/gbfs/1/en/system_information"
         },
         {
           "name": "station_information",
-          "url": "https://www.example.com/gbfs/2/en/station_information"
+          "url": "https://www.example.com/gbfs/1/en/station_information"
         }
       ]
     },
@@ -161,11 +161,11 @@ Example:
       "feeds": [
         {
           "name": "system_information",
-          "url": "https://www.example.com/gbfs/2/fr/system_information"
+          "url": "https://www.example.com/gbfs/1/fr/system_information"
         },
         {
           "name": "station_information",
-          "url": "https://www.example.com/gbfs/2/fr/station_information"
+          "url": "https://www.example.com/gbfs/1/fr/station_information"
         }
       ]
     }
--- a/gbfs.md
+++ b/gbfs.md
@@ -118,7 +118,7 @@ Example:
 {
   "last_updated": 1434054678,
   "ttl": 3600,
-  "version":"1.0",
+  "version": "1.0",
   "data": {
     "name": "Citi Bike",
     "system_id": "citibike_com"
@@ -143,7 +143,7 @@ Example:
 {
   "last_updated": 1434054678,
   "ttl": 0,
-  "version":"1.0",
+  "version": "1.0",
   "data": {
     "en": {
       "feeds": [
@@ -188,22 +188,23 @@ _versions_              | Yes         | Array that contains one object, as defin
 {
   "last_updated": 1434054678,
   "ttl": 0,
-  "version":"1.0",
+  "version": "1.0",
   "data": {
-    [
-      {
-         "version":"1.0",
-         "url":"https://www.example.com/gbfs/1.0/gbfs"
-      },
-      {
-         "version":"1.1",
-         "url":"https://www.example.com/gbfs/1.1/gbfs"
-      },
-      {
-         "version":"2.0",
-         "url":"https://www.example.com/gbfs/2.0/gbfs"
-      }
-    ]
+  	"versions":
+		[
+		  {
+			 "version": "1.0",
+			 "url":"https://www.example.com/gbfs/1.0/gbfs"
+		  },
+		  {
+			 "version":"1.1",
+			 "url":"https://www.example.com/gbfs/1.1/gbfs"
+		  },
+		  {
+			 "version":"2.0",
+			 "url":"https://www.example.com/gbfs/2.0/gbfs"
+		  }
+		]
   }
 }
 ```
@@ -288,7 +289,7 @@ Example:
 {
   "last_updated": 1434054678,
   "ttl": 0,
-  "version":"1.0",
+  "version": "1.0",
   "data": {
     "rental_hours": [
       {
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -40,6 +40,26 @@ The general outline for changing the spec has 4 steps:
...
+
+Examples of non-breaking changes include:
+ 
+* Adding or removing an optional endpoint or field
+* Adding or removing enum values
+* Modifying documentation or spec language in a way that clarifies semantics or recommended practices
+
+### Version Release Cycles
+* There is no strict limitation on the frequency of MAJOR releases, but the GBFS community aims to limit the MAJOR releases to 2 or less every 12 months. To limit releases, breaking changes can be batched together.
+* MINOR changes may be applied at any time. There is no guideline to limit the number of MINOR changes.
```
barbeau: Are minor changes batched too, or does every change that qualifies as a minor change bump the `MINOR` version?

README.md
```diff
@@ -40,6 +40,26 @@ The general outline for changing the spec has 4 steps:
...
+* GBFS documentation will include a designated long-term support (LTS) branch. The LTS branch must maintain backwards compatibility for a period of 2 years. Non-breaking changes (MINOR) may be applied to the LTS branch.
```
barbeau: I'd suggest that:

`Non-breaking changes (MINOR) may be applied to the LTS branch.`

...be changed to:

`Non-breaking changes (MINOR) will be applied to the LTS branch when relevant.`

In other words, we plan to apply all minor changes to the LTS branch unless there is a specific reason not to.

barbeau: There are a few items that IMHO need to be clarified here:
1. How are the LTS releases chosen/approved? In #15 we said `The current LTS branch would be changed according to the GBFS governance process (currently a vote of the community).`
1. `The LTS branch must maintain backwards compatibility for a period of 2 years` - So I think this means that every 2 years we'd designate a new LTS branch. If so, I'd suggest changing this language to `The LTS branch would maintain its LTS status for at least 2 years, after which a new LTS release and branch would be designated`, which is similar to the language in #15.

gbfs.md
```diff
@@ -34,12 +34,23 @@ This specification has been designed with the following concepts in mind:
 
 Historical data, including station details and ride data is to be provided by a more compact specification designed specifically for such archival purposes. The data in the specification contained in this document is intended for consumption by clients intending to provide real-time (or semi-real-time) transit advice and is designed as such.
 
+## Versioning
+The version of the GBFS documentation in which a feed is encoded is declared in all file headers. See [Output Format](#output-format).
```
barbeau: I suggest

>The version of the GBFS documentation in which a feed is encoded is declared in all file headers.

...changes to:

>The version of the GBFS specification to which a feed conforms is declared in the `version` field in all files.

gbfs.md
```diff
@@ -34,12 +34,23 @@ This specification has been designed with the following concepts in mind:
...
+
+GBFS Best Practice defines that:
+* _GBFS producers_ should provide an endpoint that conforms to the current long term support (LTS) branch and the latest release branch (within 3 months of release). See [specification versioning](README.md#specification-versioning)
```
barbeau: I think this needs to be clarified a bit. Suggested text:

>_GBFS producers_ should provide endpoints that conform to both the current specification long term support (LTS) branch as well as the latest release branch.  Producers are expected to support new releases of the GBFS spec within 3 months of a new spec `MAJOR` or `MINOR` version release.

gbfs.md
```diff
@@ -34,12 +34,23 @@ This specification has been designed with the following concepts in mind:
...
+* _GBFS consumers_ should, at a minumum, consume the current LTS branch. It highly recommended that GBFS consumers support later releases.
```
barbeau: Suggested change:

>_GBFS consumers_ should, at a minumum, consume the current LTS branch...

...to:

>_GBFS consumers_ should, at a minumum, support the current LTS branch...

gbfs.md
```diff
@@ -34,12 +34,23 @@ This specification has been designed with the following concepts in mind:
...
+
+Default GBFS feed URLs, e.g. `https://www.example.com/data/gbfs.json` or `https://www.example.com/data/fr/system_information.json` should direct consumers to the feed encoded according to the current LTS documentation branch.
```
barbeau: I would avoid the word `encoded` - how about `the feed that conforms to the current...`

gbfs.md
```diff
@@ -34,12 +34,23 @@ This specification has been designed with the following concepts in mind:
...
+
+
 ## Files
 This specification defines the following files along with their associated content:
 
 File Name                   | Required                |       Defines
 --------------------------- | ----------------------- | ----------
 gbfs.json                   | Optional                | Auto-discovery file that links to all of the other files published by the system. This file is optional, but highly recommended.
+gbfs_versions.json          | Optional                | Lists all  feed endpoints published according to versions of the GBFS documentation.
```
barbeau: ` all  feed` -> ` all feed` (extra space)

gbfs.md
```diff
@@ -96,13 +107,16 @@ Field Name          | Required  | Defines
 --------------------| ----------| ----------
 last_updated        | Yes       | Integer POSIX timestamp indicating the last time the data in this feed was updated
 ttl                 | Yes       | Integer representing the number of seconds before the data in this feed will be updated again (0 if the data should always be refreshed)
+version             | Yes       | GBFS version number in which feed data is encoded, according to the versioning framework.
```
barbeau: `GBFS version number in which feed data is encoded` -> `GBFS version number to which the feed conforms`

README.md
```diff
@@ -40,6 +40,26 @@ The general outline for changing the spec has 4 steps:
 3.	Find at least one GBFS producer to implement and test the proposed change.
 4.	Submit a final request-for-comments on the proposed change to the issue discussion. If no outstanding issues are identified after one week’s time, and there is general agreement that the proposed change is worthwhile and follows the GBFS guiding principles outlined below, the proposal will be officially adopted.
 
+## Specification Versioning
+To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. A git tag in the form of `X.Y` establishes semantic versions. Git tags batch pull requests into versions.
```
barbeau: `Git tags batch pull requests into versions.` -> `Multiple changes (commits) may be batched into a single new release.`

gbfs.md
barbeau: It would be good to indicate the data type (string) here as well.

gbfs.md
```diff
@@ -115,46 +129,85 @@ The following fields are all attributes within the main "data" object for this f
...
+          "url": "https://www.example.com/gbfs/1.0/en/station_information"
         }
       ]
     },
     "fr" : {
       "feeds": [
         {
           "name": "system_information",
-          "url": "https://www.example.com/gbfs/fr/system_information"
+          "url": "https://www.example.com/gbfs/1.0/fr/system_information"
```
barbeau: Are we going to remain silent on how versioning should be handled in the URL path?

gbfs.md
barbeau: I would also change the `should` here to `must`.

gbfs.md
```diff
@@ -115,46 +129,85 @@ The following fields are all attributes within the main "data" object for this f
...
 ```
 
+### gbfs_versions.json
+Each expression of a GBFS feed describes all of the versions that are available.
+
+The following fields are all attributes within the main "data" object for this feed.
+
+Field Name              | Required    | Defines
+------------------------| ------------| ----------
+_versions_              | Yes         | Array that contains one object, as defined below, for each of the available versions of a feed.
```
barbeau: I suggest that we add:

>The array must be sorted by increasing `MAJOR` and `MINOR` version number.

gbfs.md
```diff
@@ -115,46 +129,85 @@ The following fields are all attributes within the main "data" object for this f
 
 Field Name              | Required    | Defines
 ------------------------| ------------| ----------
+stations          | Yes       | Array that contains one object per station in the system as defined below
```
barbeau: I think this line should be removed:

>stations          | Yes       | Array that contains one object per station in the system as defined below

barbeau:
Thanks @antrim! Looks pretty good, I suggested some changes.

README.md
```diff
@@ -40,6 +40,26 @@ The general outline for changing the spec has 4 steps:
...
+* Changing the data type or semantics of an existing field
+
+Examples of non-breaking changes include:
+ 
+* Adding or removing an optional endpoint or field
+* Adding or removing enum values
+* Modifying documentation or spec language in a way that clarifies semantics or recommended practices
+
+### Version Release Cycles
+* There is no strict limitation on the frequency of MAJOR releases, but the GBFS community aims to limit the MAJOR releases to 2 or less every 12 months. To limit releases, breaking changes can be batched together.
```
heidiguenin: less -> fewer

heidiguenin (commented):
Reviewing after @barbeau means lots of thumbs-up responses from me. Looking good!

README.md
antrim: I was intentionally silent so that MINOR changes could be batched or not batched, depending on their timeline. My assumption is that if changes occur soon after each other, we might as well batch. But, with backwards-compatibility issues absent, it doesn't make sense to hold back changes too long for a batched version bump. I'll explicitly state this.

gbfs.md
antrim: Oops! Good catch. Fixed.

gbfs.md
antrim: What do you think? Should we offer a recommendation? As this is written, it's up to the discretion of the feed producers. The examples offer suggestions.

README.md
antrim: I made the suggested changes.

gbfs.md
antrim: I made the suggested changes.

gbfs.md
antrim: I made the suggested changes.

gbfs.md
antrim: I made the suggested changes.

gbfs.md
antrim: I made the suggested changes.

gbfs.md
antrim: Fixed

gbfs.md
antrim: I made the suggested changes.

README.md
antrim: I made the suggested change.

gbfs.md
antrim: I made the suggested changes.

README.md
antrim: Fixed.

antrim (commented):
Thanks for your review, @barbeau and @heidiguenin. I've modified the proposal according to most of the suggestions. There is one outstanding question around URL paths and versions.

Proposed next steps
* Let's keep this open until Tuesday (UTC) of next week for review and comment.
* And then call a vote on Wednesday (UTC). Assuming there is support for this proposal, I propose that we also call a vote on changes that break backwards-compatibility at the same time, preparing to implement the versioning process.

gbfs.md
jcn: Perhaps we should state that a URL must never change its major version. For example, if a feed doesn't want to include a version in the path, that's fine, but it must not change from v1 to v2 ever. 

This actually brings up another question - can a feed change from 1.0 to 1.1? Technically this should not break anything, and it would mean that publishers could have, say, `https://www.example.com/gbfs/1/fr/system_information` for their feed that conformed to the 1.X spec. Would this be confusing for consumers?

jcn:
A question about the transition from our unversioned to versioned world: should we just state that all unversioned feeds should be considered `1.0`? The current spec is considered an implicit 1.0 anyway, so this would help in terms of moving forward, and for consumers of feeds that have not upgraded yet.

README.md
```diff
@@ -40,6 +40,26 @@ The general outline for changing the spec has 4 steps:
 3.	Find at least one GBFS producer to implement and test the proposed change.
 4.	Submit a final request-for-comments on the proposed change to the issue discussion. If no outstanding issues are identified after one week’s time, and there is general agreement that the proposed change is worthwhile and follows the GBFS guiding principles outlined below, the proposal will be officially adopted.
 
+## Specification Versioning
+To enable the evolution of GBFS, including changes that would otherwise break backwards-compatibility with consuming applications, GBFS documentation is versioned. A git tag in the form of `X.Y` establishes semantic versions. Multiple changes (commits) may be batched into a single new release.
```
jcn: I think that git tags technically have to start with a letter, so the tag will be `vX.Y` where the version name itself will be `X.Y`

gbfs.md
```diff
@@ -34,12 +34,25 @@ This specification has been designed with the following concepts in mind:
...
 
+## Versioning
+The version of the GBFS specification to which a feed conforms is declared in the `version` field in all files. See [Output Format](#output-format).
+
+GBFS Best Practice defines that:
+
+_GBFS producers_ should provide endpoints that conform to both the current specification long term support (LTS) branch as well as the latest release branch within at least 3 months of a new spec `MAJOR` or `MINOR` version release. See [specification versioning](README.md#specification-versioning)
+
+_GBFS consumers_ should, at a minumum, support the current LTS branch. It highly recommended that GBFS consumers support later releases.
+
```
jcn, line 45: Is this "should" or "must?" If a producer does not provide LTS and the latest feed, are there any repercussions?

gbfs.md
```diff
@@ -34,12 +34,25 @@ This specification has been designed with the following concepts in mind:
...
+Default GBFS feed URLs, e.g. `https://www.example.com/data/gbfs.json` or `https://www.example.com/data/fr/system_information.json` must direct consumers to the feed that conforms to the current LTS documentation branch.
+
+
 ## Files
 This specification defines the following files along with their associated content:
 
 File Name                   | Required                |       Defines
 --------------------------- | ----------------------- | ----------
 gbfs.json                   | Optional                | Auto-discovery file that links to all of the other files published by the system. This file is optional, but highly recommended.
+gbfs_versions.json          | Optional                | Lists all feed endpoints published according to versions of the GBFS documentation.
```
jcn, line 55: I'm a little confused about where this sits. Does this file get published for every version? i.e. does this file appear in 1.0, 1.1 and 1.2 locations? Regardless, I propose we make it required in order to be able to find all of the relevant feeds.

gbfs.md
```diff
@@ -96,13 +109,16 @@ Field Name          | Required  | Defines
...
+version             | Yes       | String - GBFS version number to which the feed confirms, according to the versioning framework.
 data                | Yes       | JSON hash containing the data fields for this response
 
+
 Example:
 ```json
 {
   "last_updated": 1434054678,
   "ttl": 3600,
+  "version":"1.0",
```
jcn: For consistency with the rest of the documentation, wherever new JSON fields were added, can we add a space after the `:`?

gbfs.md
```diff
@@ -120,41 +136,78 @@ _language_              | Yes         | The language that all of the contained f
...
+  \- version            | Yes         | String identifying the semantic version of the feed in the form X.Y.
+  \- url                | Yes         | URL of the corresponding gbfs.json endpoint.
+  
+```json
+{
+  "last_updated": 1434054678,
+  "ttl": 0,
+  "version":"1.0",
+  "data": {
+    [
```
jcn: This should be 

```
"versions": [
```

(you're missing the key name)

gbfs.md
barbeau: One challenge for dictating versions in URL paths is that it could conflict with existing versioning schemes or numbers used by producers in URL paths. For example, HOPR currently uses integers in the URL path to identify systems - https://gbfs.hopr.city/api/gbfs/6 vs https://gbfs.hopr.city/api/gbfs/8.

So there could be significant backwards compatibility concerns.

If we require the auto-discovery file (see proposal at https://github.com/NABSA/gbfs/pull/189), I'm inclined to let producers manage their own URL paths.

README.md
barbeau: I would state this explicitly just so there isn't any confusion.

gbfs.md
antrim: This is framed as a "Best Practice" to work within the bounds of the authority of specification authors. This is a recommendation, not a requirement. Making any sort of requirement with repercussions is the domain of cities and other governments and/or bikeshare contract holders.

gbfs.md
antrim: This file would sit in all the versions.

Proposal: Let's keep this optional for its incorporation into a "1.0" version. Technically, the change is backwards-compatible if we make the file optional.

We could make this required in a future backwards compatibility breaking MAJOR change.

gbfs.md
antrim: Good catch. Changed.

gbfs.md
antrim: Good catch. Fixed.

README.md
antrim: Fixed

gbfs.md
antrim: @jcn I just pushed a change that responds to what you pointed out. Please let me know what you think…

antrim:
Hi @jcn: Yes, we'd call all feeds that don't declare a version 1.0. (But there are bound to be variations in the interpretation. :-)

antrim:
Thanks for all the thoughtful feedback. I've made several changes to the proposal.

I hereby call a vote on this proposal. Voting will be open for 7 full days, until 11:59PM UTC on Friday, Nov 22.

Please vote for or against the proposal to add versioning, and include the organization for which you are voting in your comment.

gcamp:
+1 for Transit

gbfs.md
jcn: A clarifying question: what does "It highly recommended that GBFS consumers support later releases." mean?

gbfs.md
jcn: > We could make this required in a future backwards compatibility breaking MAJOR change.

Let's update this field to indicate that it will probably become required in a future version so producers are encouraged to start publishing now.

gbfs.md
```diff
@@ -120,41 +136,75 @@ _language_              | Yes         | The language that all of the contained f
...
+  "data": {
+  	"versions":
+		[
+		  {
+			 "version":"1",
+			 "url":"https://www.example.com/gbfs/1/gbfs"
+		  },
+		  {
+			 "version":"2",
+			 "url":"https://www.example.com/gbfs/2/gbfs"
```
jcn, line 201: Just a formatting note, can we add the spaces between the ":" and value here as well? Thanks!

gbfs.md
```diff
@@ -96,13 +109,16 @@ Field Name          | Required  | Defines
...
+version             | Yes       | String - GBFS version number to which the feed confirms, according to the versioning framework.
 data                | Yes       | JSON hash containing the data fields for this response
 
+
 Example:
 ```json
 {
   "last_updated": 1434054678,
   "ttl": 3600,
+  "version": "2.1",
```
jcn: Just for clarity (because we're still at 1.0) I think all the example versions should be `1.0`.

johnpena:
+1 for Lime

gbfs.md
antrim: Makes sense. Updated. Since it's a small clarifying change, I think we can continue with the existing vote timeline (i.e. we don't need to call a re-vote).

antrim:
Voting has closed. Both Lime and Transit voted for the change. No organizations voted against.

Current GBFS governing rules say that at least one producer and consumer have to **implement** the change before it is adopted, I propose that we merge to the master branch (if there are no objections) for these reasons:

* The change includes new governance rules.
* An agreed upon versioning scheme is a prerequisite for some of the proposed changes that would break backwards-compatibility. Deciding on a versioning scheme now will allow GBFS stakeholders to better plan.
* This change change itself is backwards compatible. Consuming software would not need to be changed until a new version is released.

Does anyone have objections to merging into the master branch?

jcn:
I am in favor of this change and in merging it at the end of the voting period but for the sake of process we should wait for the actual end as per your original call for a vote.

> I hereby call a vote on this proposal. Voting will be open for 7 full days, until 11:59PM UTC on Friday, Nov 22.

antrim:
Oops. Sorry about that confusion with the dates. Voting is of course open through Friday.

yocontra:
+1 for Stae

tmontes:
May I suggest that if this PR is merged, as it seems it will, then the repository is tagged as soon as possible: we should have, I suppose, at least a version 1.0 tag that pinpoints which of the more than 20 commits claiming to be version 1.0 in their version histories is actually version 1.0. :-)

Thanks!

tmontes:
PS: For what it's worth, as an independent consultant working on an MDS/GBFS related project, I'm in favour of the proposed changes.

evansiroky:
+1 from IBI Group.

mplsmitch:
+1

sven4all:
+1

antrim:
The change passes! We had 6 votes in favor from consumers or operators and 0 votes against:

* Consumers (4): [Transit](http://transit.app), IBI Group, [Openbike](https://openbike.nl/), [Stae](https://stae.co/)
* Operators / producers (2): Lime, Motivate
* Unaffiliated (2) - not counted

Next steps:
* I'll merge the PR.
* We'll begin to implement the versioning scheme by tagging releases. There are a number of PRs open to voting or soon to be, so this is timely!

heidiguenin:
We'd love to make this an official part of the spec, but first we need to see this change being implemented. Could you comment here if your organization has implemented this?

@johnpena @gcamp @contra @evansiroky @sven4all ?

PR #197, opened 2019-11-26
add MOVO feeds

superruzafa:


antrim (approved):
Merging MOVO feeds -- which appear to be the first that conform to the new versioning scheme! (#188)

PR #202, opened 2019-12-11
remove dead roots from systems.csv

yocontra:
Our validator has been throwing warnings about these feeds for a while, figured I should update the list to note their status.

Wow Citybike: Bankrupt as of May 2019
BIXI Montreal: Not sure, some other GBFS URLs are resolving but not the root
Gruv Chicago: Pilot ended Oct 15 2019 and pulled out
Pike Ride: Program still exists, but I think they quit using BCycle to operate it (BCycle does not have colorado springs on their city list anymore)
Sobi Longbeach: Domain expired, I think this was renamed to https://www.longbeachbikeshare.com/ ? SocialBicycles (parent company) still reports it as an active program

### DIFF
--- a/systems.csv
+++ b/systems.csv
@@ -21,6 +21,7 @@ BR,Bike Itaú - Sampa,"São Paulo, BR",bike_sampa,https://bikeitau.com.br/bikesa
 BR,Bike Itaú - Riviera,"Riviera, BR",rivi_bike,https://www.rivieradesaolourenco.com/rivibike-2/,https://riviera.publicbikesystem.net/ube/gbfs/v1/
 BR,Bike VV,"Vila Velha, BR",bike_vv,https://www.bikevv.com.br/,https://vilavelha.publicbikesystem.net/ube/gbfs/v1/
 CA,Bike Share Toronto,"Toronto, ON",bike_share_toronto,https://www.bikesharetoronto.com/,https://tor.publicbikesystem.net/ube/gbfs/v1/
+CA,BIXI-Montreal,"Montreal, QC",Bixi_MTL,https://www.bixi.com/,https://api-core.bixi.com/gbfs/gbfs.json
 CA,HOPR Ottawa,"Ottawa, ON",6,https://gohopr.com/velogo/,https://gbfs.hopr.city/api/gbfs/6/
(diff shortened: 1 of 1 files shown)


madupras (approved):

madupras:
Approval for the removal of Reykjavik

mjarrett:
FYI Bixi (Montreal) shuts down for the winter and restarts in the spring (I'm not associated with Bixi).

yocontra:
Updated the initial ticket based on info I found about the different programs. All seem to be good candidates for removal except BIXI - if anyone knows somebody at BIXI who can help get their feed back online let me know.

LeoFrachet:
As @mjarrett said, there is nothing to be done for BIXI. The bikes *and the stations themselves* have been removed from the streets for the winter as usual. It will be back in the Spring

yocontra:
@LeoFrachet Is it valid for a system that is temporarily closed for the season to return a 404 for the root gbfs.json? I would expect that bikes/stations/etc would be empty but that the root would still be available.

I'm working on building out some deeper GBFS validation tools so this would be good to note in the spec or have in the best practices document for future reference. For this PR I'll just take BIXI out.

LeoFrachet:
@gcamp & @MuteQ, does BIXI's GBFS gives 404 every Winter? Do you have an opinion on what they should be doing? Is there a "This is Canadian Winter" status possible in GBFS? (Also @timMillet)

jcn:
It would be ideal if the bixi systems just returned empty arrays for bikes and stations when those items are missing. In addition, this is the use case that was anticipated for the [system_calendar](https://github.com/NABSA/gbfs/blob/master/gbfs.md#system_calendarjson) endpoint.

mplsmitch:
Seasonal systems have never gone to 404 in the past. They usually set all their stations to show as not installed. The Bixi feeds are still live but I haven't been able to find gbfs.json

Here's the station feed:
http://api-core.bixi.com/gbfs/en/station_information.json

MuteQ:
@LeoFrachet I don't remember `gbfs.json` returning 404 previous winters. As others have mentioned, I'd expect a valid response on `gbfs.json` and an empty list of stations in `station_information.json`. I'll reach out to contacts at BIXI to ask what's going on.

MuteQ:
BIXI restored it's `gbfs.json` feed. https://api-core.bixi.com/gbfs/gbfs.json

PR #203, opened 2019-12-20
New GBFS enabled provider added

artem-vereshchaka:
Added new record to systems.csv with GBFS enabled provider


jcn (changes requested):
This does not seem to be a valid feed - the `free_bike_status` endpoint is returning a 500. I'm not sure if it's temporarily broken or not live, but it shouldn't be merged in until it's fixed.

artem-vereshchaka:
@jcn , the given city is closed right now (due to weather conditions). Spec says that it should work in this way if there are no available vehicles.
We are about to open more cities soon that will operate in winter also.

I will also update PR in order to add another operational city.

heidiguenin:
@jcn - what are your thoughts on this given the discussion in #202 ?

jcn:
> Spec says that it should work in this way if there are no available vehicles.

@artem-vereshchaka The original problem I raised was that the endpoint was returning a 500 error HTTP response code and both proposed new feeds are now returning a 400. The spec does not indicate that this is a valid response, and as per the discussion in #202 that @heidiguenin references, the bixi systems that are closed seasonally return valid feeds when they are closed, just with no available vehicles. 

These feeds should be updated in the same way in order to be included in the systems.csv file.

artem-vereshchaka:
I don't think we should proceed with this PR anymore and merge it.
Cities do not use this repo as a source, so I assume we are not required to be mentioned here as well.

If anything -- just ping me directly.

PR #214, opened 2020-01-27
Example not according to specification.

sven4all:
There are examples included within this documentation that are not according to the specification. An ID field should be a string as described in the Field Types section.


heidiguenin:
Thanks, @sven4all  - we missed that during the spec clean-up!

PR #224, opened 2020-03-04
Yet another formatting PR

evansiroky:
This is a somewhat opinionated PR on formatting changes. It does the following:

### Make all markdown tables be in compact mode instead of fixed-width

Current experience looking at diffs: _"Wow, look every single row in a table has changed! Oh wait no, it's just changes in spaces..."_

Screen Shot 2020-03-03 at 6 12 35 PM

It's a pain when editing to create fixed-width text. And when it's time to look at a diff from the previous text, it appears that the entire table has changed. Therefore, this PR makes the change once and for all to compressed formatting for markdown tables.

### Remove all italic and highlighted field name formatting

Screen Shot 2020-03-03 at 6 15 49 PM

There is no rhyme or reason to why some fields have both code and italic formatting. Therefore, all fields that have this style are reverted to simply code highlighting.

### Correct various formatting errors

There were various missing closing formatting characters that were added.

### Remove all trailing whitespace

Some editors remove this by default on save. This PR gets rid of it now in a PR dedicated to formatting rather than in a PR where someone unlucky will get a bunch of complaints about all these random whitespace changes.


gcamp:
👍 on the change. Tip on Github to visualize markdown changes you can use the rendering button "to display rich diff", it makes everything very easy to visualize 👌

barbeau:
@evansiroky Thanks for working on this! I noticed the other day that some tables wrap text while others don't, which we should also fix. Looking at the rich diff here, it looks like `free_bike_status.json` still doesn't wrap text. Does this PR address the text wrapping too, or no?

evansiroky:
@barbeau, I'm not seeing what you are seeing. I see html <br> tags, which sort of seem to improve readability when rendered. Were you referring to that?

barbeau:
@evansiroky Below are two screenshots from the rich diff view (i.e., rendered view) - you can see that the right column in `system_information.json` wraps the text, while the `free_bike_status.json` right column doesn't wrap text - you need to scroll right to see it.

It's fine if this PR doesn't fix this - I was just curious as it was on my TODO list to fix :).

image

gbfs.md
```diff
@@ -101,11 +101,11 @@ This section defines terms that are used throughout this document.
 
 * Array - A JSON element consisting of an ordered sequence of zero or more values.
 * Object - A JSON element consisting of key-value pairs (fields).
-* Boolean - One of two possible values, `1`=true and `0`=false. *Beta (v2.0-RC): Boolean values must be JSON booleans, not strings (i.e. true or false, not `"true"` or `"false"`).
+* Boolean - One of two possible values, `1`=true and `0`=false. \*Beta (v2.0-RC): Boolean values must be JSON booleans, not strings (i.e. true or false, not `"true"` or `"false"`).
 * Date - Service day in the YYYY-MM-DD format. Example: `2019-09-13` for September 13th, 2019.
 * Email - An email address. Example: `example@example.com`
 * Enum (Enumerable values) - An option from a set of predefined constants in the "Defines" column.
-_Example: The `rental_methods` field contains values `CREDITCARD`, `PAYPASS, etc...
+Example: The `rental_methods` field contains values `CREDITCARD`, `PAYPASS`, etc..._
```
jcn: Looks like a stray trailing `_`

PR #235, opened 2020-04-16
Add Vélib Metropole to systems.csv

pichot:
Adding Paris to `systems.csv`.

Set `System ID` to `Paris` as defined in its [system_information.json](https://velib-metropole-opendata.smoove.pro/opendata/Velib_Metropole/system_information.json). Not sure if this is the correct choice.


bddq (approved):
Using `Paris` as system_id looks good.
According GBFS specification:
> It is up to the publisher of the feed to guarantee uniqueness. This value is intended to remain the same over the life of the system.

heidiguenin:
@pichot It would be better if the `system_id` in the system_information.json was more descriptive, such as the examples given in the spec (`bcycle_austin` or `biketown_pdx`). Currently there are many `system_id`s that do not follow that example, and we're hoping to work with operators to revise those in the coming year. Looks like you're not the operator, though - so that's just more for your information!

pichot:
@heidiguenin I definitely agree a more descriptive `system_id` would be better. Let me know if you'd like me to change it in the `systems.csv`. (Alas, I'm not the operator.) I'd probably go with something like `velib_metropole_paris`.

heidiguenin:
@pichot I'll go ahead and merge it as-is - we do need to stick to the name that's in system_information.json

Thanks!

PR #236, opened 2020-04-23
Transition from "bike" to "vehicle" to support new vehicle types

heidiguenin:
This PR changes references to "bike" or "bikes" (short of the name "GBFS" itself) to the more generic "vehicle" or "vehicles" throughout.

During the [vehicle types definitions discussion](https://github.com/NABSA/gbfs/pull/136), switching the references from "bikes" to the more generic "vehicles" came up repeatedly, but in order to keep those updates backwards compatible, we're tackling this piece separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # General Bikeshare Feed Specification
-Documentation for the General Bikeshare Feed Specification, a standardized data feed for bike share system availability.
+Documentation for the General Bikeshare Feed Specification, a standardized data feed for shared mobility system availability.
 
 ## What is GBFS?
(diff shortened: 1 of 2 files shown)


heidiguenin:
I've opened the PR for [your request](https://github.com/NABSA/gbfs/pull/136#issuecomment-615547928), @charlesjump. I think this should cover everything - let me know!

bddq (commented):
Should we transition `bikes` to `vehicles` filenames/properties in a specific version to not break existing apps using GBFS?

heidiguenin:
If this change were voted on and passed by the community, it would be merged into a future major release.

heidiguenin:
With all of the updates to the specification since this was opened, I'm closing it & will re-do when we're closer to a v3.0 change process!

PR #239, opened 2020-04-24
fix: license_id AND license_url should be empty for default license

barbeau:
I was looking at the spec today, and as part of this I was looking at the v3.0-RC fields for licensing - license_id and license_url.

Looking back at the original discussion for these fields in https://github.com/NABSA/gbfs/pull/182, I believe the intent here is that if in a v3.0-RC feed BOTH fields are left blank, then the feed falls under a default license (Creative Commons Universal Public Domain Dedication).

However, the draft spec currently uses `or`, which could be interpreted as if one of these fields is blank, then the default applies.

This PR changes the language to use `and` instead of `or`, which I believe is the original intent of https://github.com/NABSA/gbfs/pull/182.


heidiguenin:
Good catch @barbeau  - you're right that the change reflects the understanding behind and intention of the initial PR.

PR #241, opened 2020-04-29
improve consistency in required column

eraac:
All others optional field use `Optional` but one. It's not documented under `Term Definitions`, I guess is a mistake


bddq (approved):

eraac:
In my point of view, the change should be more

- from `Optional` to `No`
- from `Conditionally required` to `Conditional`

But this change is a lot bigger and I'm not sure is very relevant/useful

heidiguenin:
@Eraac  I agree. It's a holdover from the GTFS inspiration. I'll add it to the list of to-dos :)

PR #194, opened 2019-11-20
Proposal: Improving pricing information

kanagy:
This pull request is to discuss a [proposal (Google Doc, revised Nov 20)](https://docs.google.com/document/d/1Y5Muz_yJpwy-_zlfPPtMkpofZxV72zM09wV_h8eXIJ8) to improve pricing information in GBFS, so that consumer apps are able to display an approximate cost of a micromobility ride. 

The proposal is based on discussions in the GBFS Indianapolis workshop. Please make comments directly in the Google Doc. Once we have arrived at a proposal that is an implementation candidate, then I'll edit the GBFS spec in this PR, get final comments and put the PR for voting.

PR #252, opened 2020-07-27
Extend system_pricing_plans.json

josee-sabourin:
**Context**
In October 2019 during the [GBFS Developers' workshop](https://docs.google.com/document/d/1TXf3U7gbrJrS_r2PmTrfbJspNydAyd42xU3ctNgzNLM/edit?usp=sharing), before the NABSA annual conference, Kristian (@kanagy) shared a proposal for enhancing pricing information in GBFS. Participants agreed that it was a promising "minimum viable proposal" (MVP), and Kristian (@kanagy) opened PR #194. The proposal was discussed again at MobilityData's Shared Mobility Workshop in Paris in December 2019, and participants agreed it was a good place to start. However, we have received very little producer input on the proposal, and we will need at least one producer to vote on the proposal and to implement it. Starting in 2020, the scope of GBFS has been expanding to be able to support carsharing, which ultimately complicated the needs of a minimum viable proposal for extending system_pricing_plans.

**Pull Request**
MobilityData drafted up a proposal to cover most existing use cases along with future carsharing use cases. We heard during initial conversations with stakeholders that it was overly complex and would be difficult to implement all at once. We've scaled it back to a minimum viable proposal, while still being able to support a handful of carsharing pricing schemes. This specific PR is representative of our MVP, the full proposal can be found [here](https://docs.google.com/document/d/1xzC0nS3M-ora32k5RbfhFCsfZp4B-wVKK2-pOokpxmQ/edit?usp=sharing). 

The specific changes in the MVP are: 

- Add scheduled pricing properties (`day`, `month`, `start_time`, and `end_time` of pricing plans) to allow for future trip planning. 
- Add `vehicle_type_id` to reflect the various pricing schemes that are available to certain vehicle types.
- Add rates to reflect a majority of systems that charge per kilometre or per minute for example. 
- Add price increases that take effect after a certain period of time. 
- Add `surge_pricing` to support demand-driven pricing. 

Some of our open questions are: 
- Should we limit the unit enums to just minute and kilometers? Keeping in mind some carsharing plans are charged per week. 
- Should all units be abbreviated, if so, to which standard? 

Please let us know what you think of these open questions as well as general comments/questions/concerns!


gbfs.md
```diff
@@ -571,14 +572,90 @@ Describes pricing for the system.
...
+     }
+   ]
+   "flat_segments": [
+     {
+       "start": 25,
+       "price": 3,
+       "interval": 5,
+     }
+   ]
+ }
```
jiffyclub: Per the description above `variable_price` is an array.

```suggestion
"variable_price": [
    {
        "variable_unit": "km",
        "rate_segments": [
            {
                "start": 10,
                "rate": 1
            },
            {
                "start": 25,
                "rate": 0.5
            }
        ],
        "flat_segments": [
            {
                "start": 25,
                "price": 3,
                "interval": 5
            }
        ]
    }
]
```

jiffyclub: Would this be functionally the same thing? `rate_segments` seems like a special case of `flat_segments` where the interval is 1.

```jsonc
"variable_price": [
    {
        "variable_unit": "km",
        "flat_segments": [
            {
                "start": 10,
                "price": 1,
                "interval": 1
            },
            {
                "start": 25,
                "price": 0.5,
                "interval": 1
            }
        ]
    },
    {
        "variable_unit": "km",
        "flat_segments": [
            {
                "start": 25,
                "price": 3,
                "interval": 5
            }
        ]
    }
]
```

jiffyclub:
Something that I don't think could be reflected in this model is Gig's scheme where they have several different pricing schemes and charge you whichever one would result in the lowest total at the end of your rental: https://gigcarshare.com/rates/

josee-sabourin:
Hi all! After the developers workshop and some individual stakeholder feedback - we now have a [Pricing v2.1!](https://docs.google.com/document/d/1gJYR2u-rfZI0-HxHkhiXz5nkh7vR1JF2Ux4NwiCI4Gg/edit?usp=sharing). 

Some notable changes: 
- removal of units in exchange for `per_km_pricing` and `per_min_pricing`
- removal of `variable_price` and `flat_price` in exchange for defining `interval` every time, as suggested by @jiffyclub above
- addition of `end` to the Core MVP to account for rates that are replaced by the following rate. 

To note: 
- currently `vehicle_type_id` is added in this file, pending the vote results of #261, this may be altered to have `pricing_plan_id` in `free_bike_status.json` for every vehicle, which would allow for easier dynamic pricing based on location. 

@jiffyclub to model Gig would require the use of GBFS-CapPricing which is in the Google Doc proposal but not this MVP, I've added Gig as an example in [Section 7](https://docs.google.com/document/d/1gJYR2u-rfZI0-HxHkhiXz5nkh7vR1JF2Ux4NwiCI4Gg/edit#heading=h.udmja1pdhtpo) of the document!

gbfs.md
josee-sabourin: I've added this to the new round of changes, thanks for this suggestion!

sfaubert1:
Hi @josee-sabourin! This looks really good and it's clear, but I just realized that in our case we can have different `per_min_pricing` per `vehicle_type` for the same plan, and I don't see how we could clearly include this information currently without duplicating the same plan in the feed to include details for each vehicle type...

For example, a plan could have a single `price` to have access to different vehicle types, but then trip fees would be depending on the vehicle type used with this plan, such as:
- 0.50 per half hour if the vehicle type used is a regular bike (`human` propulsion_type) 
- 1.00 per half hour if the vehicle type used is an eBike (`electric_assist` propulsion_type)

I don't know if it's only applicable to us or if other systems would have plans that are eligible for multiple vehicle types and where the trip fees (pricing per km or min) would be different based on the vehicle type used within a given plan.

In this case, could we add the specific `vehicle_type_id` as an optional field within each section for `per_km_pricing `/ `per_min_pricing` (if not present, would apply to all vehicle types eligible for this plan) or should we define it somewhere else? (note that if we add it, maybe rename one of the `vehicle_type_id` fields since there would be 2 fields with the same name that means different things: one at the plan level and another at the additional pricing level)

Or maybe I just missed something and there's an easier way to do this?

josee-sabourin:
Hi @sfaubert1 , you are correct you would need have separate plans for each vehicle type, this was chosen to avoid parsing errors of having multiple vehicle type ID's scattered in the same plan. I'm wondering what implications your suggestion would  have on the consumer side, @kanagy, what are your thoughts on this?

kanagy:
Hi @sfaubert1, in this case I think the clearest and easiest thing to do is to treat these as two separate plans in the feed (i.e. duplicate the plan, first plan charges 0.5 per hour, second plan charges 1 per hour). I also personally think of these as two different pricing schemes, each applying to different vehicle type.

I would be against the suggestion of complicating the feed to reduce minor duplication (is that the goal?). The feed should be kept simple and a common denominator of different schemes (as long different schemes can be represented, we're good).

gbfs.md
```diff
@@ -127,6 +127,7 @@ Example: `41.890169` for the Colosseum in Rome.
 Example: `12.492269` for the Colosseum in Rome.
 * Non-negative Integer - An integer greater than or equal to 0.
 * Non-negative Float - A floating point number greater than or equal to 0.
+* Float - A floating point number.
```
kanagy: SHould this state also precision for the floating point number?

gbfs.md
```diff
@@ -571,15 +572,89 @@ Describes pricing for the system.
 
 Field Name | Required | Type | Defines
 ---|---|---|---
-`plans` | Yes | Array | Array of objects as defined below.
+`plans` | Yes | Array | Array of objects as defined below.<br /><br /> In the event of colliding plans, the earlier plan (ir order of the JSON file) takes precedence.
```
kanagy: Can we straight disallow colliding plans? A plan should be invalid when it collides with other plans. Allowing collisions just introduces potential errors (producers adding a colliding plan by mistake and consumers not being to catch this during feed validations, as spec says that it is correct).

gbfs.md
```diff
@@ -571,15 +572,89 @@ Describes pricing for the system.
...
 \-&nbsp;`plan_id` | Yes | ID | Identifier for a pricing plan in the system.
+\-&nbsp;`vehicle_type_id` | Optional | Array | `vehicle_type_id` of the vehicle eligible for this pricing plan as described in [vehicle_types.json](#vehicle_typesjson-added-in-v21-rc).<br /><br />If this field is empty, the plan applies to all vehicle types defined in the dataset.<br /><br />In the case of a vehicle type being attributed to different plans, all plans associated with the vehicle type are valid.
```
kanagy: Would love to hear community's opinion on two approaches here:
1) The one defined here where a pricing plan references vehicle type feed via vehicle_type_id.
2) One where vehicle type feed references pricing plan via plan_id (i.e. the opposite of 1). I personally slightly prefer this one (because existing plan_id in the spec can be reused for referencing).

gbfs.md
```diff
@@ -571,15 +572,89 @@ Describes pricing for the system.
...
 \-&nbsp;`url` | Optional | URL | URL where the customer can learn more about this pricing plan.
 \-&nbsp;`name` | Yes | String | Name of this pricing plan.
 \-&nbsp;`currency` | Yes | String | Currency used to pay the fare. <br /><br /> This pricing is in ISO 4217 code: http://en.wikipedia.org/wiki/ISO_4217 <br />(e.g. `CAD` for Canadian dollars, `EUR` for euros, or `JPY` for Japanese yen.)
-\-&nbsp;`price` | Yes | Non-negative float OR String | Fare price, in the unit specified by currency. If String, must be in decimal monetary value.
+\-&nbsp;`price` | Yes | Non-negative float OR String | Fare price, in the unit specified by currency. If String, must be in decimal monetary value.<br /><br />In case of non-variable price, this field is the total price. In case of variable price, this field is the base price that is charged only once per trip (e.g., price for unlocking). See `variable_price` for details.
```
kanagy: I'm not sure what is meant by "String, must be in decimal monetary value" here. In JSON all numbers are strings. So while you're here, I think this should be updated to just say "Float" instead of "String, must be in decimal monetary value".

gbfs.md
```diff
@@ -571,15 +572,89 @@ Describes pricing for the system.
...
 \-&nbsp;`is_taxable` | Yes | Boolean | Will additional tax be added to the base price?<br /><br />`true` - Yes.<br />  `false` - No.  <br /><br />`false` may be used to indicate that tax is not charged or that tax is included in the base price.
 \-&nbsp;`description` | Yes | String | Customer-readable description of the pricing plan. This should include the duration, price, conditions, etc. that the publisher would like users to see.
+\-&nbsp;`per_km_pricing` | Optional | Array | Array of segments when the price is a function of distance travelled, displayed in kilometers.<br /><br />If this array is not provided, there are no variable prices based on distance.
```
kanagy: Can you add a note that this is added to the base `price` and to the `per_min_pricing` variable price. Same note would be useful on the `per_min_pricing` field to make it more clear that all these fields are added together.

gbfs.md
```diff
@@ -571,15 +572,89 @@ Describes pricing for the system.
...
+&emsp;&emsp;\-&nbsp;`start` | Yes | Non-Negative Integer | Number of kilometers that have to elapse before this segment starts applying.
+&emsp;&emsp;\-&nbsp;`rate` | Yes | Float | Rate that is charged after the `start`. Can be a negative number, which indicates that the traveller will receive a discount.
```
kanagy: Can you add that the `rate` is charged for each kilometer `interval` of the trip? I assume that's the case. I.e. if interval=0 rate is applied once, if interval=1 rate is applied every kilometer, if interval=5 rate is applied every 5 kilometers. That is already somehwat described in the `interval` field, but would be nice to add this here too.

gbfs.md
```diff
@@ -571,15 +572,89 @@ Describes pricing for the system.
...
+&emsp;&emsp;\-&nbsp;`interval` | Yes | Non-Negative Integer | Interval in kilometers at which the `rate` of this segment is reapplied indefinitely, unless defined by `end`.<br /><br />An interval of 0 indicates the rate is only charged once.
```
kanagy: Nit: Would phrase it as 
"Interval in kilometers at which the `rate` of this segment is either reapplied indefinitely, or if defined, up until (but not including) `end` kilometer."

gbfs.md
```diff
@@ -571,15 +572,89 @@ Describes pricing for the system.
...
+&emsp;&emsp;\-&nbsp; `end` | Optional | Non-Negative Integer | Number of kilometers that have to elapse after this segment has started before this segment stops applying.<br /><br /> If this field is empty, the price issued form this segment is charged until the trip ends, in addition to following segments.
```
kanagy: Nit: typo "form" -> "for"

Ah, can you confirm that per-description this is not "The kilometer at which the rate should stop applying" but rather "The number of kilometers that have to be travelled from start for this rate to stop applying"? If that's the case, the name of the field should be different. Perhaps `duration`? Or maybe just change the description for this to be "The kilometer after which the rate will no longer apply" (I think this is clearer).

kanagy (commented):
This looks pretty good, left some comments.

gbfs.md
josee-sabourin: This was discussed in issue #116, removing either string or float would be a breaking change that we could vote on in this proposal but would wait until v3.0 to implement, the same way it was done for licensing.

gbfs.md
josee-sabourin: Good idea, what would be an appropriate number of decimal places?

gbfs.md
josee-sabourin: I would agree, I've reflected the changes in the latest commit

gbfs.md
josee-sabourin: Added in 3ccda0532089bd3f9596d488def734dd637eea63

gbfs.md
josee-sabourin: Added in 3ccda0532089bd3f9596d488def734dd637eea63

gbfs.md
josee-sabourin: Added in 3ccda0532089bd3f9596d488def734dd637eea63

gbfs.md
josee-sabourin: Updated the description to your suggestions of "The kilometer after which the rate will no longer apply". Thanks for the suggestions and all of your comments!

josee-sabourin:
As mentioned in a [previous comment](https://github.com/NABSA/gbfs/pull/252#issuecomment-692780090), now that #261 has passed, this proposal has been updated to reflect those changes. Now that each vehicle can be represented in `free_bike_status.json` regardless of being docked or dockless, `pricing_plan_id` has been added to `free_bike_status.json` and `vehicle_type_id` has been removed from `system_pricing_plans.json`. 

~~Additionally, #216 discussed revising price type to string to accommodate feeds that use a comma in their prices rather than a period. The fields `price` and `rate` have been changed to string. While it will be voted on in this vote, the change to `price` will not be implemented until v3.0.~~

With that, I am hoping we can move to a vote for this proposal. 

I hereby call a vote on this proposal. Voting will be open for 7 full calendar days until 11:59PM UTC on October 6th, 2020.

gbfs.md
kanagy: Can just say a 32-bit floating point number (more than enough precision). And same for non-negative float description.

gbfs.md
```diff
@@ -576,10 +578,83 @@ Field Name | Required | Type | Defines
...
 \-&nbsp;`is_taxable` | Yes | Boolean | Will additional tax be added to the base price?<br /><br />`true` - Yes.<br />  `false` - No.  <br /><br />`false` may be used to indicate that tax is not charged or that tax is included in the base price.
 \-&nbsp;`description` | Yes | String | Customer-readable description of the pricing plan. This should include the duration, price, conditions, etc. that the publisher would like users to see.
+\-&nbsp;`per_km_pricing` | Optional | Array | Array of segments when the price is a function of distance travelled, displayed in kilometers.<br /><br />Total price is the addition of `price` and all segments in `per_km_pricing` and `per_min_pricing`. If this array is not provided, there are no variable prices based on distance.
+&emsp;&emsp;\-&nbsp;`start` | Yes | Non-Negative Integer | Number of kilometers that have to elapse before this segment starts applying.
+&emsp;&emsp;\-&nbsp;`rate` | Yes | String | Rate that is charged for each kilometer `interval` after the `start`. Must be in decimal monetary value. Can be a negative number, which indicates that the traveller will receive a discount.
```
kanagy: I disagree with these being strings. These must not be formatted prices (i.e. prices one would show in the UI). These are inputs to a function that needs to be computed by consumer apps and so they must to be parseable as floats; consumer apps can then format the final price in whatever is the user's locale. Same goes for `price`, but as you said, that is unfortunately a breaking change (maybe it should say "Float for any new implementations"?).

josee-sabourin: That is a very good point, thanks for explaining that! I think the best way to do it would be to revert `rate` to floats and then vote on changing `price` to float, but that change to float wouldn't be implemented until v3.0, and then we write a warning that it will be changing and therefore any new implementations are recommended to be floats. I'll update the PR.

kanagy:
+1 from Google Maps. We don't have cycles to implement it in Q4, but we do already use a very similar and compatible format with Lime.

MuteQ:
+1 from Transit

jiffyclub:
+1 from Populus, looks like it can cover many cases. Populus, though, does not have any current plans to integrate with this endpoint.

evansiroky:
+1 from IBI Group

gbfs.md
```diff
@@ -576,10 +578,83 @@ Field Name | Required | Type | Defines
 \-&nbsp;`url` | Optional | URL | URL where the customer can learn more about this pricing plan.
 \-&nbsp;`name` | Yes | String | Name of this pricing plan.
 \-&nbsp;`currency` | Yes | String | Currency used to pay the fare. <br /><br /> This pricing is in ISO 4217 code: http://en.wikipedia.org/wiki/ISO_4217 <br />(e.g. `CAD` for Canadian dollars, `EUR` for euros, or `JPY` for Japanese yen.)
-\-&nbsp;`price` | Yes | Non-negative float OR String | Fare price, in the unit specified by currency. If String, must be in decimal monetary value.
+\-&nbsp;`price` | Yes | Non-Negative float OR String | Fare price, in the unit specified by currency. If string must be in decimal monetary value. Note: v3.0 will only allow non-negative float, therefore future implementations must be non-negative float.<br /><br />In case of non-rate price, this field is the total price. In case of rate price, this field is the base price that is charged only once per trip (e.g., price for unlocking) in addition to `per_km_pricing` and/or `per_min_pricing`.
```
sfaubert1: For the description of `price`, it could also be the base price that is charged to access the vehicles for a certain period of time (instead of being per trip). For example, a flat fee to have access to bikes for 24 hours, which is not charged per trip but once at the start of the 24h period, and then each trip can be charged per min or per km within these 24 hours.

josee-sabourin: Currently, this MVP is designed for modelling one-way trips, so passes are not easily modelled this way, however it can be done using two concurrent rates (one with an interval of 1440 minutes). We can add more specific language about `price` acting as an unlock fee, and once this is in the spec we will be sure to add a handful of examples to show how it can be used in these various ways. Hope that helps!

gbfs.md
```diff
@@ -468,6 +469,7 @@ Field Name | Required | Type | Defines
 \- `vehicle_type_id` <br/>*(added in v2.1-RC)* | Conditionally Required | ID | The vehicle_type_id of this vehicle as described in [vehicle_types.json](#vehicle_typesjson-added-in-v21-rc). This field is required if the [vehicle_types.json](#vehicle_typesjson-added-in-v21-rc) is defined.
 \- `last_reported` <br/>*(added in v2.1-RC)* | Optional | Timestamp | The last time this vehicle reported its status to the operator's backend.
 \- `current_range_meters` <br/>*(added in v2.1-RC)* | Conditionally Required | Non-negative float | If the corresponding vehicle_type definition for this vehicle has a motor, then this field is required. This value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel.
+\- `pricing_plan_id` | Optional | ID | The `plan_id` of the pricing plan this vehicle is eligible for as described in [system_pricing_plans.json](#system_pricing_plans.json).
```
sfaubert1: I understand that the `pricing_plan_id` is at the vehicle type level (instead of the vehicle type at the pricing level) to allow for easier dynamic pricing based on location, but based on my comment from September 15th (about the same plan having different pricing based on the vehicle type), this would be confusing. 

For example, if vehicle ABC is eligible for plan 123 with a certain `per_min_princing` and vehicle XYZ is also eligible for plan 123 but with another `per_min_princing`. Both ABC and XYZ vehicles would be linked to the same pricing_plan_id=123, but there would actually be 2 objects with `plan_id`=123 in `system_pricing_plans` including different `per_min_princing` based on the vehicle type and no way to know which vehicle is linked to which of the `plan_id`=123 objects in this case (without data manipulation). 

I'm curious to know if we are the only ones with this particular case? 

(FYI the "vehicle_type_id" is still in the `system_pricing_plans.json` example)

gbfs.md
sfaubert1: OK I see, thanks!

gbfs.md
josee-sabourin: This would effectively be two different pricing plans, and therefore would have two separate plan IDs, for example plan ID 123-ABC and plan ID 123-XYZ, the public name for these can be the same, but in the back end they are different. As a general rule of thumb, even if the public plan name is the same (ex: one-way), if the prices are different, they should be treated as separate plans with separate plan IDs. 

Thanks for catching the example error! I'll make sure to fix that!

sfaubert1:
+1 from PBSC

jayjonsson:
+1 from Lime

heidiguenin:

Votes in favor:
Google Maps (consumer)
Transit (consumer)
Populus (consumer)
IBI (consumer)
PBSC (producer)
Lime (producer)

No votes against.

We'll be merging, creating v2.1RC2, and then we'll follow up with those of you who said you'd be implementing. v2.1RC2 will include the recently passed #261 in addition to #269 if it passes - the vote's open, so please chime in there, too!

gbfs.md
```diff
@@ -576,10 +578,83 @@ Field Name | Required | Type | Defines
...
+    "per_km_pricing": {
+      "start": 0,
+      "rate": 0.25,
+      "interval": 1,
+    },
+    "per_min_pricing": {
+      "start": 0,
+      "rate": 0.50,
+      "interval": 1
+    }
```
kheraankit: nit:  `per_km_pricing ` and `per_min_pricing ` should be array in this example

barbeau: Thanks! We've fixed it.

kanagy:
Hey folks, this PR is marked in README and other places as being in v2.1-RC2 (https://github.com/NABSA/gbfs/blob/master/README.md#read-the-spec--version-history), but it is not actually merged into that branch. https://github.com/NABSA/gbfs/blob/v2.1-RC/gbfs.md#system_pricing_plansjson doesn't contain the fields added in this PR. Should it be merged to v2.1-RC2 too?

josee-sabourin:
Hi Kristian, I believe you may be looking at the wrong tag (your link is to v2.1-RC), take a look here at v2.1-RC2: https://github.com/NABSA/gbfs/blob/v2.1-RC2/gbfs.md, let me know if that clarifies things!

kanagy:
Wooops, my bad, thanks!

PR #254, opened 2020-07-28
infos regarding terms and conditions added to system_information

matt-wirtz:
In order to provide potential users the terms and conditions of the bike sharing system I added an URL to the system_information where a link can be provided. In addition the publishing date of the terms so one can determine if changes took place. Should especially be interesting for consumer apps providing some sort of reservation or booking.

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -247,6 +247,8 @@ Field Name | Required | Type | Defines
 `operator` | Optional | String | Name of the operator.
 `url` | Optional | URL | The URL of the vehicle share system.
 `purchase_url` | Optional | URL | URL where a customer can purchase a membership.
+`terms_url` | Optional | URL | URL where the terms and conditions are stated.
+`terms_release_date` | Optional | Date | Date when the terms and conditions where published.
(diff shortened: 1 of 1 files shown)


heidiguenin:
Thanks, @matt-wirtz  - do you know of a producer or consumer who is interested in publishing/consuming this information?

matt-wirtz:
@heidiguenin Yes, as a consumer (Rhein-Main Transport Association, DE) this info would help streamline manual back office processes and so there is interest in using it.

mplsmitch:
We'd like to find support for this proposal so it could go to a vote - It seems like it would be a good addition to the spec. Are there any producers that would be willing to support this? 
Maybe @sfaubert1 ?

sfaubert1:
It would be nice to have a link to the Terms and Conditions in GBFS indeed. However, in our case, it's usually a link on the bikeshare's website that we point to (not necessarily saved on our backend) and since most websites are updated by the city or the operators directly, we're not sure when it's updated. It would be easy to save and display the URL in GBFS, though. 

I still agree with that change, it could be a nice info to have.

PR #255, opened 2020-07-30
Support for carsharing

josee-sabourin:
Carsharing has become increasingly popular, especially in Europe. With the adoption of dockless bikes and scooters into GBFS, the specification is now able to handle free floating micro mobility services. The fields that exist for dockless micro mobility offer a solid foundation to model carsharing, however they lack the added specificity that a carsharing service would need. This PR aims to model carsharing services by extending `vehicle_types.json`, `station_information.json` and `system_pricing_plans.json` (in addition to already proposed extension of `system_pricing_plans.json` on [PR #252](http://github.com/NABSA/gbfs/pull/252).  A google doc of the proposal can be found [here](https://docs.google.com/document/d/16NKnf10SjmBBVwUlKrc7oeuxzeWqS_dQvztlT2F4dH8/edit?usp=sharing).

Some open questions we'd like feedback on (in addition to feedback on the proposal in general):

- Are the enum values (and their definitions) in `car_type` are representative of the car typology everywhere? If not, how can we make these more all-encompassing
- Should `share_type` be in `vehicle_types.json` or `system_information.json`? The larger question here would be are there services that have more than just one `share_type`?
- Are we missing any propulsion types?
- Should we add license plate information or a passenger facing vehicle name, and if so, how do we mitigate the threat to traveller safety?
- Should we include an `access_methods` field in `station_information.json` that describes how to access a vehicle in instances where payment (i.e. `rental_methods`) and access are completed separately?

Please let us know what you think! We are very excited about the prospect of expanding to support carshare, however as you can see we still have a lot of questions and we would love to hear from you!

PR #257, opened 2020-08-06
adding reservation time info to vehicle_types

matt-wirtz:
There a bike sharing systems which offer a reservation option. Users can reserve a bike upfront for a specific amount of time. The bike stays locked until a rent is started. If a rent is not started within the maximum reservation time the reservation is released. During the reservation time the bike can't be reserved nor rented by other users.
As a consumer we currently have to hard code the information if a vehicle can be reserved or not. So far the bike sharing systems we integrated differentiate the reservation option on the level of bike types not the individual bikes. So I think the information makes most sense on the bike type level.


Empty2k12:
This is useful and something we'd also like to see. I am not sure if this is modelled in the best way, since your solution implies that reservation time is based on the vehicle type. While this would work on our use case, I'm not sure if there're systems where reservation time does not depend on the vehicle type but rather a booked tariff.

josee-sabourin:
In #255, there is a proposed extension to `system_pricing_plans.json` with reservation information, that covers the minimum and maximum times for a reservation as well as the price of reserving. Would this be a possible solution?

matt-wirtz:
@Empty2k12 Good point, the option for reservation might depend on the used tariff, too. And I think it's not an either or but rather that both dependencies (bike_type & tariff) should be possible.
@josee-sabourin Thanks for the hint that the proposed extension #255 might cover this requirement already. 
But actually I'm not sure if the proposed extension is intended for the type of reservation discussed here. It sounds like it covers reservation made for some point in the future and you have to reserve at least for a minimum amount of time. I would rather call it a booking. In contrast to this I was focusing on reservations which only start immediately and do not have any minimum reservation time. One could start the rent immediately after the reservation, there is no minimum reservation time.

Since PR #255 is in discussion too I would rather wait what the outcome is and do not open another PR on this right now. But I will add a hint in the description that reservation options might depend on the used tariff.

matt-wirtz:
I hereby call a vote on this proposal. Voting will be open for 7 full calendar days until 11:59PM UTC on Friday, November 20th.

sfaubert1:
+1 from PBSC

gbfs.md
```diff
@@ -276,6 +276,7 @@ Field Name | Required | Type | Defines
...
 \- `max_range_meters` | Conditionally Required | Non-negative float | If the vehicle has a motor (as indicated by having a value other than `human` in the `propulsion_type` field), this field is required. This represents the furthest distance in meters that the vehicle can travel without recharging or refueling when it has the maximum amount of energy potential (for example, a full battery or full tank of gas).
 \- `name` | Optional | String | The public name of this vehicle type.
+\- `reservation_time_minutes` | Optional | Non-negative integer | Time in minutes the vehicle can be reserved before rental is started. If set to values >0 the vehicle can be reserved for the specified amount of time. The vehicle remains locked until a rental is started. During this time the vehicle can not be reserved or rented by other users. If the reservation time is elapsed and no rental has been started, the reservation is terminated automatically and the vehicle is available for other users. If set to 0 then the vehicle can not be reserved. Note that a given reservation_time might only be one prerequisite and that in addition only specific tariffs might allow a reservation.
```
sfaubert1: There might be a typo here for "reservation_time" in "Note that a given reservation_time" (should be "reservation_time_minutes").

matt-wirtz: Thanks for pointing out. I corrected the typo.

matt-wirtz:
+1 from RMV

josee-sabourin:
The vote has closed with 2 votes in favor. Since 3 are needed in addition to the proposer’s, it does not pass. As per the governance, this does not mean the PR must be abandoned. We'd like to better understand your needs and how to address them going forward. @matt-wirtz & @sfaubert1 - could you please take another look at the proposed reservations model in #255? Then, could you email me at  josee@mobilitydata.org to set up a time to chat so that we can work together towards a solution that brings these two proposals together?

matt-wirtz:
@josee-sabourin Looks like the new change process was applied here when stating that three votes in favor are needed. The vote on the PR was called before the change process was modified. Why wasn't the old process applied?

heidiguenin:
Hi @matt-wirtz - just wanted to clarify. The earlier change process also did call for three votes in favor - with at least one consumer and at least one producer. It was not clear on whether the proposer's vote counted in those three, so that was clarified to make it more in line with GTFS during the recent move from pilot to official governance. But three votes have always been required.

matt-wirtz:
Hi @josee-sabourin. Thanks for the clarification. When calling the vote I think I was looking here to see how the process looks like:
[README.md](https://github.com/NABSA/gbfs/tree/74e41f1af5b75f82623de83e9da382277f038cd3)
There was stated that one GBFS producer is required to implement and test the change. But maybe I was looking at the wrong place.

josee-sabourin:
Hi @matt-wirtz, before the new change process PR we had been testing out a pilot governance (#163) which outlined the need for 3 votes in this [comment](https://github.com/NABSA/gbfs/issues/163#issuecomment-562602035). Hope that helps! I'm happy to chat over e-mail if things are still fuzzy!

matt-wirtz:
@josee-sabourin Ah, a pilot governance with a pilot change process in it I assume. Really confusing to me. Looks like I missed that one out. Thanks for the Info.

mplsmitch:
@matt-wirtz @Empty2k12 We're planning to revisit this and try to pass something. I'm working an a needs assessment and wondering if you know of any examples where there are dependencies related to reservation of vehicles based on tariff or other factors as discussed above.

Issue #250, opened 2020-07-23
aggregate numbers of available vehicle_types at a station


davidlewis-ito:
GBFS v2.1-RC introduces the vehicle_types definition and enables the more specific description of the nature of the available docks at each station as well as details of each specific bike there.

Prior to this extension there have been a number of proprietary extensions to the station_status.json designed to communicate aspects of vehicles types such as:

    "num_bikes_available": 5,
    "num_ebikes_available": 2,
    "num_docks_available": 11,
and

    "num_bikes_available_types": {
      "mechanical": 1,
      "ebike": 0
    },

Whilst the ambition of the GBFS v2.1 to provide per-bike description is very positive, I wanted to explore with the community  whether the aggregate description of bike availability might be worth supporting formally.  Perhaps in the form:

    "vehicle_types_available": [{
      "vehicle_type_id": "abc123"
      "count": 2
     },{
       "vehicle_type_id": "def456",
       "count": 1
     }
    ]
or more concisely:

    "vehicle_types_available": [{
      "abc123": 2
       }, {
      "def456"]: 1
       }]

Without this,  it seems we have a rather all or nothing approach to describing vehicles at a station.

It would be great to get feedback from feed publishers.

kanagy:
For easy comparison, this is how the available vehicles are currently described in station_status (i.e. listing individual vehicles instead of aggregates):
```
       "vehicles": [{
          "bike_id": "stu901",
          "is_reserved": false,
          "is_disabled": false,
          "vehicle_type_id": "abc123"
        }, {
          "bike_id": "vwx234",
          "is_reserved": false,
          "is_disabled": false,
          "vehicle_type_id": "def456",
          "current_range_meters": 4321
        }]
```

Note that station_information describes station capacity as aggregates: 
```
       "vehicle_type_capacity": {
          "abc123": 7,
          "def456": 9
        }
```

I'm not quite sure why available vehicles were not done in the same format as capacity in https://github.com/NABSA/gbfs/pull/136. Perhaps this was to have the reserved and disabled booleans?

Can you explan more what do you mean by "Without this, it seems we have a rather all or nothing approach to describing vehicles at a station.". These additional fields don't seem like a big overhead for producers to describe the vehicles: bike_id is not a key into other feeds (it just needs to be unique) and the two booleans could be set to false when not known.

davidlewis-ito:
[...]

PR #261, opened 2020-08-25
Aggregate available vehicle_types at a station

sfaubert1:
Based on conversations in issue #250 

Modified station_status.json to aggregate the number of available vehicles per vehicle_type and removed the `bike_id`, `is_reserved`, `is_disabled`, `vehicle_type_id` and `current_range_meters` from the feed since they're available in free_bike_status.json. 

Also modified free_bike_status.json to specify the station_id where the vehicle is, in order to provide more details on the vehicle (such as 'current_range_meters'). So the free_bike_status.json could be used for free floating vehicles and docked vehicles as well to provide more information on each of the vehicles.


matt-wirtz:
The proposal sounds logical but includes some fundamental changes to the semantic.

So far the endpoint free_bike_status.json referred to free floating bikes. Bikes that are not docked or locked at a station. So the free in free_bike_status had the meaning of free floating rather then available to rent. But with this proposal the free now means available to rent. So I think this is a breaking change and should be labeled as this. Primarily because available docked bikes will no longer show up at the endpoint station_status.json but rather at free_bile_status.json.

And following this argumentation it would make sense to rename free_bike_status.json into bike_status.json. Bike_status.json would thus become a counterpart to station_status.json.

Although I can follow the basic idea of this PR I don't see a significant advantage for providing or consuming apps of this PR right away? What would be the most important benefits of this breaking PR?

sfaubert1:
@matt-wirtz thanks for you comment. The goal of this PR is to reduce the size of the `station_status.json` feed for large systems, by excluding the bike details and go back to an aggregated number of available vehicles in this feed. See issue #250 for more information.

The number of available docked bikes would still be in `station_status.json` in `vehicle_types_available`. Only the details of each bike that would be removed from `station_status.json` and added in `free_bike_status.json`. It was proposed in issue #250 to refactor the `free_bike_status.json` to include all bikes and their details, but I didn't change the name in this PR thinking it would be a breaking change. I agree that for better comprehension, we could change the name of `free_bike_status.json` to `bike_status.json`.

@davidlewis-ito, @kanagy, @mplsmitch, since involved in the original issue leading to this PR, just thought you would be interested to see the proposed changes.

mplsmitch:
@matt-wirtz The intent of `free_bike_status.json` endpoint has been to describe vehicles that are available for rent, but not at a station.  In other works these are free floating or dockless vehicles that are currently available for rent. This PR maintains the distinction between docked and dockless but consolidates all information relating to individual vehicles in a single file.  We're currently working and a plan to incorporate non-breaking changes from this PR into a minor release with any breaking and additional changes like renaming of `free_bike_status.json` happening in the next major release.

matt-wirtz:
@sfaubert1 Thanks for pointing again to issue #250. I kind of overlooked the reference and directly read through the proposed changes. So by removing the `vehicles` array in the `station_status` endpoint roughly 100 bytes (uncompressed) are saved per bike. With e.g. 5.000 bikes this would roughly be 500kb. 

@mplsmitch Maybe I'm overlooking something but I still don't see how this PR should be non-breaking. Let's say there is a consumer app who uses the information of the `station_information` and `station_status` endpoints to display to the user all bikes available at stations including their e.g. `current_range_meters`. Since this PR removes the `vehicles` array from the endpoint `station_status` the app would not be able the list the individual bikes available without changes to their code. So from my understanding it's a breaking change.

gbfs.md
```diff
@@ -380,12 +380,7 @@ Field Name | Required | Type | Defines
...
-&emsp;\- `is_reserved` <br/>*(added in v2.1-RC)* | Yes | Boolean | Is the vehicle currently reserved for someone else
-&emsp;\- `is_disabled` <br/>*(added in v2.1-RC)* | Yes | Boolean | Is the vehicle currently disabled (broken)
-&emsp;\- `vehicle_type_id` <br/>*(added in v2.1-RC)* | Yes | ID | The vehicle_type_id of this vehicle as described in [vehicle_types.json](#vehicle_typesjson).
-&emsp;\- `current_range_meters` <br/>*(added in v2.1-RC)* | Conditionally Required | Non-negative float | If the corresponding vehicle_type definition for this vehicle has a motor, then this field is required. This value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel.
+\- `vehicle_types_available` <br/>*(added in v2.1-RC)* | Conditionally Required | Object | This field is required if the [vehicle_types.json](#vehicle_typesjson) file has been defined. This field is an object where each key is a `vehicle_type_id` as described in [vehicle_types.json](#vehicle_typesjson) and the value is a number representing the total number of available vehicle at this station for the specified vehicle type.
```
mplsmitch: vehicles should be plural: _available vehicles_

gbfs.md
```diff
@@ -450,15 +431,15 @@ Example:
 
 ### free_bike_status.json
 
-Describes vehicles that are not at a station and are not currently in the middle of an active ride.
+Describes all vehicles that are not at a station and are not currently in the middle of an active ride. From v2.1-RC, also provides more information about vehicles that are at a station if the [vehicle_types.json](#vehicle_typesjson) file has been defined.
```
mplsmitch: Would like to clarify this definition and add line about vehicles in active rental from Best Practices. Suggesting this:

"Describes all vehicles that are not currently in active rental. Required of free floating (dockless) vehicles. Conditionally required of station based (docked) vehicles when [vehicle_types.json](#vehicle_typesjson) has been defined. Vehicles that are part of an active rental must not appear in this feed."

mplsmitch: This definition should also be updated under "Files" above on line 71.

mplsmitch:
Need to remove the reference to station_status.json in the definition of vehicle_types.json under "Files" on line 68

gbfs.md
sfaubert1: now changed, thanks for letting me know

gbfs.md
sfaubert1: Yes that's more clear, thanks! Updated at both places.

sfaubert1:
About previous comment:
> Need to remove the reference to station_status.json in the definition of vehicle_types.json under "Files" on line 68

Please correct me if I'm wrong, but I think the reference to `station_status.json` in the definition of `vehicle_types.json` should remain since there would still be references to the vehicle types in the `vehicle_types_available` field of `station_status.json` (so the `vehicle_types.json` would be required if used in `station_status.json`).

mplsmitch:
@sfaubert1 You are correct - the reference to `station_status.json` should remain in the `vehicle_types.json` definition.

sfaubert1:
I hereby call a vote on this proposal. Voting will be open for 7 full days, until 11:59PM UTC on September 23rd.

sfaubert1:
+1 from PBSC  :) 

and also committing to implement

kanagy:
Nits:
- there's a typo in free_bike_status.json description: "Conditionally required of station based (docked) vehicles when vehicle_types.json has been defined.". of --> for. 
- the indentation is off in some of the examples for some of the added lines.
- Should vehicle_types_available description follow the same style as vehicle_docks_available? I.e. vehicle_docks_available describes in the table the object fields vehicle_type_ids and count. vehicle_types_available should describe those too to be consistent.

Question: Why is docked free_bike_status conditioned on vehicle_types.json being defined? I think free_bikes could describe the vehicles at a station even when the system doesn't define different vehicle_types.

Otherwise, in principle this looks good to me (Google Maps), but we don't plan to implement this soon. In particular I think this is a good direction for supporting hybrid systems where bikes can be both dockless and docked.

davidlewis-ito:
+1 from itoworld

We will be consuming and producing

sfaubert1:
Thanks @mplsmitch for the updates!

For the description of `vehicle_types_available` that was changed to match the format of `vehicle_docks_available`, is there a reason why we should keep an array for `vehicle_type_ids`? 

For us the count of available bike will always be for one type of vehicle only; in the event of a case where multiple vehicle types would have the same count, the meaning could be confusing, for example:
```
        "vehicle_types_available": [{
          "vehicle_type_ids": ["abc123", "xyz789"],
          "count": 2
        }, {
          "vehicle_type_ids": ["def456"],
          "count": 4
        }]
```
In the example above, does it mean there's 2 vehicles of type "abc123" and 2 of type "xyz789" available as well or something else?

Since `vehicle_type_capacity` in `station_information.json` was also an object mapping the vehicle type to its count, we thought it would be easy to understand for `vehicle_types_available`. For example:
```
        "vehicle_types_available": {
          "abc123": 2,
          "xyz789": 2,
          "def456": 4
        }, 
```

If we have to go with an array of objects, maybe consider a single string instead of an array for `vehicle_type_ids` (or `vehicle_type_id` since only one). For example:
```
        "vehicle_types_available": [{
          "vehicle_type_id": "abc123",
          "count": 2
        }, {
          "vehicle_type_id": "xyz789",
          "count": 2
        }, {
          "vehicle_type_ids": "def456",
          "count": 4
        }]
``` 

The advantage of this last example would be if we need to include more properties to the `vehicle_types_available` field in the future, but it would be more heavy for sure.

mplsmitch:
@sfaubert1 Thanks for catching that - the intention was to make it a single string like in your 3rd example as was discussed in #250 . I'll make the change.

yocontra:
+1 from Stae

gbfs.md
```diff
@@ -68,7 +68,7 @@ system_information.json | Yes | Details including system operator, system locati
...
 station_information.json | Conditionally required | List of all stations, their capacities and locations. Required of systems utilizing docks.
 station_status.json | Conditionally required | Number of available vehicles and docks at each station and station availability. Required of systems utilizing docks.
-free_bike_status.json | Conditionally required | Vehicles that are available for rent. Required of systems that offer vehicles for rent outside of stations.
+free_bike_status.json | Conditionally required | Describes all vehicles that are not currently in active rental. Required for free floating (dockless) vehicles. Conditionally required for station based (docked) vehicles when [vehicle_types.json](#vehicle_typesjson) has been defined. Optional for station based (docked) vehicles when [vehicle_types.json](#vehicle_typesjson) has not been defined. Vehicles that are part of an active rental must not appear in this feed.
```
kanagy: Re: "Conditionally required when vehicle_types.json has been defined, optional for when it was not defined."

What does "conditionally required" mean: do you mean "Required when vehicle_types.json has been defined". If that's the case, why is this required when vehicle_types.json were defined? Sorry, I still don't see a relationship there. I would just make this "optional for station based vehicles".

gbfs.md
```diff
@@ -370,22 +370,19 @@ Field Name | Required | Type | Defines
 `stations` | Yes | Array | Array that contains one object per station in the system as defined below.
 \-&nbsp;`station_id` | Yes | ID | Identifier of a station see [station_information.json](#station_informationjson).
 \-&nbsp;`num_bikes_available` | Yes | Non-negative integer | Number of vehicles of any type available for rental. Number of functional vehicles physically at the station. To know if the vehicles are available for rental, see `is_renting`.
+\- `vehicle_types_available` <br/>*(added in v2.1-RC)* | Conditionally Required | Array | This field is required if the [vehicle_types.json](#vehicle_typesjson) file has been defined. This field's value is an array of objects. Each of these objects is used to model the total number of each defined vehicle type available at a station. The total number of vehicles from each of these objects should add up to match the value specified in the `num_bikes_available`  field.
+&emsp;\- `vehicle_type_id` <br/>*(added in v2.1-RC)* | Yes | ID | The `vehicle_type_id` of each vehicle type at the station as described in [vehicle_types.json](#vehicle_typesjson). This field is required if the [vehicle_types.json](#vehicle_typesjson) is defined..
```
kanagy: Nit: two dots (..)

gbfs.md
```diff
@@ -370,22 +370,19 @@ Field Name | Required | Type | Defines
...
+\- `vehicle_docks_available` <br/>*(added in v2.1-RC)* | Conditionally Required | Array | This field is required in feeds where the [vehicle_types.json](#vehicle_typesjson) is defined and where certain docks are only able to accept certain vehicle types. If every dock at the station is able to accept any vehicle type, then this field is not required. This field's value is an array of objects. Each of these objects is used to model the number of docks available for certain vehicle types. The total number of docks from each of these objects should add up to match the value specified in the `num_docks_available` field.
+&emsp;\- `vehicle_type_ids` <br/>*(added in v2.1-RC)* | Yes | Array of Strings | An array of strings where each string represents a vehicle_type_id that is able to use a particular type of dock at the station
+&emsp;\- `count` <br/>*(added in v2.1-RC)* | Yes | Non-negative integer | A number representing the total number of available vehicles of the corresponding vehicle type as defined in the `vehicle_types` array at the station that can accept vehicles of the specified types in the `vehicle_types` array.
```
kanagy, line 381: sfaubert1: FYI I believe the reason this is an array is so you are easily able to count how many docks in total are at the station (i.e. by sumimng up the counts in each entry). However, I would prefer just having a single ID + count (it is clearer and it is consistent with vehicle_types_available definition); perhaps the existing field num_docks_available is sufficient for saying how many available docks are there in total?

gbfs.md
```diff
@@ -370,22 +370,19 @@ Field Name | Required | Type | Defines
...
-&emsp;\- `is_disabled` <br/>*(added in v2.1-RC)* | Yes | Boolean | Is the vehicle currently disabled (broken)
-&emsp;\- `vehicle_type_id` <br/>*(added in v2.1-RC)* | Yes | ID | The vehicle_type_id of this vehicle as described in [vehicle_types.json](#vehicle_typesjson).
-&emsp;\- `current_range_meters` <br/>*(added in v2.1-RC)* | Conditionally Required | Non-negative float | If the corresponding vehicle_type definition for this vehicle has a motor, then this field is required. This value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel.
+\-&nbsp;`last_reported` | Yes | Timestamp | The last time this station reported its status to the operator's backend.\- `vehicle_types_available` <br/>*(added in v2.1-RC)* | Conditionally Required | Object | This field is required if the [vehicle_types.json](#vehicle_typesjson) file has been defined. This field is an object where each key is a `vehicle_type_id` as described in [vehicle_types.json](#vehicle_typesjson) and the value is a number representing the total number of available vehicles at this station for the specified vehicle type.
```
kanagy: vehicle_types_available should not be in this description; it is already described above. Forgot to delete?

gbfs.md
```diff
@@ -450,15 +446,15 @@ Example:
 
 ### free_bike_status.json
 
-Describes vehicles that are not at a station and are not currently in the middle of an active ride.
+Describes all vehicles that are not currently in active rental. Required for free floating (dockless) vehicles. Conditionally required for station based (docked) vehicles when [vehicle_types.json](#vehicle_typesjson) has been defined. Optional for station based (docked) vehicles when [vehicle_types.json](#vehicle_typesjson) has not been defined. Vehicles that are part of an active rental must not appear in this feed.
```
kanagy: Same question here and for station_id: Why are these conditioned on vehicle_types feed being present? I would just say that it is "optional for station based vehicles". I think having vehicle_types feed (i.e. being able to describe what type each vehicle is) is independent from being able to list just list vehicles at a station.

kanagy (commented):
Following up on previous comments and also noticed few typos.

gbfs.md
mplsmitch: I see your point, it doesn't need to be this complicated - I've updated these definitions to make free_bike_status.json optional for station based vehicles.

nighthawk:
+1 from SkedGo

We will be consuming

gbfs.md
sfaubert1: @kanagy no problem with `vehicle_type_ids` being an array for `vehicle_docks_available`, for us it makes perfect sense since a dock can support many different vehicle types. Our issue was this `vehicle_type_ids` field being an array within `vehicle_type_available` but it was updated to an ID so no issue there :)

cmonagle:
+1 from Transit

kanagy:
+1 from Google Maps

heidiguenin:

Votes in favor:
PBSC (producer)
ItoWorld (consumer)
Stae (consumer)
SkedGo (consumer)
Transit (consumer)
Google Maps (consumer)

No votes against.

We'll be merging, creating v2.2RC (likely after the vote on pricing!), and then we'll follow up with those of you who said you'd be implementing. Thanks for all of your help making sure the new dockless support is implementable!

PR #263, opened 2020-09-09
Updates to Chicago, IL, US entries in systems.csv

levyj:


heidiguenin:
Thanks, @levyj. Does that mean that Bolt, HOPR, Lyft, Sherpa/Bird, Veoride, and Wheels are no longer operating in Chicago?

levyj:
Other than Lyft now running Divvy, I think so.

I'm not an expert but ran the pull request past our transportation people before submitting it.
________________________________

heidiguenin:
Great -thanks! Any chance you know if Lyft is publishing GBFS for Divvy?

levyj:
I think we left it in the list. If I recall right, there are now four Chicago feeds but two were unchanged so do not show up in the diff.
________________________________

heidiguenin:
Ah got it - thanks. Missed that Divvy feed when I poked through before.  Thanks for the update to the list!

PR #267, opened 2020-09-28
Added unofficial BiciMAD feed, Madrid

kresp0:


futuretap:
@kresp0 We noticed that the station_status and station_information feeds have invalid JSON. Can you fix the feeds?

kresp0:
> @kresp0 We noticed that the station_status and station_information feeds have invalid JSON. Can you fix the feeds?

Thanks for the heads up! I'll find out what changed, fix the feeds and let you know when done.

kresp0:
@futuretap [Fixed](https://gbfs-validator.netlify.app/?url=https%3A%2F%2Fgbfs.bici.madrid%2Fgbfs.json) and upgraded to GBFS version 2.3

futuretap:
Thanks @kresp0, looks good! There's one issue though: system_information.json states `"discovery_uri":"bicimad://"`. However, the iOS app doesn't support this URL scheme. It would be great if it did (bonus for station-specific deep-links!) because we could launch the app right away instead of falling back to the website.

kresp0:
Agree that it would be nice if it would support it, but unfortunately it seems that the official bicimad app doesn't. 
I see that `discovery_uri` is a mandatory field when using `rental_uris`. Do you have any suggestions?

futuretap:
The easiest suggestion would be to add this to the Info.plist of the iOS app:

```
<key>CFBundleURLSchemes</key>
<array>
	<string>bicimad</string>
</array>

```

That's all what's needed to simply open the app using `bicimad:`. For deep-linking support, code in the app delegate is needed but that's a nice-to-have.

kresp0:
I meant regarding the feed, as I'm not working for bicimad and this GBFS is an unofficial hobby project 😅

futuretap:
Oh I see. Good question, probably I'd resort to stating `https://bicimad.com` as for Android. Do you have a contact to the developers of the app?

kresp0:
I don't have contact with the app developers, sorry.

Knowing that it doesn't work, I've changed `discovery_uri` to also state `https://bicimad.com` for iOS

josee-sabourin:
Hi @kresp0
I see in the feed you published a `feed_contact_email`, this would be the way to contact the developers about this issue. 

Have you tried that?

kresp0:
The email on `feed_contact_email` is mine, as I'm the tech contact for this unofficial GBFS.

Tried finding an email for the operator (bicimad/EMT) to fill the `email` field, but haven't found it.

mplsmitch:
@fbouchPBSC I see Bicimad is running E-Fits now. Do you have any contacts there that you could share with us? Is PBSC planning to publish any Bicimad data?

fbouchPBSC:
Hi Mitch, we are currently in a coexistence period in Madrid where two bike share systems are operating in parallel sharing stations.  We will be providing the official GBFS feed once the transition period is over and only the PBSC system remains.

PR #271, opened 2020-10-12
Incorporate Best Practice Recommendations

mplsmitch:
This proposal incorporates best practice recommendations into the specification. These recommendations have been developed over the past two years based on community feedback.

The objectives of incorporating GBFS Best Practices recommendations into the specification are:
- To improve end-user customer experience in shared mobility apps
- Support broad data interoperability to make it easier for software developers to deploy and scale applications, products, and services
- Facilitate the use of GBFS in various application categories (beyond its original focus on trip planning)

These additions touch many areas of the specification but generally fall into the following three categories:
- Clarifying the intended use of various GBFS endpoints and fields.
- Ensuring uninterrupted access to GBFS data.
- Improving the overall quality and completeness of GBFS feeds.

This proposal is intended to be non-breaking. There will be a second, smaller proposal following this one containing breaking changes.


gbfs.md
```diff
@@ -76,6 +79,19 @@ system_pricing_plans.json | Optional | System pricing scheme.
 system_alerts.json | Optional | Current system alerts.
 geofencing_zones.json <br/>*(added in v2.1-RC)* | Optional | Geofencing zones and their associated rules and attributes.    
 
+## Accessibility 
+Datasets should be published at an easily accessible, public, permanent URL. (e.g., www.agency.org/gbfs/gbfs.json). Ideally, the URL should be directly available without requiring login to access the file to facilitate download by consuming software applications. While it is recommended (and the most common practice) to make a GBFS dataset openly downloadable, if a data provider does need to control access to GBFS for licensing or other reasons, it is recommended to control access to the GBFS dataset using API keys, which will facilitate automatic downloads.  
+
+To be compliant with GBFS, all systems must have an entry in the [systems.csv](https://github.com/NABSA/gbfs/blob/master/systems.csv) file.
```
nighthawk, line 85: This does not apply to private feeds, right?

```suggestion
To be compliant with GBFS, all openly available systems must have an entry in the [systems.csv](https://github.com/NABSA/gbfs/blob/master/systems.csv) file.
```

gbfs.md
```diff
@@ -76,6 +79,19 @@ system_pricing_plans.json | Optional | System pricing scheme.
...
+### Feed Availability 
+Automated tools for application performance monitoring should be used to ensure feed availability.  
+Producers should provide a technical contact who can respond to feed outages in the `feed_contact_email` field in the `system_information.json` file.
+ 
+### Seasonal Shutdowns, Disruptions of Service
+Feeds should continue to be published during seasonal or temporary shutdowns.  Feed URLs should not return a 404.  An empty bikes array should be returned by `free_bike_status.json`. Stations in `station_status.json` should be set to `is_renting:false`, `is_returning:false` and `is_installed:false` where applicable. Seasonal shut down dates should be reflected in `system_calendar.json`.
```
nighthawk:
```suggestion
Feeds should continue to be published during seasonal or temporary shutdowns.  Feed URLs should not return a 404.  An empty bikes array should be returned by `free_bike_status.json`. Stations in `station_status.json` should be set to `is_renting:false`, `is_returning:false` and `is_installed:false` where applicable. Seasonal shutdown dates should be reflected in `system_calendar.json`.
```

gbfs.md
```diff
@@ -84,23 +100,60 @@ geofencing_zones.json <br/>*(added in v2.1-RC)* | Optional | Geofencing zones an
...
+* Files are distributed as individual HTTP endpoints.
+    * Required files must not 404. They must return a properly formatted JSON file as defined in [Output Format](#output-format).
     * Optional files may 404. A 404 of an optional file should not be considered an error.
-* Auto-Discovery:
-    * This specification supports auto-discovery.
-    * The location of the auto-discovery file will be provided in the HTML area of the shared mobility landing page hosted at the URL specified in the URL field of the system_infomation.json file.
-    * This is referenced via a _link_ tag with the following format:
+    
+### Auto-Discovery 
+Publishers should implement auto-discovery of GBFS feeds by linking to the location of the `gbf.json` auto-discovery endpoint.
```
nighthawk:
```suggestion
Publishers should implement auto-discovery of GBFS feeds by linking to the location of the `gbfs.json` auto-discovery endpoint.
```

gbfs.md
```diff
@@ -267,7 +331,7 @@ Field Name | Required | Type | Defines
 
 ### vehicle_types.json *(added in v2.1-RC)*
 
-The following fields are all attributes within the main "data" object for this feed.
+This file should be published by systems offering multiple vehicle types for rental, for example pedal bikes and ebikes. <br>The following fields are all attributes within the main "data" object for this feed.
```
nighthawk:
```suggestion
This file should be published by systems offering multiple vehicle types for rental, for example pedal bikes and ebikes. <br/>The following fields are all attributes within the main "data" object for this feed.
```

gbfs.md
```diff
@@ -313,31 +377,32 @@ Example:
...
+\-&nbsp;`station_area` <br/>*(added in v2.1-RC)* | Optional | GeoJSON Multipolygon | A GeoJASON multipolygon that describes the area of a virtual station. If `station_area` is supplied then the record describes a virtual station. <br /><br /> If lat/lon and `station_area` are both defined, the lat/lon is the significant coordinate of the station (e.g. dock facility or valet drop-off and pick up point).
```
nighthawk:
```suggestion
\-&nbsp;`station_area` <br/>*(added in v2.1-RC)* | Optional | GeoJSON Multipolygon | A GeoJSON multipolygon that describes the area of a virtual station. If `station_area` is supplied then the record describes a virtual station. <br /><br /> If lat/lon and `station_area` are both defined, the lat/lon is the significant coordinate of the station (e.g. dock facility or valet drop-off and pick up point).
```

gbfs.md
```diff
@@ -313,31 +377,32 @@ Example:
...
 \-&nbsp;`capacity` | Optional | Non-negative integer | Number of total docking points installed at this station, both available and unavailable, regardless of what vehicle types are allowed at each dock. Empty indicates unlimited capacity.
-\-&nbsp;`vehicle_capacity` <br/>*(added in v2.1-RC)* | Optional | Object | An object where each key is a vehicle_type_id as described in [vehicle_types.json](#vehicle_typesjson) and the value is a number representing the total number of vehicles of this type that can park within the area defined in the station_area field. If the field station_area is defined and a particular vehicle type id is not defined in this object, then an unlimited virtual capacity is assumed for that vehicle type.
```
nighthawk, line 334: Was `vehicle_capacity` deleted on purpose? Should part of the description be copied to `vehicle_type_capacity`?

nighthawk (commented):
Minor suggestions from proof reading.

gbfs.md
mplsmitch: Good catch - it should not have been deleted.

gbfs.md
mplsmitch: Currently to be fully compliant feeds should be public and have an entry in systems.csv. That hasn't been made as clear as maybe it should have been and we're hoping to clear some of it up with this proposal. We're aware of the need for private feeds and we're planning work on private feeds in the near future. We'll be sure to make it clear what full compliance means in the case of private feeds.

nighthawk: That makes sense. Thanks for clarifying.

gbfs.md
```diff
@@ -267,7 +331,7 @@ Field Name | Required | Type | Defines
 
 ### vehicle_types.json *(added in v2.1-RC)*
 
-The following fields are all attributes within the main "data" object for this feed.
+This file should be published by systems offering multiple vehicle types for rental, for example pedal bikes and ebikes. <br/>The following fields are all attributes within the main "data" object for this feed.
```
evansiroky: I think it'd be good to add stronger wording here to reflect the requirement of this file in these types of systems. Maybe something like:

> This file **must** be published...

gbfs.md
```diff
@@ -313,31 +377,31 @@ Example:
...
+\-&nbsp;`station_area` <br/>*(added in v2.1-RC)* | Optional | GeoJSON Multipolygon | A GeoJSON multipolygon that describes the area of a virtual station. If `station_area` is supplied then the record describes a virtual station. <br /><br /> If lat/lon and `station_area` are both defined, the lat/lon is the significant coordinate of the station (e.g. dock facility or valet drop-off and pick up point).
```
evansiroky: Could this PR be a good opportunity to follow up on https://github.com/NABSA/gbfs/pull/269#issuecomment-708026313 and add clarifying wording in here saying that the station_area takes precedence over any ride_allowed rules in overlapping geofencing_zones?

gbfs.md
```diff
@@ -364,26 +428,29 @@ Example:
...
-\-&nbsp;`last_reported` | Yes | Timestamp | The last time this station reported its status to the operator's backend. 
+&emsp;\- `count` <br/>*(added in v2.1-RC)* | Yes | Non-negative integer | A number representing the total number of available docks at the station that can accept vehicles of the specified types in the `vehicle_types` array.
+\- `vehicles` <br/>*(added in v2.1-RC)* | Conditionally Required | Array | This field is required if the [vehicle_types.json](#vehicle_typesjson-added-in-v21-rc) file has been defined. This field's value is an array of objects. Each object contains data about a specific vehicle that is currently present at the docking station. Each of these vehicles is assumed to be rentable unless otherwise indicated with the is_reserved or is_disabled flags. All of the remaining fields in this table represent key/values for each vehicle object in this array. The length of this array must equal the value of the `num_bikes_available` field.
```
evansiroky: This was removed in https://github.com/NABSA/gbfs/pull/261. Seems like this PR needs to be updated to reflect those changes.

gbfs.md
```diff
@@ -84,23 +100,60 @@ geofencing_zones.json <br/>*(added in v2.1-RC)* | Optional | Geofencing zones an
...
+   * Examples:
+     * Central Park South
+     * Villiers-sur-Marne
+     * Market Street
+
+Abbreviations should not be used for names and other text (e.g. St. for Street) unless a location is called by its abbreviated name (e.g. “JFK Airport”). Abbreviations may be problematic for accessibility by screen reader software and voice user interfaces. Consuming software can be engineered to reliably convert full words to abbreviations for display, but converting from abbreviations to full words is prone to more risk of error.
+
+Names used for stations, virtual stations and geofenced areas should be human readable. Naming conventions used for locations should consider a variety of use cases including both text and maps. 
+
+Descriptions should not include information so specific that it could be used in tracking of vehicles or trips.
```
kanagy, line 138: Can you provide an example of how that could happen?

gbfs.md
```diff
@@ -84,23 +100,60 @@ geofencing_zones.json <br/>*(added in v2.1-RC)* | Optional | Geofencing zones an
...
+---|---|---
+0|1.0|111 km
+1|0.1|11.1 km
+2|0.01|1.11 km
+3|0.001|111 m
+4|0.0001|11.1 m
+5|0.00001|1.11 m
+**6**|**0.000001**|**0.111 m**
+7|0.0000001|1.11 cm
+### Data Latentcy
```
kanagy, line 153: Nit:typo

gbfs.md
```diff
@@ -84,23 +100,60 @@ geofencing_zones.json <br/>*(added in v2.1-RC)* | Optional | Geofencing zones an
...
+0|1.0|111 km
+1|0.1|11.1 km
+2|0.01|1.11 km
+3|0.001|111 m
+4|0.0001|11.1 m
+5|0.00001|1.11 m
+**6**|**0.000001**|**0.111 m**
+7|0.0000001|1.11 cm
+### Data Latentcy 
+The data returned by the near-realtime endpoints `station_status.json` and `free_bike_status.json` should be as close to realtime as possible, but in no case should it be more than 5 minutes out-of-date.  Appropriate values should be set using the `ttl` property for each endpoint based on how often the data in feeds are refreshed or updated.
```
kanagy: Should this also say something about appropriately setting last_updated and last_reported timestamps to reflect the staleness of data?

gbfs.md
```diff
@@ -364,26 +428,29 @@ Example:
...
-\- `vehicle_docks_available` <br/>*(added in v2.1-RC)* | Conditionally Required | Array | This field is required in feeds where the [vehicle_types.json](#vehicle_typesjson) is defined and where certain docks are only able to accept certain vehicle types. If every dock at the station is able to accept any vehicle type, then this field is not required. This field's value is an array of objects. Each of these objects is used to model the number of docks available for certain vehicle types. The total number of docks from each of these objects should add up to match the value specified in the `num_docks_available` field.
+\-&nbsp;`station_id` | Yes | ID | Identifier of a station. See [station_information.json](#station_informationjson).
+\-&nbsp;`num_bikes_available` | Yes | Non-negative integer | The total number of vehicles of all types that are currently available and offered for rental at the station. If the `is_renting` boolean is set to`false` this number should be zero.
```
kanagy: Should this be conditionally required for valet stations, similar to num_docks_available?

gbfs.md
```diff
@@ -364,26 +428,29 @@ Example:
...
+\-&nbsp;`is_returning` | Yes | Boolean | Is the station currently accepting vehicle returns? <br /><br />`true` - Station is currently accepting vehicle returns. Even if the station is full, if it would otherwise allow vehicle returns this value must be `true`.<br /> `false` - Station is not accepting vehicle returns. <br><br>If the station is temporarily taken out of service and not allowing vehicle returns this field must be set to `false`.<br><br>If a station becomes inaccessible to users due to road construction or other factors this field should be set to `false`.
+\-&nbsp;`last_reported` | Yes | Timestamp | The last time this station reported its status to the operator's backend.
```
kanagy: Should this clarify that this is also the timestamp at which all the station data in this feed were recorded?

gbfs.md
```diff
@@ -447,17 +514,16 @@ Example:
...
+\-&nbsp;`bike_id` | Yes | ID | Identifier of a vehicle. The `bike_id` identifier must be rotated to a random string after each trip to protect user privacy *(as of v2.0)*. Use of persistent vehicle IDs poses a threat to user privacy. The `bike_id` identifier should only be rotated once per trip.
 \-&nbsp;`system_id` <br/>*(added in v3.0-RC)* | Conditionally required | ID | Identifier referencing the system_id field in system_information.json. Required in the case of feeds that specify free (undocked) bikes and define systems in system_information.json.
-\-&nbsp;`lat` | Conditionally required <br/>*(as of v2.1-RC2)* | Latitude | Latitude of the vehicle. *(as of v2.1-RC2)* This field is required if station_id is not provided for this vehicle (free floating).
-\-&nbsp;`lon` | Conditionally required <br/>*(as of v2.1-RC2)* | Longitude | Longitude of the vehicle. *(as of v2.1-RC2)* This field is required if station_id is not provided for this vehicle (free floating).
```
kanagy, line 458: You dropped longitude from the fields.

gbfs.md
mplsmitch: That's a good question, what I don't understand is if there were a valet station with an unlimited supply of vehicles, how would the operator indicate that there are any bikes available without using this field? In practice I don't think it matters that much because where i've seen valet service, the operator uses the station to initiate the rentals so it's constantly being refilled to maintain the value in `num_bikes_available`

gbfs.md
mplsmitch: Fixed

gbfs.md
mplsmitch: `last_reported` in this context represents the time when each individual station in the array last updated its status with the  provider's backend so it's nor when all the station data in the feed was recorded. If a feed contained 20 stations, there could be 20 different `last_reported` values.

gbfs.md
mplsmitch: I've updated it to reflect this

gbfs.md
mplsmitch: Fixed

gbfs.md
mplsmitch: Fixed

gbfs.md
mplsmitch: this has been added

gbfs.md
mplsmitch: I believe this originally came up in reference to car sharing. I know there are bike systems where each individual bike has a unique name which becomes the equivalent of a bike ID that would allow for the recreation of trips. In car sharing it would be using
`"vehicle_desc": "Four-door black sedan" `  
in stead of
 `"vehicle_desc": "Four-door black sedan #3" `

gbfs.md
```diff
@@ -379,13 +443,12 @@ Field Name | Required | Type | Defines
...
+\-&nbsp;`is_renting` | Yes | Boolean | Is the station currently renting vehicles? <br /><br />`true` - Station is renting vehicles. Even if the station is empty, if it would otherwise allow rentals this value must be `true`.<br/>`false` - Station is not renting vehicles.<br/><br/>If the station is temporarily taken out of service and not allowing rentals this field must be set to `false`.<br/><br/>If a station becomes inaccessible to users due to road construction or other factors this field should be set to `false`.
+\-&nbsp;`is_returning` | Yes | Boolean | Is the station accepting vehicle returns? <br /><br />`true` - Station is accepting vehicle returns. IEven if the station is full, if it would otherwise allow vehicle returns this value must be `true`.<br /> `false` - Station is not accepting vehicle returns.<br/><br/>If the station is temporarily taken out of service and not allowing vehicle returns this field must be set to `false`.<br/><br/>If a station becomes inaccessible to users due to road construction or other factors this field should be set to `false`.
```
evansiroky: typo: `IEven`

gbfs.md
mplsmitch: @evansiroky I'm reluctant to make `vehicle_types.json` a requirement when this is intended to be a non-breaking PR. While requiring it may not break down stream apps it seems like a significant change for producers. Currently it's conditionally required of systems who publish vehicle type data in `free_bike_status.json`. I think we should revisit how it's defined in v3.0.

mplsmitch:
I hereby call a vote on this proposal. Voting will be open for 7 full calendar days until 11:59PM UTC on Monday, November 9th.

gbfs.md
evansiroky: I'm confused. In the [Files](https://github.com/NABSA/gbfs/blob/best-practices-nb/gbfs.md#files) section, it says that this file is conditionally required already. Seems like that same language ought to be used here to be consistent.

Screen Shot 2020-11-02 at 10 10 00 AM

mplsmitch: I've added the language from the Files section-

As it's currently defined under Files,  systems publishing vehicle type information in `free_bike_status` MUST publish `vehicle_types.json`. If they are not including vehicle type information in `free_bike_status` they would be exempted from that requirement regardless of the types of vehicles offered.  This says that for those that operate multiple types of vehicles, publishing vehicle_types.json is something they SHOULD do.  If that's not adequate, we could redefine `vehicle_types.json` in v3.0 to require it of systems operating multiple types of vehicles

gbfs.md
```diff
@@ -501,7 +564,7 @@ Example:
 ```
 
 ### system_hours.json
-Describes the system hours of operation.
+This optional file is used to describe hours and days of operation when vehicles are available for rental. If `system_hours.json` is not published it indicates that vehicles are available for rental 24 hours a day, 7 days a week.
```
kanagy, line 567: I think more clarification is needed that: if a system is outside of hours of operation, it should have no vehicles available in the realtime feeds (i.e. it should set num_bikes_available=0 and have no vehicles in free_bike_status.json). I'm not sure that is mentioned anywhere. The last sentence here makes it sound like consumers need to ingest system_hours just to determine if current real-time availability is outside of hours of operation.

kanagy (commented):
Would you be able to also update the system_pricing_plans documentation to clarify start/end interval points in pricing? When I looked at it recently it wasn't clear whether the start/end points in pricing are exclusive or inclusive, and exactly at which point the price increases. 

E.g. the start documentation says "Number of kilometers that have to elapse before this segment starts applying." Does this mean that if start=1 and trip takes 1 min, we don't charge the user, but if the trip is taking 1.01 min we would charge the user? Similar confusion for the end field.

gbfs.md
mplsmitch: > if a system is outside of hours of operation, it should have no vehicles available in the realtime feeds (i.e. it should set num_bikes_available=0 and have no vehicles in free_bike_status.json).

The definition of `num_bikes_available` was changed (discussion in #94, PR #195 )to say just the opposite of this. During off hours `num_bikes_available` now remains constant so someone trying to plan a trip at 6:45 can see how many vehicles are going to be available at 7:00 when the system opens. Same should be true for `free_bike_status`. In a station based system the operator could set `is_renting` to `false` in `station_status` during the time when the system is closed  but there's no equivalent in `free_bike_status`  so you would need to ingest `system_hours`.

mplsmitch: > Would you be able to also update the system_pricing_plans documentation to clarify start/end interval points in pricing? 

`start` and `end` definitions have been updated in pricing

gbfs.md
```diff
@@ -579,15 +642,15 @@ Field Name | Required | Type | Defines
...
 &emsp;&emsp;\-&nbsp;`rate` <br/>*(added in v2.1-RC2)* | Yes | Float | Rate that is charged for each kilometer `interval` after the `start`. Can be a negative number, which indicates that the traveller will receive a discount.
 &emsp;&emsp;\-&nbsp;`interval` <br/>*(added in v2.1-RC2)* | Yes | Non-Negative Integer | Interval in kilometers at which the `rate` of this segment is either reapplied indefinitely, or if defined, up until (but not including) `end` kilometer.<br /><br />An interval of 0 indicates the rate is only charged once.
-&emsp;&emsp;\-&nbsp; `end` <br/>*(added in v2.1-RC2)* | Optional | Non-Negative Integer | The kilometer at which the rate will no longer apply.<br /><br /> If this field is empty, the price issued for this segment is charged until the trip ends, in addition to following segments.
+&emsp;&emsp;\-&nbsp; `end` <br/>*(added in v2.1-RC2)* | Optional | Non-Negative Integer | The kilometer at which the rate will no longer apply *(exclusive)* e.g. if `end` is `20` the rate no longer applies at 20.00 km.<br /><br /> If this field is empty, the price issued for this segment is charged until the trip ends, in addition to following segments.
```
kanagy, line 648: Sorry, this still doesn't make the rate function clear to me. It seems neither the start or the end minute/kilometer is included in the range. Assume per_min_pricing:
{start: 1, end: 3, rate: 1}, {start: 3, end: 4, rate: 0.5}

Is this how much trips would cost at these times:
- 1 min: 0
- 1 min 1 second to 2 min: 1
- 2 min: 1
- 2 min 1 second to 3 min: 2
- 3 min: 2
- 3 min 1 second: 2.5
?

gbfs.md
kanagy: I see. I don't like that the realtime part of the spec does not reflect the realtime rental availability and consumers have to double-check with another mostly static feed, but I do respect conversation in #94. Would it make sense to add guidance to is_renting on whether this should be set to false outside of hours of operation? Currently the description seems to point towards it: "temporarily taken out of service".

gbfs.md
mplsmitch: Because `start` is inclusive, your example {start: 1, end: 3, rate: 1}, {start: 3, end: 4, rate: 0.5}  assuming `interval` is 1 would result in:

0-59 sec: 0
1 min: 1
1 min 1 second to 1 min 59 sec: 1
2 min: 2
2 min 1 second to 2 min 59 sec: 2
3 min: 3
3 min 1 second to 3 min 59 sec: 3.5
4 min 3.5

To set per min pricing you would use { start: 0, interval: 1,  rate: 1 } which would result in:
0-59 sec: 1
1 min  to 1 min 59 sec: 2
2 min to 2 min 59 sec: 3

mplsmitch: > Would it make sense to add guidance to is_renting on whether this should be set to false outside of hours of operation? 

added  "Field should be set to `false` during hours or days when the system is not offering vehicles for rent"

kanagy: The description for start says: "kilometers *(inclusive)* that have to elapse before this segment starts applying". Note the "before" and "elapse" bits. 

I.e. if start=1, this reads to me that 1 kilometer (inclusive) has to elapse before the segment starts applying. I.e. we have to get past the 1km mark. If it said exclusive, it would make more sense to me. But I would suggest chancing this sentence completely.  How about: "The kilometer at which this segment rate starts being charged (inclusive)."

kanagy:
Thanks for documenting the best practices. Voting +1 from Google Maps

cmonagle:
+1 on behalf of Transit.

nighthawk:
+1 from SkedGo.

evansiroky:
+1 from IBI Group

mplsmitch:
This vote has been extended until **11:59PM UTC on Thursday, November 12th**. Please weigh in for or against the proposal, and include the organization for which you are voting in your comment.

jayjonsson:
+1 from Lime

bhandzo:
+1 from Bird.

sfaubert1:
+1 from PBSC

mplsmitch:
This vote is now closed - and it passes

Votes in favor:
Google Maps (consumer)
Transit (consumer)
IBI  (consumer)
Lime  (producer)
Bird  (producer)
PBSC  (producer)
SkedGo (consumer)

There we no votes against

PR #278, opened 2020-12-10
Update Beryl feeds in systems.csv

daaain:
We (at Beryl) finally got around to move the GBFS feeds to HTTPS URLs and to a subdomain on our main domain. Oh, and also upgraded the feed specs to v2 🎉 

There's also a new scheme on Isle of Wight which is now included.


mplsmitch:
Nice work @daaain  - thanks for the update and welcome to V2.1!

PR #303, opened 2021-03-31
Fix system_pricing_plans examples and internal links 

nbdh:
Hi there,

we at nextbike recently started implementing GBFS version 2.2 to finally replace our old v1 feeds. Exciting to see how alive the community is!

While working at it I found two minor issues:
- three internal links to the `vehicle_types.json` chapter were broken when `2.1-RC` became `2.1`.
- The examples for the `system_pricing_plans.json` were incorrect.
The documentation states that `plans` shall be an "array of objects", yet it was only an object in the two examples.

I've used the opportunity to slightly reformat them and put spaces after the colons after the key. I found that in gbfs.md both ways (with and without space) are used and evenly common, so I went with what I find more readable. It could be considered to specify this somewhere and unify existing examples.

I think it also should be considered to incorporate those changes into `v2.2`.

- [ ] Yes 
- [x] No
- [ ] Unsure

gbfs.md


mplsmitch:
These changes have been incorporated into v2.2. Thanks @nbdh !

PR #305, opened 2021-03-31
Update nextbike systems and adjust url to auto-discovery file for GBFS v2.2

nbdh:
- [ ] Yes 
- [x] No
- [ ] Unsure

systems.csv

### DIFF
--- a/systems.csv
+++ b/systems.csv
@@ -165,7 +165,6 @@ FI,Donkey Republic Kouvola,Kouvola,donkey_kouvola,https://www.donkey.bike/cities
 FI,Donkey Republic Lappeenranta,Lappeenranta,donkey_lappeenranta,https://www.donkey.bike/cities/bike-rental-lappeenranta/,https://stables.donkey.bike/api/public/gbfs/donkey_lappeenranta/gbfs.json
 FI,Donkey Republic Mäntsälä,Mäntsälä,donkey_maentsaelae,https://www.donkey.bike/cities/bike-rental-mantsala/,https://stables.donkey.bike/api/public/gbfs/donkey_maentsaelae/gbfs.json
 FI,Donkey Republic Porvoo,Porvoo,donkey_porvoo,https://www.donkey.bike/cities/bike-rental-porvoo/,https://stables.donkey.bike/api/public/gbfs/donkey_porvoo/gbfs.json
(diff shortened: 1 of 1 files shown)


systems.csv
```diff
@@ -164,15 +165,15 @@ FI,Donkey Republic Kouvola,Kouvola,donkey_kouvola,https://www.donkey.bike/cities
 FI,Donkey Republic Lappeenranta,Lappeenranta,donkey_lappeenranta,https://www.donkey.bike/cities/bike-rental-lappeenranta/,https://stables.donkey.bike/api/public/gbfs/donkey_lappeenranta/gbfs.json
 FI,Donkey Republic Mäntsälä,Mäntsälä,donkey_maentsaelae,https://www.donkey.bike/cities/bike-rental-mantsala/,https://stables.donkey.bike/api/public/gbfs/donkey_maentsaelae/gbfs.json
 FI,Donkey Republic Porvoo,Porvoo,donkey_porvoo,https://www.donkey.bike/cities/bike-rental-porvoo/,https://stables.donkey.bike/api/public/gbfs/donkey_porvoo/gbfs.json
-FI,Oulu Finland,"Oulu, FI",nextbike_fo,https://kaupunkipyorat.ouka.fi/,https://gbfs.nextbike.net/maps/gbfs/v1/nextbike_fo/gbfs.json
+FI,Oulu Poland,"Oulu, FI",nextbike_fo,https://kaupunkipyorat.ouka.fi/,https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_fo/gbfs.json
```
mplsmitch: Shouldn't this be Oulu Finland, not Poland?

merjakaj: Oulu indeed is located in Finland. The city bike system in Oulu was not functional during 2020 and appears not to be starting this year. See the press release from city of Oulu (in Finnish):

https://www.ouka.fi/oulu/ajankohtaista/uutiset-ja-tiedotteet/-/asset_publisher/s8Z1/content/oulun-kaupunkipyorajarjestelma-ei-kaynnisty-toukokuussa/50266

nbdh: Good catch. 

As this is the "name of the mobility system", which "MUST match the name field in system_information.json" I can not simply rename it here. At least the location is correctly set ;-) 

Even though we are aware that Oulu is not in Poland, I can't rename the system without checking back with my colleagues either. As you can read in the article that @merjakaj shared, this system is operated by Nextbike Polska, so I guess that's how "Poland" got mixed into the name.

I'd just remove this line for now and discuss it internally, perhaps it the system will be (properly named) listed again with our next update of the systems.csv.

```suggestion

```

mplsmitch:
Terrific - Thanks!

PR #316, opened 2021-04-21
gbfs.md formatting and consitency changes

petoc:
### **What problem does your proposal solve?**

- increase readability of document
- increase readability of document source during editing
- increase consistency across document
- conform to Style Guide and GBFS specification

### **What is the proposal?**

Updated document source, to follow Style Guide:
- separated title sections with new line (not directly in Style Guide, but to increases readability during editing)
- replaced tabs with spaces
- removed unnecessary spaces at line endings and from empty lines

Examples updated to follow GBFS specification and updated formatting:
- unified formatting
- unified parameter values
- link and email domains changed to `example.com`
- added required parameters for feeds
- truncated coordinate precision to 6 decimals (as suggested by specification)
- changed ID type from `integer` to `string` at one place
- removed ellipsis from some examples to make json valid

Protocol of all external links changed to `https`, if target supports it.

I believe also replacing `&nbsp;` elements with space, between dash and parameter name, could clean source a little bit, but it would make PR less readable, so left if for now.

- [x] No

gbfs.md

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -1,13 +1,17 @@
 
 # General Bikeshare Feed Specification (GBFS)
+
 This document explains the types of files and data that comprise the General Bikeshare Feed Specification (GBFS) and defines the fields used in all of those files.
 
 # Reference version
-This documentation refers to **v3.0-RC (release candidate)**. <br>
+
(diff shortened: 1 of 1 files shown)


mplsmitch:
Thank you @petoc - terrific work!

Issue #298, opened 2021-03-23
Pricing plans valid time and week day, add services file


cizmarf:
### **Who I am**

I am working on the project Intermodal Routing Machine for the city of Prague. Currently, I create common data format holding the data of all share mobility providers which should be consume by this routing machine.

### **What is the issue and _why_ is it an issue?**

There are missing important attributes in the pricing plans file or an extra file needs to be created.

I found the share mobility providers who are using different pricing plans valid in chosen day time (morning and night prices may differ) and valid in a week day (special weekend sales) and specials pricing plans.

Examples: 
https://anytimecar.cz/rates.html 
https://www.rekola.cz/en/pricing

### **Potential solutions I have considered.**

The simplest possible solution is to add those attributes to the pricing plan file.

start_time_of_period: "Mon-Fri 09:00AM", 
end_time_of_period: "Mon-Fri 09:00PM",
pricing_plan_description: "member card holders"

To solve this problem in general we would need much more attributes to cover all business pricing plans. I would consider creating an extra file describing the services provided by a vehicle. It means if a bike rental company would decide to make specials for weekend users or provide a student sale or a long-time rental sale we can cover it in this service file. This solution would add one or more services to a vehicle and one or more pricing plans to a service.

The reason for creating an extra file and do not add all the attributes to the pricing plans file is keeping separated the business layer and the simple pricing plans. In other words from the point of view of a customer, the customer would be able to choose the service he/she wants to use (I am student, I am a member card holder) but the pricing plans are connected to the each of those services and cannot be impacted by the customer decision. It is possible to implements all the changes into the pricing plans file but it would break its format by a lot.

### **Is your potential solution a breaking change?**

I do not need to implement those changes in the next version of GBFS format. I would like to discuss the suggestions with you, if it is the good way to develop and extend the GBFS if I base my own format on GBSF or if you plan to implement those changes in the future. Or we can consider any other suggestions how to cover the described problems.

Thank you.

- [x] Yes 
- [ ] No
- [ ] Unsure

josee-sabourin:
Hi there! Thanks for bringing up this issue and welcome to the repository! 

[...]

Issue #98, opened 2018-05-06
Different rental hours per station


skinkie:
At this moment different rental hours per station can only be modeled if every station is modeled as a different system. Wouldn't it be more elegant to have the ability to set the rental hours per station opposed to per system?

sven4all:
This is exactly what I want to propose next :-) . In the Netherlands we do have manned stations as well, therefor this would be a nice addition.

jcn:
I guess the real question is whether this is the kind of information that needs to be modeled in GBFS, or whether it's something that should just be implemented on the backend and updates the actual rental status of a station as needed.

i.e. is this information that is needed ahead of time ("this station will become a manned station at 9am") or is it more important in the moment ("this station is currently online, regardless of whether it will be offline in 20 minutes")?

sven4all:
If you know it ahead of time, what is the reason to communicate it? The actual rental status is already part of the specification, right?

Situations where this information can be handy is for trip planning (planning a trip 2 hours in the future for example, or the rental station is not opened right now but after having a train ride the rental station is already opened) , also it's useful information to see until what time you can return your bicycle.

mplsmitch:
This sounds a similar to an old discussion we had in #16 regarding additional station data and services. I'm generally supportive although I would like to see a vendor present a specific use case so we can avoid making speculative changes.

skinkie:
@mplsmitch what is it then what you want to see?

The specific vendor OVfiets (NL) has the following operational rules:
 - there are manned an unmanned stations
 - each stations has another operational availability (per station is not supported in GBFS)
 - the availability has calendar exceptions (not supported at all in GBFS)
 - bicycles brought back to a different station have an increased rental fee

mplsmitch:
I'm sure we can identify many vendors with operational rules that aren't currently supported. The question is, is there a vendor who is publishing GBFS who has identified a need for additional features, or are we speculating that if we model their operational rules they will use the new features? Ideally we would get a proposal from OVfiets.

skinkie:
Both @sven4all and me are creating a GBFS feed for OVfiets and others in The Netherlands, and have difficulties to model it. http://gbfs.openov.nl/

mplsmitch:
[...]

PR #328, opened 2021-06-01
Change to OSM opening_hours format for times and dates

mplsmitch:
Adds fields to system_information and station_information to describe times and dates of operation using OSM opening_hours format. Deprecates  system_hours and system_calendar endpoints.

Currently GBFS sues a proprietary format to describe hours and dates of service. This format lacks flexibility to address new use cases such as setting hours for individual stations. 

This proposal changes the format used to describe hours and dates of operation from the current GBFS proprietary formats to Open Street Map’s [opening_hours](https://wiki.openstreetmap.org/wiki/Key:opening_hours) format.  Changing to this format adds additional functionality and is supported by existing code libraries and tools. This change will allow support for calendar exceptions such as holidays and allows operators to set different operating hours per station. 

This change was initially discussed in #6 . A detailed writeup of this proposal is contained in [this document](https://docs.google.com/document/d/1FN_7FnyqqK7g6unN70fqlMMhE3THIoM0jN_AlmSR4VY/edit?usp=sharing). The need for different hours per station was covered in #98 . This change may also facilitate a future proposal to set valid times and dates for pricing plans as discussed in #298 .

- [x] Yes 
- [ ] No
- [ ] Unsure

gbfs.md:

- system_information.json
- station_information.json


kanagy:
Thanks! The spec looks neater having system_hours and system_calendar endpoints removed. Questions:
- Is the OSM format possibly protected by a license that could limit it's use?
- Would it be useful to provide some guidance in this PR for what real-time systems should set in GBFS data when they are outside of opening hours? E.g. we've seen different systems use different mechanisms: setting is_disabled, is_renting, is_returning, setting the number of bikes/docks to 0 or just straight up shutting down the feed for winter.

mplsmitch:
We're planning to call a vote on this proposal soon and we would like to understand the level of support for this change. There are other proposals coming up that depend on the passing of this PR. This is a breaking change that would become part of v3.0.

Please let us know the following:
- Are you willing to vote in support of this proposal as it's currently written?
- If not, are there any changes that you would require in order to support the proposal?


 I'm tagging those who have expressed interest in this or related issues. 
CC: @contra @brookemckim @evansiroky @skinkie @sven4all @cizmarf @cmonagle

yocontra:
@mplsmitch This looks great to me!

cmonagle:
We'd support this change. Thanks @mplsmitch

evansiroky:
IBI Group is supportive of this change.

gbfs.md
```diff
@@ -369,6 +378,7 @@ Field Name | REQUIRED | Type | Defines
     "phone_number": "1-800-555-1234",
     "name": "Example Bike Rental",
     "operator": "Example Sharing, Inc",
+    "opening_hours":"Apr 1 - Nov 3 00:00-24:00"
```
nbdh:
```suggestion
    "opening_hours": "Apr 1 - Nov 3 00:00-24:00"
```

gbfs.md
```diff
@@ -472,6 +483,7 @@ Field Name | REQUIRED | Type | Defines
         "name": "Parking garage A",
         "lat": 12.345678,
         "lon": 45.678901,
+        "station_opening_hours": Su-Th 05:00-22:00; Fr-Sa 05:00-01:00
```
nbdh:
```suggestion
        "station_opening_hours": "Su-Th 05:00-22:00; Fr-Sa 05:00-01:00",
```

nbdh (changes requested):
Just some typos/formatting in the JSON examples.

nbdh:
Following up on @josee-sabourin mail here.

Though we are publishing the existing feeds for all our systems, we only have two with limited opening hours, so practical usage is rather low.

About a year ago I spent a quite some time to implement some kind of conversion from OSM opening hours format to GBFS format, as we started to use (a subset of the) OSM format internally for defining the opening hours. (As a producer) we'll happily throw that away again and just go with publishing what we have anyways - so I guess **we support the proposal** in general.

I can't evaluate how this consumers will like this change. The OSM format allows some pretty tricky stuff (well, reality is complex). On the other hand, there are some libraries available e.g. to clarify "is it open right now?", but the current format definitely answers this question more easily without using 3rd party library. Answering this question for the OSM format eventually requires (depending on the used features) additional context, e.g. coordinates, country and state, for example to evaluate whether or not today is a public holiday or when sunrise/sunset is.

This change also means that validating a GBFS feed in theory also means to validate the opening hours.

gbfs.md
```diff
@@ -369,6 +378,7 @@ Field Name | REQUIRED | Type | Defines
     "phone_number": "1-800-555-1234",
     "name": "Example Bike Rental",
     "operator": "Example Sharing, Inc",
+    "opening_hours": "Apr 1 - Nov 3 00:00-24:00"
```
nbdh: Sorry, the missing space in this line must have distracted me from the missing comma at the end yesterday
```suggestion
    "opening_hours": "Apr 1 - Nov 3 00:00-24:00",
```

mplsmitch:

Voting will be open for 10 full calendar days until 11:59PM UTC on December 9, 2021.

skinkie:
This pull requests mixes 'the proposal' and 'formatting fixes'. I would suggest not to do so.

cmonagle:
+1 from Transit

gbfs.md
```diff
@@ -101,9 +101,17 @@ Producers SHOULD provide a technical contact who can respond to feed outages in
...
+###  Hours and Dates of Operation
+Beginning with v3.0, hours and dates of operation are described using the Open Street Map [opening_hours](https://wiki.openstreetmap.org/wiki/Key:opening_hours) format. The OSM opening_hours syntax is quite complex, therefore it is RECOMMENDED that publishers validate their opening_hours data to ensure its accuaracy.
+* [OSM opening_hours examples](https://wiki.openstreetmap.org/wiki/Key:opening_hours)
+* [OSM opening_hours syntax guide](https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification)
+* [OSM opening_hours validation tool](https://openingh.openstreetmap.de/evaluation_tool/)
+* [OSM opening_hours project and code libraries](https://github.com/opening-hours)
+
+Hours and dates of operation SHOULD be published even in cases where services are continuously available 24/7. During periods when a system or station is outside of opening hours, stations SHOULD be set to `is_renting = false`. During these periods, `station_status.json.num_bikes_available` and `station_status.json.num_docks_available` SHOULD reflect the number of vehicles and docks that would be available if the system or station were open.
```
kanagy: Hi, could this clarify how free_bike_status.json should behave wrt. opening hours since that feed doesn't have is_renting field? I.e. if a system is outside of opening hours, would there be no free bikes? Also, in this paragraph stations SHOULD be set to `is_returning=false` too outside of opening hours.

We're trying to figure out whether we MUST check system_information.opening_hours when ingesting real-time feeds before surfacing bikes (i.e. this is a breaking proposal) or we can continue gating by is_renting=false, is_returning=false and no bikes in free_bike_status.

mplsmitch: if a system is outside of opening hours, there would not be any vehicles available for rent, including free floating bikes. This can be added to the system_information.opening_hours definition.

> Also, in this paragraph stations SHOULD be set to is_returning=false too outside of opening hours.

Actually this is not the case - systems typically allow you to end a rental and/or return vehicles outside of opening hours.  

> We're trying to figure out whether we MUST check system_information.opening_hours when ingesting real-time feeds before surfacing bikes (i.e. this is a breaking proposal) or we can continue gating by is_renting=false, is_returning=false and no bikes in free_bike_status. 

This one is harder to answer - It seems like for now, you would have to check `system_information.opening_hours` in the case of free bikes. In  `station_status` ,  the `num_bikes_available` and `num_docks_available` fields give the numbers that would be available if the station were set to allow rentals. There's a long thread on how this came to be in #94 . The reason was so that you could plan a trip or reserve a vehicle at 5:45 when the system opened at 6:00.  There's no equivalent to `num_bikes_available` in `free_bike_status` but there's also no where that it says there should be no bikes in `free_bike_status` when rentals aren't allowed, only when there are no bikes on the street.  If there are bikes on the street, they could still appear in `free_bike_status` when `opening_hours` evaluates to `closed`, that's no different from the current spec, it's just that the format and location of the hours of operation are different.

kanagy: I see. There's also `is_disabled` field in free_bike_status which says "This field is used to indicate vehicles that are in the field but not available for rental." It sounds like this field should then be set to false outside of opening hours?

mplsmitch: I don't know if I agree with that. With station based ( docked) vehicles that's not really how it works and I think it should be the same for free floating.  The numbers represented by `ststion_status.num_bikes_available` and `num_docks_available` are the number of functional bikes/docks that would be available if `is_renting` = `true`. Suppose there's a free floating vehicle with a dead battery and you set 'is_disabled`= `true'  when the system is closed. In the above scenario where you're planning a trip at 5:45 for the system that opens at 6:00, it would appear that this vehicle will be available for rent at 6 but it won't be.

yocontra:
+1 from Stae

gbfs.md
kanagy: There should be a separate PR to improve documentation / meaning of the is_disabled wrt opening hours.

testower:
Entur supports this proposal

nbdh:
+1 from nextbike

heidiguenin:

ncancelliere:
+1 from Spin on this proposal.

heidiguenin:

Votes in favor:
Transit (consumer)
Stae (consumer)
Entur (consumer)
nextbike (producer)
Spin (producer)

There were no votes against. 
Thank you to everyone who took the time to review and to vote on this!

We'll update the documentation as discussed, and then we will tag and merge this into v3.0-RC2 in the coming weeks.

PR #331, opened 2021-06-10
Reserve time for vehicle types and pricing plans

mplsmitch:

Many systems allow users to place a hold on a vehicle immediately in advance of a rental. The length of time that vehicle may be reserved is typically < 30 minutes. During this time the vehicle cannot be rented by another user. If a ride is not initiated before the reserve time elapses, the vehicle is released and becomes available to other users.

Trip planners need to know if vehicles can be reserved and the maximum duration of reservations.

This type of service is distinct from reservations placed at some point in the future, as is commonly used in car sharing.

This issue was previously covered in PR #257  but failed to garner enough votes to pass.

A needs analysis of this issue can be found [here](https://docs.google.com/document/d/1z7jmxAEZa2Lzz_h2mM25WdvSIC7lmwMpraKCsfBhaNg/edit).

This proposal adds a field to vehicle_types.json and system_pricing_plans.json that contains the maximum duration in minutes for which a vehicle may be reserved in advance of the start of a rental.

There are are known use cases where vehicle reservations can be modeled by vehicle type. I have not found any where the ability to reserve a vehicle it limited by the pricing plan so that part may be speculative. 

- [ ] Yes 
- [x] No
- [ ] Unsure

gbfs.md:
- vehicle_types.json
- system_pricing_plans.json


mplsmitch:

nbdh:
Nextbike supports this proposal and we would probably add `reserve_time` while implementing `vehicle_types.json`.

viestat:
Dott supports this proposal, however this option will not be implemented in our end in the foreseeable feature.

evansiroky:
IBI Group supports this proposal.

gbfs.md
```diff
@@ -840,6 +844,7 @@ Field Name | REQUIRED | Type | Defines
...
 \-&nbsp;`description` | Yes | String | Customer-readable description of the pricing plan. This SHOULD include the duration, price, conditions, etc. that the publisher would like users to see.
+\-&nbsp; `reserve_time` | OPTIONAL | Non-negative Integer | Maximum time in minutes that a vehicle can be reserved before a rental begins. This field SHOULD be used in cases where the ability to reserve a vehicle is dependant on the pricing plan. Where this is not the case `vehicle_types.json#reserve_time` SHOULD be used. When a vehicle is reserved by a user the vehicle remains locked until the rental begins. During this time the vehicle is unavailable and cannot be be reserved or rented by other users. The vehicle status in `free_bike_status.json` MUST be set to `is_reserved = true`. If the value of `reserve_time` elapses without a rental beginning, the vehicle status MUST change to `is_reserved = false`. If `reserve_time` is set to `0` vehicles cannot be reserved using this pricing plan.
```
kanagy: Sorry, I find having reserve_time in two different feeds quite confusing. I don't want us having to implement a feature in two different ways because half of partners do it via pricing plans another half via vehicle types. Also, when you introduce default_pricing_plan_id in vehicle_types, you'll have to define which reserve time takes precedence (because a vehicle_type with reserve time A can reference default_pricing_plan with reserve time B). Since you already mention that this is a speculative use case, why not just remove it from the pricing plan feed? Other options: add guidance for which field should be used in which circumstances and perhaps rename the one in vehicle_types to default_reserve_time.

That said, we don't plan to use this field any time soon and I don't want to block the PR. Will just abstain from voting.

cait32:
BCycle supports this proposal, but we are unlikely to implement it in the near future since we do not allow for vehicle reservations.

ncancelliere:
Spin supports this proposal, we do allow reservations; however, I want to echo the concerns of @kanagy (https://github.com/NABSA/gbfs/pull/331#discussion_r699865773 ) which I share. I suppose it offers flexibility for providers but it does maybe overcomplicate things - especially if no one has asked for it or has use cases with respect to pricing plans determining reservation capability.

mplsmitch:
Based on the comments from @kanagy and @ncancelliere, how do people feel about amending this proposal before it gets merged? Since we don't have an actual use case for including the `reserve_time` field in pricing plans I think we should drop it for now.  We could rename the field in `vehicle_types` to `default_reserve_time` and then if this comes up in the future we would define the rules to override  the default based on the use case.

ezmckinn:
Superpedestrian supports this proposal, and echoes the concerns raised by @ncancelliere and @kanagy, as well as the solution proposed by @mplsmitch. We are unlikely to implement in the near future, as we don't currently allow reservations.

josee-sabourin:
This vote is now closed, results will be shared tomorrow morning.

mplsmitch:
This proposal has passed.
Votes in favor:
NextBike (producer)
Dott (producer)
IBI Group (consumer)
Bcycle (producer)
Spin (producer)
Super Pedestrian (producer)

Please see my note above regarding amending this proposal before merging into RC-v2.3 and weigh in with any comments.

Issue #326, opened 2021-05-19
Station Services Needs Assessment


mplsmitch:
### **What is the issue and _why_ is it an issue?**
As station hardware and services evolve, there is a need to represent a wider array of services and attributes at the station level.  The presence of non-standard fields in feeds describing station services and hardware indicates a need for extension to the specification.  

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**

MobilityData has drafted a [needs assessment ](https://docs.google.com/document/d/1cd7pLD9bmGZSQYsseMez5JM3J9xznzEUETvPnUvNnYQ/edit?usp=sharing)including a number of potential use cases that could be addressed in a future extension to the specification. 

Potential additions to the specification might include:

- Fields that would allow operators to describe services available at the station along with hours and dates of availability.
- Fields that would better describe station hardware and related services like vehicle charging or the presence of a payment kiosk.
- Fields that would describe features of the station environment like lighting or accessibility issues.

Please [comment on the needs assessment document](https://docs.google.com/document/d/1cd7pLD9bmGZSQYsseMez5JM3J9xznzEUETvPnUvNnYQ/edit?usp=sharing) or here in this issue. We are particularly interested hearing from data producers and consumers who have an interest in the use cases identified in the needs assessment. We're also interested in identifying additional needs related to station services and hardware.

### **Is your potential solution a breaking change?**
- [ ] Yes 
- [ ] No
- [x] Unsure

heidiguenin:
This issue has been moved to PR #340  - I'm closing this issue. Please continue the conversation on the PR.

Issue #280, opened 2020-12-16
Station type as part of station information


gerazenobi:
Hi there,

I would like to know whether there is a way to know if a station is of a certain type ? E.g. I would like to be able to differentiate those stations that support charging electric bikes from those that don't. 

Examples seen out there
```json
             {
                "station_id": "7656",
                "physical_configuration": "REGULAR",
                
            },
            {
                "station_id": "7657"
                "physical_configuration": "ELECTRICBIKESTATION"
                
            }
```
```json
            {
                "station_id": "1",
                "is_charging_station": true,
            },
            {
                "station_id": "7",
                "is_charging_station": false,
            }
```

Is this already supported and I don't seem to make the connection between the different files/datasets? E.g we know which types of vehicles and how many can park on the station but it will be wrong to assume that by allowing electric bikes to park that they will be able to charge as well.

Thanks in advance.

l4b4r4b4b4:
I can only second that request! It is an essential feature, if not yet implemented.

mplsmitch:
@gerazenobi To answer your question, this is not currently supported by the specification.  I agree that there is need for it. To start with I'd like to gather more information to better understand the needs.  Could you post the URLs for the examples you've seen?

gerazenobi:
Thanks for the quick response.

The examples belonged to the same system but presented in 2 different ways depending on the file/dataset:
- https://toronto-us.publicbikesystem.net/ube/gbfs/v1/en/station_information
- https://toronto-us.publicbikesystem.net/ube/gbfs/v1/en/station_status

other systems sharing the same mechanism/approach:
- https://barcelona-sp.publicbikesystem.net/ube/gbfs/v1/en/station_information
- https://barcelona-sp.publicbikesystem.net/ube/gbfs/v1/en/station_status
- https://riodejaneiro-br.publicbikesystem.net/ube/gbfs/v1/en/station_information
- https://riodejaneiro-br.publicbikesystem.net/ube/gbfs/v1/en/station_status

gerazenobi:
< reviving this issue so it doesn't go stale >

mplsmitch:
@gerazenobi We're planning to work on this issue next month - May. If anyone has other ideas about enhancements to stations please let me know.

mplsmitch:
[...]

PR #329, opened 2021-06-01
Add vehicle drop off restrictions

josee-sabourin:
#### **What problem does your proposal solve? Please begin with the relevant issue number. If there is no existing issue, please also describe alternative solutions you have considered.**
As discussed in issue #270, there is no way to describe how a vehicle should be returned at the end of a trip.  

#### **What is the proposal?**
This proposal adds a `return_type` field to `vehicle_types.json` that allows operators to define if the vehicle is `free_floating` (vehicle must be returned anywhere within the service area), `roundtrip` (vehicle must be returned at original rental station), or `station` (vehicle must be returned to any station within the service area). The field is modelled as an array to account for hybrid vehicles that have multiple ways of being returned. 

Additionally, to account for edge cases where a vehicle must be returned to a particular station that is not the original station, a `home_station` field has been added to `free_bike_status.json`.  

Read the full extension proposal [here](http://bit.ly/GBFS-DropOffs-Proposal).

#### **Is this a breaking change?**
- [ ] Yes 
- [x] No
- [ ] Unsure

#### **Which files are affected by this change?**
- vehicle_types.json
- free_bike_status.json

PR #340, opened 2021-07-20
Add field to designate stations that support vehicle charging

mplsmitch:
Adds field to designate stations that offer vehicle charging

The specification lacks a way to designate stations that offer electric vehicle charing. Discussion of this issue can be found in #280 and #326.  Publishers are already doing this in  a variety of ways using non-standard fields.

This proposal adds a single Boolean field: `is_charging_station`,  to `station_information.json`

- [ ] Yes 
- [x] No
- [ ] Unsure

gbfs.md:
- station_information.json


mplsmitch:
CC: 
@gerazenobi @l4b4r4b4b4 @thlaunois

kanagy:
Hello,

How does this proposal interact with https://github.com/NABSA/gbfs/pull/329 ? We're wondering how to model whether an electric bike can be returned to any station, or we should suggest returning it only to charging stations.

mplsmitch:
@kanagy Good question. I'm not aware of any cases where only returning electric bikes to a charging station is a requirement. This may change in the future as charging capable stations become more common. If that's the case we could add another enum like `charging_station` to the `return_type` array in #329.  What I expect to see is that only certain locations will offer charging and operators will offer incentives to get users to return vehicles where they want them.

gerazenobi:
@mplsmitch a small addition to your [comment](https://github.com/NABSA/gbfs/pull/340#issuecomment-892080642): there would be also the rebalancer's side that would decide its following actions based on that info for example.

mplsmitch:

mplsmitch:
CC: @gerazenobi  @l4b4r4b4b4 are you still interested in supporting this?

nbdh:
Nextbike supports this proposal and we would add this field to our feeds in the future.

gerazenobi:
> CC: @gerazenobi @l4b4r4b4b4 are you still interested in supporting this?

yes @mplsmitch 👍

heidiguenin:
Hi @gerazenobi ! For your vote to count, we need to know which organization you're working with 😄

gerazenobi:
> Hi @gerazenobi ! For your vote to count, we need to know which organisation you're working with 😄

Oups, my bad @heidiguenin : the organisation is Qucit.

viestat:
Dott supports this change, however at the current moment this is a field that would not apply to our free floating model.

evansiroky:
At IBI Group we probably won't use this information since there don't seem to be dropoff restrictions or incentives associated with this particular addition, therefore we are neutral on this topic. 

As an aside, I wonder how this would interact with the potential changes in https://github.com/NABSA/gbfs/pull/350 that would expand the number of propulsion type definitions in the vehicle_types.json file. In doing so, this raises the question of whether this field is appropriate for just e-bikes and e-scooters or if it also applies to vehicles with any propulsion type. If it does apply to propulsion types other than electric vehicles, then "is_charging_station" is probably a confusing name. If it does apply only to those vehicle types, then the field's description should note this information.

cait32:
BCycle supports this change. We do not currently have stations that support vehicle charging so we will not add the field immediately, but we would add the field in the future.

ncancelliere:
Spin supports this change. We likely will not immediately implement this.

ezmckinn:
Superpedestrian supports this change. We don't plan to implement this, as our system does not use stations.

josee-sabourin:

Votes in favour: 
Nextbike (producer)
Qucit (consumer)
Dott (producer)
IBI Group (consumer)
BCycle (producer)
Spin (producer)
Superpedestrian (producer)

There were no votes against. 

Regarding @evansiroky's comment: this field will likely only represent charging stations for electric vehicles, when we merge this into a v2.3-RC we will make sure to that is clear.

PR #341, opened 2021-07-21
Add GoAbout feed to systems.csv

jcassee:
My name is Joost Cassee. I'm with GoAbout, which is a mobility provider in The Netherlands. This is our GBFS feed.


mplsmitch:
Thanks @jcassee! Always good to see a new provider. I see GoAbout is a car sharing provider. Do you plan to use GBFS for car sharing also?

jcassee:
Hi @mplsmitch! Thanks for welcoming into the systems list.

Our cars are only used with a subscription, so I think it does not really make sense to have them in our GBFS feed. :slightly_smiling_face:

PR #343, opened 2021-07-23
Fix typos in Canadian city names

tarvos:
The primary change in this PR is a correction of misspelled Canadian city names (`*Ottowa` and `*Edmontan`).

Secondarily, this PR ensures all Canadian city names replicate the `[city], [country code]` format used in most other entries--though I noticed this scheme is not consistent. Exceptionally, cities in the United States are largely listed `[city], [state code]` (which are not [ISO codes](https://www.iso.org/iso-3166-country-codes.html)). My instinct is to identify Canadian cities as `[city], [province code]`, but I recognize my bias as a Canadian-American, which is why I deferred to the country code convention.

The consistent scheme that would demand the fewest changes to this database would be to identify all (U. S. of) American cities as `[city], US`. Alternatively, because `Country Code` is its own value anyway, we could remove _all_ country codes in the `name` field and add subdivision abbreviations to applicable cities in any country. This is my first contribution to this (or any!) open source project so I leave that choice to more authoritative community members.


mplsmitch:
Thanks @tarvos.  It's true that the use of US states doesn't conform to the rest of the document. We're planning a rebuild of the systems catalog in the near future and that's one of the issues to be resolved.

PR #345, opened 2021-07-30
Adds feeds for Helbiz, PBSC and Bcycle

mplsmitch:
Adds feeds for Helbiz in Italy and US, PBSC in France and Bcycle in US


bddq:
@mplsmitch Why all Helbiz systems have the same `system_id`?

mplsmitch:
@bixcorp that's what they've published, unfortunately the same is true of the feeds from Neuron and Wheels. I'm hoping that by including them in the catalog more data consumers will take an interest in what they are publishing and we can get them to update and improve their feeds.

Issue #338, opened 2021-07-12
 Proposal for car-sharing support


Miryad3108:
### Introduction

[Transport.data.gouv](https://transport.data.gouv.fr/?locale=en) is the French national access point for mobility data.

We are opening this issue on our behalf and on behalf of a working group involving  the “Association des Acteurs de l’Autopartage” (AAA), a French National Association for carsharing [grouping 20+ operators](https://asso-autopartage.fr/), and [Iodines](https://www.iodines.fr).

### Why are we opening this issue?

Following [European Directive EU 2010/40/UE](https://eur-lex.europa.eu/legal-content/en/TXT/HTML/?uri=CELEX:32017R1926&from=EN), French carsharing operators will be requested to open and publish their data.

We have noticed strong similarities between car sharing and bike sharing. We are considering the opportunity to evolve the GBFS standard in order to support both use cases and to leverage GBFS proven efficiency and popularity.

**The goal of this issue is to discuss this idea with the community, before we go forward with a PR**.

### Proposal overview

We would like to propose an evolution of the GBFS which allows to model car-sharing data by extending the following files.

The detailed proposal to discuss is currently in this [Google Document](https://docs.google.com/document/d/1bgNsiTcTfjKxG6khGq0ro0x-vEaToihp0_t-krGyj1o/edit?ts=606c5d87#).

- **vehicle_types** - new optional data only:
  - rider capacity
  - trunk size
  - C02 emission
  - environmental classification
  - accessories
  - ...

- **station_information.json**: 
  - BREAKING: share type allowed by station : free floating, one-way or round-trip (see below for motivation)
  - Other optional types specific to car stations (type of station...)

- **free_bike_status.json**:
  - BREAKING: the file would be renamed to **free_vehicle_status.json** 
  - new optional data only: 
    - a datetime specifying how much time the vehicle is available (for round trip carsharing)
    - accessories present in the vehicle (such as child seats)

### **Is your potential solution a breaking change?**

Yes.

It seems that currently in the GBFS the share type is implicitly defined by the type of station : in case of a physical station, the share type is one-way. In case of a virtual station, it is free floating.

As the round trip share type is commonly used by car-sharing companies, the share type would become explicitly defined among one-way, free-floating and round-trip.

### Note on potential variant of the proposal

[...]

PR #350, opened 2021-08-13
Carsharing extension

fchabouis:
#### Who we are
[Transport.data.gouv.fr](https://transport.data.gouv.fr/?locale=en) is the French national access point for mobility data.

We are opening this PR on our behalf and on behalf of a working group involving the “Association des Acteurs de l’Autopartage” (AAA), a French National Association for carsharing grouping 20+ operators, and Iodines.

#### What problem does your proposal solve?
This PR aims to extend the current GBFS specification to allow car sharing support.
More details can be found in the issue #338.

#### Proposal overview
We would like to propose an evolution of the GBFS which allows to model car-sharing data by extending the following files. This evolution is based on the addition of new fields and the modification of existing fields. 

- `vehicle_types.json` 
New optional fields:
        - `rider capacity`
        - `trunk size`
        - `C02 emission`
        - `environmental classification`
        - `accessories`
        - ...
Field that has been modified: 
- addition of new allowed values in `propulsion_type`

- `station_information.json` 
New optional fields:
      - `parking_type`
      - `parking_hoop`
      - `contact_phone`
Field that has been modified: 
-  description of the field `is_virtual_station`

- `free_bike_status.json` 
New optional fields:
      - `available_until`, a datetime specifying how much time the vehicle is available (for round trip carsharing)
      - `vehicle_equipment`, accessories present in the vehicle (such as child seats)
Field that has been modified: 
- modification of the description of `current_range_meters`

- [ ] Yes 
- [x] No
- [ ] Unsure

- `vehicle_types.json`
- `station_information.json`
- `free_bike_status.json`


fchabouis:
In this PR, we have added the `Phone number` field definition, similarly to the proposal made in #351. As this field type is already referenced in the `file_system.json` file, but with a description different from the one we give, it could be confusing.

To keep the proposal non breaking, we could remove the new field definition and let #351 add it afterward.

Concerning `contact_phone`, the field we added in `station_information.json`, we could specify in its description that it should follow the E.164 standard. Then, when #351 is merged, we could revert the field description as it wouldn't be useful anymore.

gbfs.md
```diff
@@ -416,14 +434,37 @@ Field Name | REQUIRED | Type | Defines
         "form_factor": "scooter",
         "propulsion_type": "electric",
         "name": "Example E-scooter V2",
-        "max_range_meters": 12345
+        "max_range_meters": 12345,
```
mplsmitch: comma at the end of 437 should be removed

mplsmitch:
@fchabouis  I don't see `current_range_percent` as a suitable  substitute for `current_range_meters`. If the intention is that a vehicle with `current_range_percent` of `0.75` has a range of 75% of `max_range_meters`, then that calculation is already possible using existing fields: `current_range_meters` / `max_range_meters` = `current_range_percent`.  If `current_range_percent` is intended as a measure of vehicle range, and the producer is able to provide `current_range_percent` then there's no reason they can't calculate and provide `current_range_meters`. If that's the case then it would be better if this  range calculation were done by the data producer so that consuming applications wouldn't have to look at multiple fields for the data, making the calculation more prone to error. 

I understand that an end user would find `current_range_percent` useful, for example they may prefer to choose the vehicle that is 85% full versus the one that is 25% full, but as it's name implies that this is the percentage vehicle range, not it's fuel or charge level. The `current_range_percent` definition doesn't mention range, only fuel or charge level. I suggest renaming the field to something like `fuel_charge_percent` and either making it OPTIONAL or conditionally Required if the vehicle has a motor.

fchabouis:
Hello @mplsmitch , thank you for your feddback !
We had an interesting conversation with a carsharing operator, whose fleet is only composed of electrical vehicles. I think it was Iodines. Their point was : they didn't want to give users a range in meters because an electrical car range heavily depends on the driving style and type of roads used. Highways at 130km/h and cities would yield very different ranges. They felt that giving users a range in meters could therefore be problematic, if they say the car has a current range of 300km and the driver finds himself with an empty battery after a 200km drive. That's why they felt more comfortable giving a percentage of charge of the battery. But for a e-scooter company operating in a city, giving ranges as a distance makes more sense, as the driving conditions are more homogeneous, and the consequences of an empty battery are not the same (no baby crying on the back seat!).
What do you think ?

NicolasFrasie:
Hello @mplsmitch , there is another reason than the one mentionned by @fchabouis. It is generally not possible to know directly from the vehicle which range the vehicle displays on the dashboard unlike the % of charge of battery. Now, if the range in km is calculated as you propose, the end user may feel deceived if the range actually displayed is very different from the range previously given (for the reasons mentionned previsouly).

This being said, this is technical constraint isn't ideal for the users and I hope in the future operators will be able to provide a precise range in km, maybe depending on each user, geography, local temperature, etc.

mplsmitch:
@fchabouis I think it's important to remember that the consumers of these data are journey planning and MaaS applications, not end users.  For the journey planner it's impossible to match an appropriate vehicle with a user without knowing the estimated vehicle range. 

I agree that the carshare use case is significantly different from escooters and ebikes but I think the need for range data to support these applications is still there.  If the fuel or charge lever is substituted for `current_range_meters` ,  journey planners will have to calculate the vehicle range in order to do their work. The operator of the vehicle is much better equipped to make this calculation than the journey planner.  Since there are many variables affecting range, calculation of `current_range_meters` must be very conservative.  Where I live, the range of electric vehicles is far less in winter than in summer depending on the temperature. A vehicle who's battery has been through 50 charge cycles may have a  range that is close to 100% of its maximum whereas the same vehicle with 700 charge cycles would have much less range. For these reasons you would not say that a vehicle with 50% charge can travel 50% of its maximum range.  

> It is generally not possible to know directly from the vehicle which range the vehicle displays on the dashboard unlike the % of charge of battery. Now, if the range in km is calculated as you propose, the end user may feel deceived if the range actually displayed is very different from the range previously given (for the reasons mentionned previsouly).

There's a lot of data contained in the specification that's used to facilitate journey planning but is never shown to end users. While a journey planner may use `current_range_meters` to match a user with a vehicle,  they don't need to display that data to the user. Many of them prefer to display the  % charge level which is currently being calculated as `current_range_meters / max_range_meters`.

NicolasFrasie:
> The operator of the vehicle is much better equipped to make this calculation than the journey planner.

I disagree. Yes, the range of a vehicle (in km) can vary along the lifespan of the battery, but - for what I understand - this should be modelled by udpating the max_range_meter field (provided that the operator knows how to measure this drop in capacity).

Moreover, as mentionned earlier, it is generally impossible for carsharers to provide the current_range_meters (unless calculating current_range_percent * max_range_meters).

evansiroky:
I remember discussing this topic of including range as meters or range as percent a lot in the initial work to add vehicle type definitions. See a comment in the middle of the thread here: https://github.com/NABSA/gbfs/pull/136#issuecomment-583976555.

As far as the proposed modifications proposed in this PR, I like that the `current_range_*` fields are still defined as numeric values. However, it does seem somewhat unnecessary since either percent or absolute range can be naïvely derived from the other. It seems the primary concern and motivation for including the percentage field is not wanting to mislead users on range estimates. However, from a trip planning perspective, trip planners must know the range or pretend they know it by making speculative calculations about how the vehicle itself, the environment the vehicle travels through and certain user behaviors all interact to result in how far a vehicle can be expected to travel before it runs out of energy. And I would say this applies even within cities although more so in those with lots of hills.

Maybe one differentiating factor about the car use case, in particular, is that it is likely more common for someone to rent a car and refill the gas tank. This is opposed to the e-scooters and e-bikes short-term rental use case where a vehicle may not have as long of a rental time period and users, therefore, may not recharge/refill the vehicle or even have the ability to do so. And on the trip planner side, lots if not most trip planners provide driving directions without built-in calculations of where to stop to refill/recharge a car. It is assumed that people have a common knowledge that refilling/recharging a vehicle will be required at least a few times if one were to attempt to drive across the US for example.

I'm not sure what the best user experience is in terms of having this current range communicated to users, but maybe a good start on that topic is to add a field (perhaps in this PR or elsewhere) that describes whether a vehicle can be recharged/refueled, how long that takes and where reenergizing can happen. Then trip planners would at least be able to better communicate whether journeys could occur that exceed a vehicle's maximum full-energy range. Also, it could help to communicate an expected "average" range per charge/full tank. Perhaps this topic is something that could eventually be in a best practices document for how to produce this value and for how client applications should communicate it to users.

kanagy:
I agree with both Evan and Mitch points. My 2c for an alternative solution: As a trip planner, knowing maximum range is useful for picking the vehicle around you that can actually make the desired trip. However, for petrol vehicles that can be easily refueled this doesn't really matter and we wouldn't take it into account. Perhaps a simple solution is to make current_range_meters only required for electric propulsion, and optional for other use cases? From experience, many carsharing companies don't know in realtime how much petrol would be in a vehicle.

NicolasFrasie:
I share your analysis of the difference between electric vehicle and ICE vehicles, and I understand your needs as a trip planner. However, operators generally don't know the current_range_meters of electric vehicle unless calculating current_range_percent (provided by the vehicle) / max_range_meters (theorical one, based on WLTP). 

I have the understanding that a trip planner is much more equipped to make the calculation of the required range to realize a trip, based on km, elevation gain, speed (if highway, if congestion, etc.) and temperature. I speak here about a trip of tens of km, not just a few kilometers that should be better realized by bike, escooter, etc.

fchabouis:
What about giving a `current_range_meters` and an optional `current_minimum_range_meters` ?

* `current_range_meters` could be used to give the estimated vehicle range, and battery capacity could be deduced from that value, dividing it by `max_range_meters` (that's the current situation).
* for renting situation where the effective range can have important consequences, a minimum range in meters could be given, using conservative hypothesis. This value could be used by trip planners to ensure the vehicule can make the desired trip, or at least refuel on the way safely.

For example, with `current_range_meters = 300 000` and `current_minimum_range_meters = 100 000` :

* a 50km trip is planned => no problem
* a 150km trip is planned => vehicle is proposed for renting with a message "a refill may be necessary on the way"
* a 350km trip is planned => vehicle is proposed for renting with a message "a refill will be necessary on the way"

This way data producers can avoid deceiving users with false promises, but still have to provide an information about the range of the vehicule they are renting. For vehicles where refueling is not a problem, like gas cars, the `current_minimum_range_meters` could be omitted.

fchabouis:
I'd like to know what a data producer (@NicolasFrasie ) and other people involved (@kanagy @evansiroky @mplsmitch ) think of a `current_minimum_range_meters` field. Thanks !

mplsmitch:
> current_range_meters could be used to give the estimated vehicle range, and battery capacity could be deduced from that value, dividing it by max_range_meters (that's the current situation).
for renting situation where the effective range can have important consequences, a minimum range in meters could be given, using conservative hypothesis. This value could be used by trip planners to ensure the vehicule can make the desired trip, or at least refuel on the way safely.
For example, with current_range_meters = 300 000 and current_minimum_range_meters = 100 000

In this example, if the data producer believes that the vehicle can realistically travel 100 000, then they should set `current_range_meters` at 100 000.   

Adding another field to the specification would increase ambiguity. Which of these should a data consumer be using to match  a traveler with a vehicle? If `current_minimum_range_meters` were defined as the actual range the vehicle is capable of traveling with its current fuel/charge level then what would be the purpose of publishing `current_range_meters`?

NicolasFrasie:
Thanks @fchabouis for the proposal. I producer / operator can already provide a "minimum range" at the  current_range_meter 

I suggest to keep it very simple at first :
- add current_range_% 
- keep current_range_meter field optionnal
- specify to producer that, if they don't provide a current_range_meter, it will be estimated from current_range_%  x max_range

This may be enough for urban trip planner : for trips less than 10 km, range isn't a problem (as an example, at Communauto, EV with less than 20% range are blocked). I think a problem will arise when it comes to considering longer journeys like round trips (as an example : Montreal, February, minus 20 degrees, I want to pick aup a relative with lot of luggages at the airport 30 km away, and come back, at peak time).

fchabouis:
I think, after reading again all the comments, that we could :
 - rename `current_range_percent` to `fuel_charge_percent` (or something similar) as proposed by @mplsmitch in his first comment, effectively reflecting that this field indicates the current percentage of the battery / fuel tank. And make this field OPTIONAL to keep the proposal non breaking.
 - keep `current_range_meter` REQUIRED when the vehicle has a motor, and let the data producer be as conservative as needed about that information.

NicolasFrasie:
I support @fchabouis's last comment and hope we can have the opportunity to discuss about EV range calculation and round trips integration into A-to-B trip planner.

So : 
- rename `current_range_percent` to `fuel_charge_percent` (or something similar) and make it optional. Watch out hybrid vehicles with 2 range / fuel level. I suggest to speak about "main energy supply".
- keep `current_range_meter` required in order to avoid the breaking change BUT add a "disclaimer" in the definition of the field to indicate that the value can be different from the autonomy displayed at the dashboard in the vehicle.

NicolasFrasie:
I have reviewed the whole proposal and would like to suggest three modifications : 
- Delete mention of `combustion_diesel`, `hybrid`, `hydrogen `and `plug-in hybrid` in the `propulsion_type`, but change "_Contains throttle mode with a battery-powered motor" per_" _Contains throttle mode with a battery-powered **or fuel-cell** powered motor_". The purpose of this adding was to manage Low-Emission Zone, but the `g_CO2_km` and `eco_sticker` fields solve this issue. And this would lead to a futur issue to suit with MDS requirements.
- In `max_range_meters`, require WLTP.
- Add `stat_desc` in `station_information.` to provide for example the space number and floor number. Definition could be "_(Optional, Object) An object where each key defines one of the items listed below.- `Text `: Description of the station that provides useful, quality information. Do not simply duplicate the information already specified. 50 characters. - `Language `: XXX_"

Miryad3108:
Thank you @NicolasFrasie  for your suggestions: 

- Concerning fuel, this seems to be relevant information for traveler information services such as route planners. For example, a route planner will not direct you to the same stations if your car is hydrogen because it is not available everywhere.
Moreover, some operators may not fill in the Crit'Air field, especially outside the metropolis or outside the European Union. Maintaining these values can be useful in these cases.
For the modification "Contains throttle mode with a battery-powered motor" per" Contains throttle mode with a battery-powered or fuel-cell powered motor": thank you for having raised this point. We will add it to the proposal

- WLTP is only for passenger cars, not for scooters, electric bikes, etc. It cannot be required. Moreover, this protocol is specific to the countries of the European Union. We suggest to recommend the use of this protocol in the documentation of transport.data.gouv.fr and to specify which protocol is used if it is not this one.  

- For the addition of the stat_desc field in station_information, we have had feedback from reusers who advise against the use of free fields because the producer could be tempted to put important information in them but the reuser will not necessarily reuse this information. We prefer not to put it in this version for the moment. However, the GBFS is an evolving standard, if so if operators or reusers need a free text field, we can add it later

NicolasFrasie:
Thanks @Miryad3108 for your comments : 

- Delete mention of combustion_diesel, hybrid, hydrogen and plug-in hybrid in `propulsion type` : This is a required field, and I still don't see the point for providers .. is there any known use case by a reuser? If there isn't, I suggest to have this discussion in the future when the need will show up.
- WLTP : Well seen ! I suggest to mention it in the description of the field at GBFS level.
- Add `stat_desc` in `station_information` : Fine.

fchabouis:
Maybe @kanagy could give us his input from a trip planner perspective. Would a detailed `propulsion_type` be useful (including combustion_diesel, hybrid, hydrogen and plug-in hybrid) or would the existing choices be enough?

mplsmitch:
I've read through the proposal and I have a few comments and questions.

- First of all, in it's current state I don't think this is ready for a vote because many of the tables in the PR are full of extra spaces which makes it impossible to read the diff. There's no way for reviewers to understand where the additions and subtractions are because nearly every line has been changed with the addition of these spaces. See the "Files changed" tab and select "Load diff" to see what I'm talking about.

- In `vehicle_types.json` there are two new fields, `min_cargo_load_capacity` and `min_cargo_volume_capacity`. Shouldn't these both be expressed in terms of maximums? As it's written they are expressed as minimums which implies for example there there is no upper limit to the cargo capacity. 
-  In the `make` and `model` fields, why are these defined as all lowercase? How are consuming applications expected to know when these contain proper nouns and mixed-cases needing capitalization? For example, how would a consuming application know that `peugeot ion` needs to be converted to `Peugeot iOn` ?
- in `vehicle_types.json#vehicle_type_accessories`, first I suggest shortening this field name to `vehicle_accessories` to match the naming  conventions used for other fields in the endpoint.  The way that the Enum `doors_4` is defined implies that if this Enum is not included, then the vehicle has 2 doors. I don't think that's the intent. I would suggest either adding a separate `num_doors` field defined as non-negative integer, or adding addition Enum values `doors_2` `doors_3` `doors_4` `doors_5` to the list of vehicle accessories.
- Lastly, the code examples  for `vehicle_types.json` and `free_bike_status.json` aren't valid JSON. In the case of vehicle_types it looks like there's a duplicate key `max_range_meters`.

AntoineAugusti:
@mplsmitch Thanks for taking the time to read our proposal. We'll address your comments and answer your questions in the coming days.

gbfs.md
```diff
@@ -1264,7 +1323,7 @@ Note that the Android URI and iOS Universal Link URLs don’t necessarily use th
     "language": "en",
     "rental_apps": {
       "android": {
-        "discovery_uri": "com.example.android://",
+        "discovery_uri": "com.example.android://"
```
thbar: @fchabouis is the comma removal voluntary?

fchabouis: thanks @thbar I forgot to cherry-pick a [commit](https://github.com/NABSA/gbfs/pull/374) when fixing the formatting issue

fchabouis:
Thank you @mplsmitch for reading our proposal and for your comments. Concerning the whitespace issue and the faulty examples, I fixed it yesterday.

`min_cargo_load_capacity` and `min_cargo_volume_capacity` were originally called `cargo_load_capacity` and `cargo_volume_capacity` but carsharing operators where worried that if similar vehicles (but slightly different) belonged to the same `vehicule_type`, the trunk capacity could vary.  For example 200 liters for a car and 210 liters for another one. The idea was thus to give a minimum capacity (200 liters), so the user knows that the vehicle he is going to rent has at least 200 liters capacity for the trunk.

An another option would be to create one vehicle type for each different vehicle, no matter if the difference is small. Carsharing operators we talked to were worried that this would lead to having one vehicle type per vehicule in their fleet.

Do you have an opinion about this ?

We'll come back to you shortly about the other comments/questions you made.

fchabouis:
I have updated the proposal to remove the make/model case constraint and to add `doors_2`, `doors_3` and `doors_5` options.

mplsmitch:
> For example 200 liters for a car and 210 liters for another one. The idea was thus to give a minimum capacity (200 liters), so the user knows that the vehicle he is going to rent has at least 200 liters capacity for the trunk.

This method puts the burden on the user when it should be on the operator.  Given the example above, If the user and needs 230 liters, are they to take on that risk and hope that the vehicle will meet their requirements? Expressing them as minimums doesn't really help the user and only adds uncertainty and that's potentially a terrible experience for the user. If the `min_cargo_load_capacity` is 500kg what would keep the user from overloading it with 750kg? Would the user be liable for damages when the capacity is implied to be greater than 500kg?

> An another option would be to create one vehicle type for each different vehicle, no matter if the difference is small. Carsharing operators we talked to were worried that this would lead to having one vehicle type per vehicle in their fleet.

There's no reason for  operators to create a large number of vehicle types, they just need to group the vehicles that have similar capacities as is done in this proposal.  In the example 200 liters and 210 liters would be combined in the same vehicle type with a capacity of 200 liters. In the case of the larger 210 liter vehicle, this would would be a positive experience for the user and It's no more work for the operator. I'd drop the `min_` from the field names and simply call them `cargo_load_capacity` and `cargo_volume_capacity` and remove any mention of minimum from the definitions.

Miryad3108:
Hello @mplsmitch , 
Thanks for your comment
We change `min_cargo_volume_capacity` to `cargo_volume_capacity` and `min_cargo_load_capacity` to `cargo_load_capacity`

gbfs.md
```diff
@@ -414,9 +418,21 @@ Field Name | REQUIRED | Type | Defines
...
-\- `propulsion_type` | Yes | Enum | The primary propulsion type of the vehicle. <br /><br />Current valid values are:<br /><ul><li>`human` _(Pedal or foot propulsion)_</li><li>`electric_assist` _(Provides power only alongside human propulsion)_</li><li>`electric` _(Contains throttle mode with a battery-powered motor)_</li><li>`combustion` _(Contains throttle mode with a gas engine-powered motor)_</li></ul> This field was inspired by, but differs from the propulsion types field described in the [Open Mobility Foundation Mobility Data Specification](https://github.com/openmobilityfoundation/mobility-data-specification/blob/master/provider/README.md#propulsion-types).
+\- `rider_capacity`<br/>*(added in vXXX)* | OPTIONAL | Non-negative integer | The number of riders (driver included) the vehicle can legally accommodate.
+\- `cargo_volume_ capacity`<br/>*(added in vXXX)* | OPTIONAL | Non-negative integer | Cargo volume available in the vehicule, expressed in liters. For cars, it corresponds to the space between the boot floor, including the storage under the hatch, to the rear shelf in the trunk.
```
testower: There's a typo here: whitespace character in `cargo_volume_ capacity`.

gbfs.md
```diff
@@ -414,9 +418,21 @@ Field Name | REQUIRED | Type | Defines
...
+\- `propulsion_type` | Yes | Enum | The primary propulsion type of the vehicle. <br /><br />Current valid values are:<br /><ul><li>`human` _(Pedal or foot propulsion)_</li><li>`electric_assist` _(Provides power only alongside human propulsion)_</li><li>`electric` _(Contains throttle mode with a battery-powered motor or fuel-cell powered motor)_</li><li>`combustion` _(Contains throttle mode with a gas engine-powered motor)_</li><li>`combustion_diesel` _(Contains throttle mode with a diesel engine-powered motor)_</li><li>`hybrid` _(Contains throttle mode with a mixed gas engine-powered and battery-powered motor)_</li><li>`hydrogen` _(Contains throttle mode with a hydrogen fuel cell powered motor)_</li><li>`plug-in hybrid` _(Contains throttle mode with a mixed gas engine-powered and battery-powered motor with plug-in)_</li></ul> This field was inspired by, but differs from the propulsion types field described in the [Open Mobility Foundation Mobility Data Specification](https://github.com/openmobilityfoundation/mobility-data-specification/blob/master/provider/README.md#propulsion-types).
```
testower: The enum value `plug-in hybrid` has hyphen and whitespace in it. I suggest changing it to `plug_in_hybrid` or similar.

gbfs.md
```diff
@@ -723,10 +767,14 @@ Field Name | REQUIRED | Type | Defines
...
+\- `current_range_meters` <br/>*(added in v2.1)* | Conditionally REQUIRED | Non-negative float | If the corresponding `vehicle_type` definition for this vehicle has a motor, and current_fuel_percent is not defined, then this field is REQUIRED. This value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel. Note that the given range is indicative and can be different from the one displayed on the vehicule's dashboard.
+\- `current_fuel_percent` <br/>*(added in vXXX)*| Conditionally REQUIRED | Non-negative float | If the corresponding vehicle_type definition for this vehicle has a motor, and current_range_meters is not defined, then this field is required. This value represents the percentage, expressed from 0 to 1 of fuel or battery power remaining with the vehicle’s current charge or fuel level.
```
testower: This conditional is not possible. If there is a vehicle type definition and the vehicle has a motor, then current_range_meter is REQUIRED. This means current_fuel_percent should always be OPTIONAL. And I think that makes more sense anyway, as per @mplsmitch earlier comments, if it should be included at all. I don't think it's a good idea to relax the requirement on providing current_range_meters, since it changes the functional usefulness of max_range_meters in vehicle_types, which amounts to a breaking change imho.

testower:
This proposal is sorely needed, but I think the suggested field `current_fuel_percent`, along with the weakening of the requirement of `current_range_meters` (which is a breaking change), makes the specification less useful. I think all the arguments to this effect have already been presented.

As it currently stands, Entur will not vote in favour of this proposal, but if it is modified to leave `current_range_meters` conditional requirement as is, and remove `current_fuel_percent` - or at the very least, make it fully optional and clarifying its intent, not as a replacement of `current_range_meters`, then we will vote in favour.

iodines-team:
> Hello @mplsmitch , thank you for your feddback ! We had an interesting conversation with a carsharing operator, whose fleet is only composed of electrical vehicles. I think it was Iodines. Their point was : they didn't want to give users a range in meters because an electrical car range heavily depends on the driving style and type of roads used. Highways at 130km/h and cities would yield very different ranges. They felt that giving users a range in meters could therefore be problematic, if they say the car has a current range of 300km and the driver finds himself with an empty battery after a 200km drive. That's why they felt more comfortable giving a percentage of charge of the battery. But for a e-scooter company operating in a city, giving ranges as a distance makes more sense, as the driving conditions are more homogeneous, and the consequences of an empty battery are not the same (no baby crying on the back seat!). What do you think ?

It is totally true and it is not an exception, it is one of the main information to decide to take a vehicle or not. We have to take care here on both user experience and legal aspects. If the information provied to our customer is not relevant, it MUST not be shown. Heater/clim, eco mode, B/D modes, type of car, the way the car is driven, average speed are for example features that can decrease car autonomy by more then 50%. We have been faced to people blocked in the left side of a highway with their battery at 0% at midnight for a 20km estimated trip. Who is than responsible for the extra fees, dammages and security of the people?

testower:
> > Hello @mplsmitch , thank you for your feddback ! We had an interesting conversation with a carsharing operator, whose fleet is only composed of electrical vehicles. I think it was Iodines. Their point was : they didn't want to give users a range in meters because an electrical car range heavily depends on the driving style and type of roads used. Highways at 130km/h and cities would yield very different ranges. They felt that giving users a range in meters could therefore be problematic, if they say the car has a current range of 300km and the driver finds himself with an empty battery after a 200km drive. That's why they felt more comfortable giving a percentage of charge of the battery. But for a e-scooter company operating in a city, giving ranges as a distance makes more sense, as the driving conditions are more homogeneous, and the consequences of an empty battery are not the same (no baby crying on the back seat!). What do you think ?
> 
> It is totally true and it is not an exception, it is one of the main information to decide to take a vehicle or not. We have to take care here on both user experience and legal aspects. If the information provied to our customer is not relevant, it MUST not be shown. Heater/clim, eco mode, B/D modes, type of car, the way the car is driven, average speed are for example features that can decrease car autonomy by more then 50%. We have been faced to people blocked in the left side of a highway with their battery at 0% at midnight for a 20km estimated trip. Who is than responsible for the extra fees, dammages and security of the people?

Yet, my electric vehicle tells me the estimated range in km while I'm driving. This is what I'm looking at while deciding when to start looking for a charging station, not the battery percentage, which tells me absolutely nothing about how far I can go.

iodines-team:
If it is your personal car, it is OK because you know your car, the setup you have done, how the car reacts and the way you drive. In addition, to help you to regulate your driving behavior, the remaining autonomy in meter is shown and updated in real time. Cool and fine in this case !  
But we are not a manufacturer who sales cars. We are carsharing operators who sale mobility services. If the client doesn't don't know the car, the way the last customer has driven, how to setup features to increase its autonomy, he will never be able to know how far he can travel with the vehicle that is in front of him. 
IODINES shows to his clients: Maximum car autonomy range (WLTP) and % battery remaining. We are then legally protected with opposable data shown with a simple and reliable way for custumers to know if they have enough autonomy for their next trip.

testower:
> If it is your personal car, it is OK because you know your car, the setup you have done, how the car reacts and the way you drive. In addition, to help you to regulate your driving behavior, the remaining autonomy in meter is shown and updated in real time. Cool and fine in this case ! But we are not a manufacturer who sales cars. We are carsharing operators who sale mobility services. If the client doesn't don't know the car, the way the last customer has driven, how to setup features to increase its autonomy, he will never be able to know how far he can travel with the vehicle that is in front of him. IODINES shows to his clients: Maximum car autonomy range (WLTP) and % battery remaining. We are then legally protected with opposable data shown with a simple and reliable way for custumers to know if they have enough autonomy for their next trip.

I understand the motivation, which is to limit liability, but the result here is that it weakens the specification and makes the end user experience worse, not better.

I think this matter needs more treatment, so suggest you take it out of the proposal so the rest can be voted upon.

iodines-team:
> I understand the argument, which is to limit liability, but the result here is that it weakens the specification and makes the end user experience worse, not better.
Even the Ux will be better at the end. 
Giving false data to customers who will not be able to end their trip will be worse. 
Ask customers who spent up to 4 hours in cars waiting for assistance, the end of charge, who miss their Rdv because they don't have been able to estimate their autonomy ! 
What about their reaction if the mileage shown in their app was initially compliant with their trip?

> I think this matter needs more treatment, so suggest you take it out of the proposal so the rest can be voted upon.
I agree

fchabouis:
If we take the `current_fuel_percent` out of the proposal, `current_range_meters` will still be required for motor vehicles. So you'll have to fill it anyway @iodines-team .

Have you seen we added `Note that the given range is indicative and can be different from the one displayed on the vehicule's dashboard.` to `current_range_meters` description to make it clear the range is only indicative?

mplsmitch:
@iodines-team 

> If the information provied to our customer is not relevant, it MUST not be shown.

It's been said earlier in this thread but it's worth repeating;  the consumers of these data are journey planning and MaaS applications, not end users.  

The journey planners can't plan a trip based on the traveler's requirements without vehicle range as a factor. Not providing a range estimate will not change this. If they are given `current_fuel_percent` in place of `current_range_meters`, they will be forced to calculate the vehicle range using `max_range_meters` * `current_fuel_percent` because those are the only data they are given.  

If the vehicle operator provides the range estimate, they can do it conservatively, based on some of the  factors you listed above that may affect vehicle range. If the journey planner does the range estimate, those factors probably won't be considered and the vehicle operator gives up this control.

iodines-team:
> @iodines-team

>It's been said earlier in this thread but it's worth repeating; the consumers of these data are journey planning and MaaS applications, not end users.

At the end, it is the same because the goal is to provide reliable data to be computed and the remaining kilometer range data uploaded from a shared car will never be guaranteed.

> The journey planners can't plan a trip based on the traveler's requirements without vehicle range as a factor. Not providing a range estimate will not change this. If they are given `current_fuel_percent` in place of `current_range_meters`, they will be forced to calculate the vehicle range using `max_range_meters` * `current_fuel_percent` because those are the only data they are given.

That is from my point of view the best computing to do, better then kilometer range data uploaded from the car

> If the vehicle operator provides the range estimate, they can do it conservatively, based on some of the factors you listed above that may affect vehicle range. If the journey planner does the range estimate, those factors probably won't be considered and the vehicle operator gives up this control.

Again, from my point of view, to complex to setup with low level reliability at the end. The best way is be the above computing.

testower:
> > @iodines-team
> 
> > It's been said earlier in this thread but it's worth repeating; the consumers of these data are journey planning and MaaS applications, not end users.
> 
> At the end, it is the same because the goal is to provide reliable data to be computed and the remaining kilometer range data uploaded from a shared car will never be guaranteed.
> 
> > The journey planners can't plan a trip based on the traveler's requirements without vehicle range as a factor. Not providing a range estimate will not change this. If they are given `current_fuel_percent` in place of `current_range_meters`, they will be forced to calculate the vehicle range using `max_range_meters` * `current_fuel_percent` because those are the only data they are given.
> 
> That is from my point of view the best computing to do, better then kilometer range data uploaded from the car
> 
> > If the vehicle operator provides the range estimate, they can do it conservatively, based on some of the factors you listed above that may affect vehicle range. If the journey planner does the range estimate, those factors probably won't be considered and the vehicle operator gives up this control.
> 
> Again, from my point of view, to complex to setup with low level reliability at the end. The best way is be the above computing.

Then the data provider should compute it and populate current_range_meters with the result.

mplsmitch:
>  the remaining kilometer range data uploaded from a shared car will never be guaranteed.

How the range is calculated is entirely up to the data provider. No where in the specification does it say that this number should come from the vehicle alone. If the best way to do this is using the above formula, that's what you should use. If you're concerned that this formula might overstate the vehicle range and lead to a bad user experience, then you should adjust the calculation by some percentage until you're comfortable with the result.

iodines-team:
> > > @iodines-team
I understand your need but you must be very carefull with this value when the vehicle is a car, in particular if it is an electric or hybrid car.

> Then the data provider should compute it and populate current_range_meters with the result.
Carsharing operators who operate electric cars (and it will be the mainstream in the future in smartcities) can not engage their responsability with this value because they doesn't know the way the car will be used. ex: Number of passengers, load, city/highway/mixte trip, external temperature estimated, ...

Operators also want to keep a minimum autonomy level of their car at the end of a trip.

But...It could be an added value to journey planner to estimate it because he should be able to cross-compute these data.

For small distances (less than 20 km), the basic range computing would be enough. For more kilometers range, journey planners have to take care. Here, it could be a differentiation between players.

NicolasFrasie:
Hi @testower, @mplsmitch,

I would like to complement the argument from @iodines-team, considering that I have also promoted this model of data. 

- The atfermarket telematics installed in vehicles aren't always able to "read" the range (km/mile) displayed on the dashboard of the vehicle, unlike the state of charge (%). And even if the range displayed on the dashboard is "read", it is influenced by many factors : the way of driving, traffic / speed, typology of the road, etc.

- The consumers of these data are free to use the data the way they want which raises an issue of liability AND end users experience, unless there is a requirement to hide the estimated range. As a mobility provider, I want to garanty the best experience possible to my customer, and I don't want to be in a situation that would mislead her/him. What would her/his experience when she/he enters into the vehicle and discovers the range calculated by the vehicle? or when the vehicle runs out of battery before arriving to destination ? 

How the range should or could be calculated raises an interesting subject. This can be a value add for a trip planner which can anticipate the trip the end user plans to do (distance, elevation, traffic, etc.), and - maybe - the way the end user drives. This can also be a value add from the operator which knows the vehicle well (age of the battery, etc), and knows the end users from previous trips whatever the MaaS / trip planner he uses. 

Until there is no available solution, I believe we should go ahead with what's available. There are already different services among carsharing or scooter sharing which publicize `current_fuel_percent` only. See 3 different examples on https://drive.google.com/drive/folders/1m6QDn905Zt0h2lPCW77WOUfl6YkT95q4?usp=sharing from Moovit in Germany with an electric shared motobike, FreeNow in France with an electric shared scooter and Transit in Canada with an electric shared car.

This proposal is a first step to model carsharing. It is unperfect and could be improved in the future. I will be happy to participate to a working group on this very subject. Based on this, I would raise concern if a member votes against this proposal and remain open to discuss any alternative.

testower:
@NicolasFrasie Thanks for taking the time to share your thoughts.

I think we could discuss this _ad nauseam_ in this thread, yet I feel the end result is the same for me: If current_range_meters is not provided it's going to be calculated with max_range_meters * current_fuel_percent, and the end result is the same: The vehicle is assumed to have some expected range based on the current fuel situation. The only difference is that the operator somehow is no longer liable and that the specification is suddenly more ambiguous and complex. I don't think the GBFS specification should be used as an instrument to limit the liability of the data providers.

I have already suggested that a solution to this right now is to remove this part from the proposal so the rest can be voted upon, and continue the discussion separately.

NicolasFrasie:
@testower, my concerns go far beyond the question of liability. 

I think you are right to worry about the integration of carsharing into a trip planner, but this is still very rare (I only know 1 case). The proposal allows to duplicate the general use case : show availability of free-floating vehicles and round-trip carsharing stations on a map (like Transit does with Communauto for example, and many local PT apps already do with the local carsharing organization).

Is it possible to exclude carsharing from the "conditionnally required" on `current_range_meters` ?

testower:
I just don't understand what other arguments than liability there is to introduce a new field which can be calculated from existing fields?

NicolasFrasie:
Hi @testower, 

As far as I know, you cannot calculate the `current_fuel_percent` from existing fields. 

The general use case for carsharing in a multimodal app is the one described in the 3 examples :   https://drive.google.com/drive/folders/1m6QDn905Zt0h2lPCW77WOUfl6YkT95q4?usp=sharing. This use case doesn't require the range, and operators (at least @iodines-team and Communauto) don't want the range to be displayed to the end user, for liability as well as customer experience.

It is needed to provide a range for trip planner calculation, but this will require a bit more work.

If carsharing is excluded from the "conditionnally required" on current_range_meters and current_fuel_percent is optionnal, would it be suitable for you @mplsmitch and @testower ?

Nicolas

testower:
Apologies for withdrawing my comment, I was unsure if it was appropriate to keep this discussion going.

You can indeed calculate current_fuel_percent from existing fields: current_range_meters / max_range_meters. Conversely, if you take the value you *would* use for current_fuel_percent, call it F, and use it to populate current_range_meters by max_range_meters * F, then the problem is solved without adding new fields.

>  operators don't want the range to be displayed to the end user

They don't have to. If a client want to display a %, they can calculate it and arrive at the same value that was used to derive the range.

So that leaves us with liability, and I don't think the spec should be used for that purpose.

> If carsharing is excluded from the "conditionnally required" on current_range_meters and current_fuel_percent is optionnal, would it be suitable for you @mplsmitch and @testower ?

No, for me, the situation doesn't change.

NicolasFrasie:
Sorry but no : `current_fuel_percent` isn't equal to `current_range_meters` / `max_range_meters`.

thbar:
Hello! I'm Thibaut, from the French national access point (colleague of @fchabouis & @Miryad3108).

After having exchanged with @NicolasFrasie from Communauto, we have modified the Pull Request: 

- The field `current_fuel_percent` has become optional (46e0b07).
- We deleted `and current_fuel_percent is not defined` from the description of `current_range_meters` (4b460ab). So the condition for current_range_meters to be mandatory is the same as in the current GBFS.

We keep in mind that not all carsharing operators can currently provide the `current_range_meters` information and that some have stated that providing this information may expose them to liabilities towards their users.

Therefore, we will allow carsharing operators to publish their data on transport.data.gouv.fr without providing `current_range_meters`. We will also customize the GBFS validator so that their data will not appear as invalid. 

Re-users will be notified that this information is missing and we will encourage carsharing operators to provide this information when they have it.

We hope that with this change, the PR is acceptable!

We also hope that after a few months of data production with this improvement, and feedback from early users, we'll be able to better understand what can be done next, but after this change, it would be a good enough step in the right direction for the French NAP.

Miryad3108:
Hello, 

Voting will be open for 10 full calendar days until 11:59 PM UTC on 9.12.2021.

Miryad - French national access point for mobility data, [Transport.data.gouv.fr](https://transport.data.gouv.fr/)

gbfs.md
```diff
@@ -723,10 +767,14 @@ Field Name | REQUIRED | Type | Defines
...
 \- `last_reported` <br/>*(added in v2.1)* | OPTIONAL | Timestamp | The last time this vehicle reported its status to the operator's backend.
-\- `current_range_meters` <br/>*(added in v2.1)* | Conditionally REQUIRED | Non-negative float | If the corresponding `vehicle_type` definition for this vehicle has a motor, then this field is REQUIRED. This value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel.
+\- `current_range_meters` <br/>*(added in v2.1)* | Conditionally REQUIRED | Non-negative float | If the corresponding `vehicle_type` definition for this vehicle has a motor, then this field is REQUIRED. This value represents the furthest distance in meters that the vehicle can travel without recharging or refueling with the vehicle's current charge or fuel. Note that the given range is indicative and can be different from the one displayed on the vehicule's dashboard.
```
testower: Spelling: vehicule -> vehicle

gbfs.md
```diff
@@ -723,10 +767,14 @@ Field Name | REQUIRED | Type | Defines
...
+\- `current_fuel_percent` <br/>*(added in vXXX)*| OPTIONAL | Non-negative float | If the corresponding vehicle_type definition for this vehicle has a motor, then this field is required. This value represents the percentage, expressed from 0 to 1 of fuel or battery power remaining with the vehicle’s current charge or fuel level.
```
testower: This should be removed for the field to be optional: "If the corresponding vehicle_type definition for this vehicle has a motor, then this field is required."

gbfs.md
fchabouis: So French :fr:

PeterCOCarShare:
On behalf of Colorado CarShare, we support this proposal!

Peter D. Krahenbuhl
CEO

cmonagle:
+1 from Transit

NicolasFrasie:
Communauto supports this proposal.

gbfs.md
```diff
@@ -723,10 +767,14 @@ Field Name | REQUIRED | Type | Defines
...
 \- `pricing_plan_id` <br/>*(added in v2.2)* | OPTIONAL | ID | The `plan_id` of the pricing plan this vehicle is eligible for as described in [system_pricing_plans.json](#system_pricing_plans.json). If this field is defined it supersedes `default_pricing_plan_id` in `vehicle_types.json`. This field SHOULD be used to override `default_pricing_plan_id` in `vehicle_types.json` to define pricing plans for individual vehicles when necessary.
+\- `vehicle_equipment`<br/>*(added in vXXX)* | OPTIONAL | Array | Description of vehicle equipment that can be provided by the operator in addition to the accessories already provided in the vehicle (field `vehicle_accessories` of vehicle_type.json) but subject to more frequent updates.<br/><br/>Current valid values are:<ul><li>`child_seat_0` _(Baby seat ("0-10kg"))_</li><li>`child_seat_1`	 _(Seat or seat extension for small children ("9-18 kg"))_</li><li>`child_seat_4`	_(Seat or seat extension for older children ("15-36 kg"))_</li><li>`winter_tires` 	_(Vehicle has tires for winter weather)_</li><li>`snow_chains`</li></ul>
```
heidiguenin: I have a question about this. I do not believe it is a big change, so it should not impact the timing of the vote but should be addressed before the vote closes.

For the child seats, I believe it should be presented as an enum, not an array. For readability, the car seat enums should be 'a', 'b', 'c' so that they are not mistaken for the number of child seats in the car. Which raises the question - how would a producer represent if there are multiple child seats?

gbfs.md
```diff
@@ -205,8 +205,12 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
...
 Example: `Asia/Tokyo`, `America/Los_Angeles` or `Africa/Cairo`.
+* Country code - Country code following the [ISO 3166-1 alpha-2 notation](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2).
 * URI *(added in v1.1)* - A fully qualified URI that includes the scheme (e.g., `com.example.android://`), and any special characters in the URI MUST be correctly escaped. See the following https://www.w3.org/Addressing/URL/4_URI_Recommentations.html for a description of how to create fully qualified URI values. Note that URIs MAY be URLs.
 * URL - A fully qualified URL that includes `http://` or `https://`, and any special characters in the URL MUST be correctly escaped. See the following https://www.w3.org/Addressing/URL/4_URI_Recommentations.html for a description of how to create fully qualified URL values.
+* Phone number - Phone number in international formatting (specification [E.164](https://en.wikipedia.org/wiki/E.164)). In particular, the phone number MUST start with a "+". The characters after the "+" MUST be numbers and MUST NOT contain any hyphen, space or parenthesis.
```
heidiguenin, line 206: With #351 open for a vote, I'd suggest removing this one to minimize merge conflicts. (They refer to the same standard.)

testower:
Entur supports this proposal

MGidel:
+1 from the City of Paris (Direction de la Voirie et des Déplacements)

gbfs.md
fchabouis: Currently `vehicle_equipment` is an array. So if there are 2 `child_seat_0` in the vehicle, I guess you can put the item twice in the array.

gbfs.md
fchabouis: Ok, you can just merge #351 first and I will resolve the resulting conflict.

heidiguenin:

iodines-team:
IODINES supports this proposal.

gbfs.md
```diff
@@ -414,9 +418,21 @@ Field Name | REQUIRED | Type | Defines
...
+\- `g_CO2_km`<br/>*(added in vXXX)* | OPTIONAL | Non-negative integer | Maximum quantity of CO2, in grams, emitted per kilometer, according to the [WLTP](https://en.wikipedia.org/wiki/Worldwide_Harmonised_Light_Vehicles_Test_Procedure).
+\- `vehicle_image`<br/>*(added in vXXX)* | OPTIONAL | URL | URL to an image that would assist the user in identifying the vehicle (e.g. logo, image of vehicle).<br /> Allowed formats: JPEG, JPG, PNG. Minimum resolution of 300 pixels per inch (ppi).
```
nbdh: 1. Aren't `JPG` and `JPEG` the same format, just both file extensions are common?
2. `300 ppi` is not a resolution, as ppi describes the pixel density. Defining an aspect ratio (or multiple allowed) could also be considered.  

> Note that pixel density is not the same as resolution, where the former describes the amount of detail on a physical surface or device, the latter describes the amount of pixel information regardless of its scale.

https://en.wikipedia.org/wiki/Pixel_density

gbfs.md
```diff
@@ -514,8 +552,11 @@ Field Name | REQUIRED | Type | Defines
...
+\-&nbsp;`is_virtual_station` <br/>*(added in v2.1)* | OPTIONAL | Boolean | Is this station a location with or without physical infrastructures ? <br /><br /> `true` - The station is a location without physical infrastructure, defined by a point (lat/lon) and/or `station_area` (below). <br /> `false` - The station consists of physical infrastructure. <br /><br /> If this field is empty, it means the station consists of physical infrastructure.<br/><br/>Physical infrastructures can be docks, charging stations, dropbox, operator branch, gate arm or parking hoops.<br><br>This field SHOULD be published in systems that have station locations without standard, internet connected physical docking infrastructure. These may be racks or geofenced areas designated for rental and/or return of vehicles. Locations that fit within this description SHOULD have the `is_virtual_station` boolean set to `true`.
```
nbdh: There is an extra space left before the question mark here.

Btw: The examples in this MR are partially not properly formatted (not matching indentation, spaces around `:` not the same as for the other attributes (`"foo": "bar"`)), I'd appreciate some unification if this gets merged.

```suggestion
\-&nbsp;`is_virtual_station` <br/>*(added in v2.1)* | OPTIONAL | Boolean | Is this station a location with or without physical infrastructures? <br /><br /> `true` - The station is a location without physical infrastructure, defined by a point (lat/lon) and/or `station_area` (below). <br /> `false` - The station consists of physical infrastructure. <br /><br /> If this field is empty, it means the station consists of physical infrastructure.<br/><br/>Physical infrastructures can be docks, charging stations, dropbox, operator branch, gate arm or parking hoops.<br><br>This field SHOULD be published in systems that have station locations without standard, internet connected physical docking infrastructure. These may be racks or geofenced areas designated for rental and/or return of vehicles. Locations that fit within this description SHOULD have the `is_virtual_station` boolean set to `true`.
```

gbfs.md
fchabouis: 1. True. We can remove JPG wich is just a possible extension of the JPEG format.
2. Would "Minimum density of 300 pixels per inch (ppi)" be correct then ?

gbfs.md
fchabouis: Thank you, I have formatted the modified json examples.

gbfs.md
nbdh: 1. Sounds reasonable.
2. That would imho make slightly more sense in a technical sense. However, the ppi are dependent from the viewing device... So stating an actual minimal resolution probably was not a bad idea, that's why I ended up thinking about aspect ratios, as you could only specify something like "The shorter edge SHOULD (or MUST?) have at least x pixels." However, the maximum pixels (of the longer edge?) might also is worth to be specified, otherwise huge files (with very high resolutions) would be allowed by the spec, but probably not desirable.

ncancelliere:
+1 from Spin. We don't do car sharing but this is not backwards breaking so we're okay with this proposal.

gbfs.md
fchabouis: We could remove the constraint on the image size / resolution for the moment and see if it's necessary to add such details afterwards.

fredpetit:
+1 from Grenoble-Alpes Métropole (mobilités)

gbfs.md
heidiguenin: @fchabouis I believe that is the right approach for now. When we added [vehicle icons and brand assets](https://github.com/NABSA/gbfs/pull/330) in v2.3-RC, we recognized that it was a first step, and other additions might be necessarily in the future, too.

gbfs.md
heidiguenin: @fchabouis Would you be ok with changing the naming from `child_seat_0`, `child_seat_1`, etc to `child_seat_a`, `child_seat_b`, etc.? To address my concern about the names being mistaken for the number of child seats?

veddan:
+1 from Move About.

heidiguenin:

Votes in favor:
Colorado CarShare (producer)
Transit (consumer)
Communauto (producer)
Entur (consumer)
City of Paris (consumer)
IODINES (producer)
Spin (producer)
Grenoble-Alpes Métropole (mobilités) (consumer)
Move About (producer)

There were no votes against. 
Thank you to everyone who took the time to review and to vote on this! We appreciate the careful review and compromise that went into the decision to add carsharing to GBFS. 

We'll update the documentation as discussed, and then we will tag and merge this into v2.3-RC2 in the coming weeks.

gbfs.md
Miryad3108: Details on the number of pixels have been removed

gbfs.md
Miryad3108: @heidiguenin The changes have been made, thank you

Miryad3108:
Hello, 

Thanks to all the people who contributed to the improvement of our initial proposal and all those who voted in favor of this Pull Request.

Are there any producers who can produce data on their carsharing offer in GBFS format in accordance with this proposal and reusers who will be able to integrate it into their services before april 2022 ? 
This will allow to integrate this proposal into the GBFS and move it from the "Release candidate" status. 
See. https://github.com/NABSA/gbfs#governance--overview-of-the-change-process

The transport.data.gouv.fr team will support potential producers and reusers if they have difficulties 


Thank you in advance,  

@testower @NicolasFrasie @iodines-team @veddan @fredpetit @ncancelliere @MGidel @cmonagle @PeterCOCarShare @kanagy @evansiroky @nbdh @mplsmitch

fchabouis:
@mplsmitch thanks for merging !
For info for the newly added fields, we put `added in vXXX`, and that will need to be replaced with the actual targeted version.

PR #363, opened 2021-09-06
Update Spin Raleigh Auto-Discovery URL in systems.csv

richtaylor-ito:
Change Spin Raleigh's Auto-Discovery URL from system_information to gbfs

systems.csv


mplsmitch:
This fix was made in #362

Issue #319, opened 2021-04-26
Extending vehicle_types.json


josee-sabourin:
### **What is the issue and _why_ is it an issue?**

In version 2.1 of GBFS, vehicle types were added to the specification to account for the increasing variety of vehicles available for on-demand hire. These vehicle types were added as part of a Minimum Viable Proposal (MVP). Since then, we have seen an increase in the need for providing more granular vehicle details to better equip travellers with the information they need for their trips. For the purpose of this needs assessment, vehicle details specific to carsharing have been omitted to avoid overlapping on [work that is already being done](https://docs.google.com/document/d/1bgNsiTcTfjKxG6khGq0ro0x-vEaToihp0_t-krGyj1o/edit?ts=606c5d87) on the subject.

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**

MobilityData has drafted a [needs assessment](http://bit.ly/vehicle-types-needs) surveying the bike, scooter, and moped industry to determine what vehicle attributes, accessories, and specifications are widely included, and promoted on provider or manufacturer websites. This issue is a means of starting a discussion on which of these features (or others we've missed!) would be beneficial to include in the next iteration of `vehicle_types.json`, and how they could be modelled. Please feel free to discuss in this thread or to comment directly in the document. 

Note: The way features are outlined in this needs assessment does not necessarily mean this is how they will be represented in the specification. For example: wheel size and seat height may be grouped into a `rider_height` field that would display the height range of riders the vehicle can accommodate. 

### **Is your potential solution a breaking change?**
- [ ] Yes 
- [x] No
- [ ] Unsure

EDIT: The extension proposal is now available [here](https://docs.google.com/document/d/1P_oDBnFvr9qzo0_5YbnrCDYptFQV9ZUOJGfi8ACD1GE/edit#).

schnuerle:
[...]

PR #370, opened 2021-09-08
Extending vehicle_types.json

josee-sabourin:

This proposal is in response to the discussion in #319 to extend vehicle_types.json. After an in-depth [needs assessment](http://bit.ly/vehicle-types-needs) and community consultation, these fields are what appear to be the priority for the GBFS community when it comes to vehicle descriptions. The proposal document can be found [here](https://docs.google.com/document/d/1P_oDBnFvr9qzo0_5YbnrCDYptFQV9ZUOJGfi8ACD1GE/edit#).  

The proposal adds enums to the `form_factor` field: 
- `cargo_bicycle`
- `scooter_seated`
- `scooter` becomes `scooter_standing`

The proposal also adds 3 fields to the specification:
- `wheel_count`: to inform the user of the number of wheels on the vehicle
- `max_permitted_speed`: to inform the user of the maximum speed allowed as well as to better inform trip planners for route planning
- `rated_power`: the motor wattage of the vehicle to better inform trip planners. 

- [x] Yes - however, the proposal can be rewritten to not include the breaking change by specifying that `form_factor.scooter` implies a standing scooter. 
- [ ] No
- [ ] Unsure

- vehicle_types.json


bartwr:
Looking good! We definitely can use this in The Netherlands, i.e. for listing cargo bikes and 'snorfietsen' (mopeds <= 25 km/h) following this updated standard.

schnuerle:
Per issue https://github.com/openmobilityfoundation/mobility-data-specification/issues/692 opened by @sven4all, for MDS 1.2.0 we _may_ be able to add `cargo_bicycle` as a vehicle type, but would not be able to remove `scooter` as this would be a breaking change. So we would have to revisit later to align these changes to MDS in the next major release 2.0.0, which will start work in the next few weeks.

sven4all:
Maybe it's a good idea for both MDS and GBFS to introduce scooter_seated and scooter_standing in the next minor release, but still allow scooter (with a will be deprecated in the next major release annotation).

josee-sabourin:

Voting will be open for 10 full calendar days until 11:59PM UTC on December 9, 2021.

kanagy:
Hi, could there be a bit more clarification on the difference between seated_scooter and moped? I read that in the motorcycle world, a scooter is more powerful than moped (e.g. need a motorcycle license in Australia for scooter, but not for moped). However, I thought in the bikeshare world (and in this spec) it was the other way round.

josee-sabourin:
Hi @kanagy 
A seated scooter would look like this: 
image

Whereas a moped would be this: 
image


We can update the documentation to be more clear if this passes.

sven4all:
+1 from CROW

cmonagle:
+1 from Transit

kanagy:
+1 From Google Maps. Yes, please clarify the documentation, particularly since this spec will be shared with car sharing (and so potentially motorbike sharing).

testower:
Entur supports this proposal, but I agree with @sven4all that this could easily be introduced as a non-breaking change, with aiming to remove "scooter" in v3.

nbdh:
+1 from nextbike

heidiguenin:

ncancelliere:
+1 from Spin

heidiguenin:

Votes in favor:
CROW (producer)
Transit (consumer)
Google Maps (consumer)
Entur (consumer)
nextbike (producer)
Spin (producer)
 
There were no votes against. 
Thank you to everyone who took the time to review and to vote on this! 

We'll update the documentation as discussed, and then we will tag and merge this into v2.3-RC2 (with the non-breaking approach discussed in the proposal) and v3.0-RC2 (with the longer-term breaking approach) in the coming weeks.

PR #393, opened 2021-11-29
Add nextbike sg

richtaylor-ito:


nbdh:
Hi @richtaylor-ito,

just saw that you manually added this system as we did not update our systems for a while. Thanks for that!

If you miss one of the nextbikes systems in the `systems.csv` feel free to ping me or @j0kan here and we'll be glad to open a MR updating all our systems asap.

PR #395, opened 2021-12-07
Update systems.csv

nbdh:
I noticed that our cronjob to update this file became a victim of a broken json schema (fixed [here](https://github.com/MobilityData/gbfs-json-schema/pull/58/commits/3ea9fac003bb4f34d2358cb3fdfed06c844128ab), thanks @isabelle-dr) and did not run properly for a while as we could not validate our feeds.

We now make two separate commits, the first one just sorts the existing systems (according to bash `sort --field-separator=',' --key=1,2 --ignore-case`, not considering the first line), the second commit updates our own systems.

I hope this makes it easier to see that (in the first commit) it is really just the order that changes. I guess sorting `systems.csv` could also be done automatically after merging it.


mplsmitch:
Beautiful - Thanks!

PR #400, opened 2022-01-11
Changes to return_types field

mplsmitch:
We’ve discovered an issue with a field that was added in v2.3RC and it's significant enough that we'd like to fix it before the v2.3RC2 and v3.0RC release candidates are completed. The issue is with the `return_type` field that describes constraints placed on the return of vehicles at the end of a rental. The field was defined as an array of enums which allowed multiple constraints to be placed on a vehicle. The problem is that the enums as defined were exclusive and not able to be combined, for example a vehicle can’t be both free floating and be required to be returned to a station. To fix this I’ve changed the field from an array of enums to a simple enum, and added an option for hybrid systems where vehicles can be returned to stations or free floating. I’ve also renamed the field from `return_type` to `return_constraint` as was suggested in #329.

We plan to call a vote on this proposal on January 18th. Please submit any feedback before then.


mplsmitch:
I hereby call a vote on this proposal. Voting will be open for 10 full calendar days until 11:59PM UTC on January 29, 2022.

cmonagle:
+1 From transit

kanagy:
+1 From Google Maps, we don't ingest it yet but it definitely makes for a less ambiguous spec.

testower:
+1 from Entur

mattroeske:
+1 from Spin

heidiguenin:

Votes in favor:
Transit (consumer)
Google Maps (consumer)
Entur (consumer)
Spin (producer)
 
There were no votes against.
Thank you to everyone who took the time to review and to vote on this! We incorporate it into v2.3-RC2, which should be ready to go in the coming week.

Issue #306, opened 2021-04-04
Discovery endpoint for GBFS urls per geography/region


dirkdk:
### About me
I'm senior engineering manager at Spin for all Partner integrations including compliance. In that role I have been on the [OMF Technology council](https://github.com/openmobilityfoundation/governance/wiki/Technology-Council) since its inception.

### **What is the issue and _why_ is it an issue?**
Consumer apps, partners and cities need to guess urls for GBFS feeds per geography (or region). Sometimes these urls are listed in https://github.com/NABSA/gbfs/blob/master/systems.csv, but this is hard to maintain and providers rarely update it themselves.

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**
Every mobility provider with a GBFS feed would have a discovery endpoint like `gbfs_geographies.json`

Data could be formatted like this:
```
{
  "last_updated": 1434054678,
  "ttl": 0,
  "version": "1.1",
  "data": {
    "countries": [
      {
        "name" : "United States",
        "iso_alpha_3": "USA", 
        "geographies": [
          {
            "name": "San Francisco",
            "url": "https://www.mobility_provider.com/gbfs/1/san_francisco/gbfs"
          },
          {
            "name":"Denver",
            "url":"https://www.mobility_provider.com/gbfs/1/denver/gbfs"
          }
        ]
      }
    ]
  }
}
```

Because GBFS urls in general contain a version, I'm wondering if we should list each individual version per geography. For Spin we don't have geographies with different versions than others. All our GBFS versions are supported for all published geographies. So I'm inclined to say no and just have 1 url per geography. If the community feels we should, then the format could be like this:

```
{
  "last_updated": 1434054678,
  "ttl": 0,
  "version": "1.1",
  "data": {
    "countries": [
      {
        "name" : "United States",
        "iso_alpha_3": "USA", 
        "geographies": [
          {
            "name": "San Francisco",
            "urls": [
              {
                "version": "2.1",
                "url":"https://www.mobility_provider.com/gbfs/2.1/san_francisco/gbfs"
              },
              {
                "version": "2.2",
                "url":"https://www.mobility_provider.com/gbfs/2.2/san_francisco/gbfs"
              }
            ]
          },
          {
            "name": "Denver",
            "urls": [
              {
                "version": "2.1",
                "url":"https://www.mobility_provider.com/gbfs/2.1/denver/gbfs"
              },
              {
                "version": "2.2",
[...]

PR #407, opened 2022-02-07
Changes to data structures

mplsmitch:
We have two fields,`vehicle_type_area_capacity` and `vehicle_type_dock_capacity` that use data from other fields as object keys. This is inconsistent with other similar fields in the specification such as `vehicle_types_available` and has caused some challenges in generating models and validation. 

This proposal changes the objects to arrays as was done with `vehicle_types _available`. I've read through the rest of the specification and haven't found this issue elsewhere. If anyone is aware other examples please chime in so we can roll them into this PR.

- [x] Yes 
- [ ] No
- [ ] Unsure


testower:
Entur supports this change

testower:
@mplsmitch How can we move this forward? Call to vote?

mplsmitch:
Hello @testower - the next step is to call a vote on this along with #454, #457 and #460 and anything else we'd like to get into the v3.0 release. In the past we've made the most progress if we group the votes an a number of issues, as opposed to holding individual votes. The PRs need to be open for comment for 7 days before a vote can be called.

I'd like to do something about #306, basically it would be like your file: https://api.entur.io/mobility/v2/gbfs/
I should be able to write it up today and get a PR open.  

  Great work on the Internationalization!

josee-sabourin:

cmonagle:
+1 from Transit

testower:
+1 from Entur

gbfs.md
```diff
@@ -625,11 +629,17 @@ Field Name | REQUIRED | Type | Defines
         "parking_type": "underground_parking",
         "parking_hoop": false,
         "contact_phone": "+33109874321",
-        "is_charging_station": true,
-        "vehicle_type_dock_capacity": {
-          "abc123": 7,
-          "def456": 9
-        }
+        "is_charging_station": "true",
```
bdferris-v2: did you mean to stringify `true` here?

gbfs.md
```diff
@@ -595,11 +595,15 @@ Field Name | REQUIRED | Type | Defines
...
-\-&nbsp;`vehicle_type_dock_capacity` <br/>*(added in v2.1)* | OPTIONAL | Object | An object used to describe the docking capacity of a station where each key is a `vehicle_type_id` as described in [vehicle_types.json](#vehicle_typesjson) and the value is a number representing the total docking points installed at this station, both available and unavailable for the specified vehicle type.
+\-&nbsp;`vehicle_type_area_capacity` <br/>*(as of v3.0)* | OPTIONAL | Array| This field's value is an array of objects containing the keys `vehicle_type_id` and `count` defined below.  These objects are used to model the parking capacity of virtual stations (defined using the `is_virtual_station` field) for each vehicle type defined in `vehicle_types.json`.
+&emsp;&emsp;\-&nbsp;`vehicle_type_id`| Conditionally REQUIRED | ID | A `vehicle_type_id`, as defined in `vehicle_types.json`, that may park at the virtual station.
```
bdferris-v2: Dumb question, but why is this "Conditionally Required"?  Under what conditions would it be left empty?  Same for `count` field below.

gbfs.md
josee-sabourin: Probably a mistake - will fix when/if we merge.

gbfs.md
josee-sabourin: They are required _**if**_ `vehicle_type_area_capacity` is used, if not, they are not required. We can clarify this when/if we merge.

bdferris-v2: What is the convention in the spec for a field that is required if its parent element is specified, where the parent element may be optional?  If I understand correctly, in this PR the required child fields are being marked as "Conditionally Required".  However, I see in PR #460 that newly added child fields are being marked simply as "Required", even if there parent element is optional.  I'll admit the latter convention seems more natural to me, but understood if that's not the convention used in the spec.

gbfs.md
bdferris-v2: Dumb question, but why not fix it now?  Isn't it better to be precise on what we are voting for?

gbfs.md
mplsmitch: Sorry for the delayed response, I've been traveling and just got back. I'll update the PR with corrections.  @bdferris-v2 you have a good point, we need to be consistent on how parent-child relationships are described throughout. I think you are correct that they should be labeled simply as required in this case because we could have parent-child relationships with other dependencies making them conditionally required.  Making these consistent throughout the spec wouldn't change the behavior of any fields so it won't require a vote. We should just go ahead and do it now so it can get rolled into the next release.

Some day I want to write a style guide that would contain all these conventions.

josee-sabourin:

gbfs.md
```diff
@@ -595,11 +595,15 @@ Field Name | REQUIRED | Type | Defines
...
+\-&nbsp;`vehicle_type_dock_capacity` <br/>*(as of v3.0)* | OPTIONAL | Array | This field's value is an array of objects containing the keys `vehicle_type_id` and `count` defined below. These objects are used to model the total docking capacity of a station, both available and unavailable, for each type of vehicle defined in `vehicle_types.json`.
+&emsp;&emsp;\-&nbsp;`vehicle_type_id`| Yes | ID | REQUIRED if `vehicle_type_id` is defined. A `vehicle_type_id`, as defined in `vehicle_types.json`, that may dock at the station.
+&emsp;&emsp;\-&nbsp;`count`| Yes | Non-negative integer | REQUIRED if `vehicle_type_id` is defined. The total number of docks at the station, both available and unavailable, that may accept the corresponding vehicle type as defined by its `vehicle_type_id`.
```
bdferris-v2: I don't think your follow-up change was quite right.  Shouldn't this be:

> REQUIRED if `vehicle_type_dock_capacity` is defined.

Both here and for the line above?

Though in generally, I think you could get away with deleting all "REQUIRED if `parent_element` is defined." statements and just be explicit about the behavior of a required field with an optional parent in the spec preamble.

mplsmitch: oops! thanks for catching that- fixed it

rootsixtysix:
+1 for harmonization of data structure from nextbike

kheraankit:
+1 form Lime

benwedge:
+1 from Joyride

cait32:
+1 from BCycle

josee-sabourin:

Votes in favor:
Transit (consumer)
Entur (consumer)
Nextbike (producer)
Lime (producer)
Joyride (producer)
BCycle (producer)
 
There were no votes against.

josee-sabourin (approved):

PR #408, opened 2022-02-08
Fix requirements of `default_pricing_plan_id` in `vehicle_types.json`

tdelmas:

Tom from @Fluctuo

The definition of  `default_pricing_plan_id` in `vehicle_types.json` require `vehicle_types.json`, which seams redondant and makes the definition unnecessarily more complicated  

Remove the redundant requirement of `vehicle_types.json` for that field of `vehicle_types.json`

- [ ] Yes 
- [x] No
- [ ] Unsure

 `vehicle_types.json`


mplsmitch:
Thanks Tom!

PR #412, opened 2022-03-02
[Format] Fix markdown toc structure

sccmcca:

Make the spec friendlier to auto generated TOCs. Improves navigability. The current heading structure of gbfs.md is messy for some markdown parsers when generating a TOC (see pic below from GitHub). This makes the spec tedious to navigate.

Screen Shot 2022-03-02 at 3 06 48 PM

This proposal reformats the heading levels so that a more navigable TOC is produced (see pic below for a sample of what this would look like in MkDocs).

Screen Shot 2022-03-02 at 3 16 01 PM

- [ ] Yes 
- [x] No
- [ ] Unsure

gbfs.md


mplsmitch:
Thanks @scmcca !

PR #415, opened 2022-03-10
Fix 2 typo in data examples

isabelle-dr:
- In the `vehicle_status.json` example: `station_id` is a type string, it needs to be enclosed in " characters
- In the `station_information.json` example, `is_charging_station` is a type boolean, no need for " characters


mplsmitch:
Thanks @isabelle-dr !

PR #421, opened 2022-03-30
correct city / state / system names in systems.csv

michalmiesiak:
Hello!
I have automatic tools that parse city names and state/country codes so as to create cities' identifiers, the tiny corrections described below would help me a lot! Thanks :)

1) Correct the state code for Pittsburgh, Pennsylvania.
This bug had been already corrected here  
https://github.com/NABSA/gbfs/pull/191/commits/e6376588fd06bc4dd6ba592f2e0357fda7b01894
then in was put back there:
https://github.com/NABSA/gbfs/commit/f40c5ded79c01bfc853036d226bb9233f6190a2c

I guess there is a cron task that puts data from an external source (and that source would have the wrong state code ?)

2. Olso --> Oslo

3. Removing some redundant spaces

4. For Warsaw, this is not a test system. This the actual Warsaw's Nextbike system.
This "-testy" suffix name was introduced here:
https://github.com/NABSA/gbfs/commit/0252f00479393c74a160108a8e5d68ebfaaaa8be
But the URL stayed the same, pointing at the actual Warsaw's system.

5. Added city names for some Nextbike systems where it was missing (ps. "łódzkie" is a region in Poland, it's written in lower case, don't know if the convention is to start with upper case anyway?)


mplsmitch:
Thanks @michalmiesiak ! NextBike runs a chron that updates their data occasionally so it's possible that these could get written over again at some point.

PR #423, opened 2022-04-01
Updating the GBFS URL for Montpellier, France

michalmiesiak:
(the former one returns wrong data)


mplsmitch:
Thanks @michalmiesiak - this one is interesting . Smoove's data is pretty low quality, for example the `system_id` is empty. But the data coming from Transport Data Gouv isn't real-time. Every station has the same `last_reported` time stamp and althought they set `ttl` to 300, it appears they only refresh it about once per hour which is useless.

AntoineAugusti:
Just saw this issue by chance, will look into that! Please report it to our team next time!

PR #437, opened 2022-06-27
inconsistencies in examples

mikael-epigram:
The examples do not seem to conform with the specification itself for some key-value pairs. 

- [ ] Yes 
- [x] No
- [ ] Unsure

gbfs.md


mplsmitch:
One thing I noticed: `lat` and `lon` are not required/necessary in that example because the vehicle's location can be derived using 'station_id`.  If they are included, they should have 6 decimals of precision i.e. xx.xxxxxx

mikael-epigram:
@mplsmitch You are right, I fixed it accordingly

mplsmitch:
Thanks @mikael-epigram , Nice Work!

PR #431, opened 2022-05-13
adding Berkeley, CA

ezmckinn:

PR #441, opened 2022-07-04
Update systems.csv

emmambd:
Updating systems.csv, including removing old feeds, adding several new ones, and fixing an inaccurate country code and auto-discovery URL for Vejer de la Frontera.

@[ezmckinn](https://github.com/ezmckinn) This PR removes LINK Berkeley based on [the error found here](https://github.com/NABSA/gbfs/pull/431#discussion_r873947667). If you could provide the correct auto-discovery URL, we can add it back!

@josee-sabourin I will leave you to add any additional feeds you find here.


mplsmitch:
Thanks @emmambd !

PR #442, opened 2022-07-14
Update systems.csv

nbdh:
We did not update our feeds for a while - we'll try to do that more often again to keep updates small and `systems.csv` as accurate and complete as possible.


mplsmitch:
Awesome - thanks @nbdh !

PR #444, opened 2022-07-29
Update systems.csv

fbouchPBSC:
Added a new bike share: Sibiu BikeCity


mplsmitch:
@fbouchPBSC  I noticed this url doesn't match the format of your others, was that intentional? what's the difference between:
publicbikesystem.net/operation/customer/gbfs/
and: 
publicbikesystem.net/customer/gbfs/

fbouchPBSC:
> @fbouchPBSC I noticed this url doesn't match the format of your others, was that intentional? what's the difference between: publicbikesystem.net/operation/customer/gbfs/ and: publicbikesystem.net/customer/gbfs/

Hi, both are valids.  The second one (shortest one) redirects to the first one.  The others will be fixed over time.

PR #446, opened 2022-08-12
Request to update system.csv

iniad-bessho:
We are working for Association for Open Data of Public Transportation in Japan.
We'd like you to add the following two feeds.

- docomo bike share service
- HELLO CYCLING


mplsmitch:
Thank you! @iniad-bessho - these are our first feeds from Japan

PR #449, opened 2022-08-30
Update to systems.CSV

fbouchPBSC:


nbdh:
Hi @mplsmitch and @josee-sabourin,

could you please have a look at this PR and eventually merge it? I'd like to update our systems and I think it would cause conflicts with this one, so to save @fbouchPBSC some work it would be best to have it merged before.

Thank you :)

Issue #353, opened 2021-08-19
free_bike_status feed lacks region_id


nbdh:
### **What is the issue and _why_ is it an issue?**
I noticed that the bikes listed in `free_bike_status` can't be associated with a certain _system region_. That means, consumers can only know to which region of a system a bike belongs as long as it is present at a station (and thus has a `station_id`). For all free-floating bikes the region is essentially unknown.

This is an issue because e.g. a consumer wants to tell how many free floating bikes are available in each of the systems regions.

Probably (without having looked at many feeds) only very few producers are using the option to fragment a system further into into regions and are publishing the optional `system_regions` feed and are putting the `region_id` into `station_information`. Does anybody have insights on how both publishers and consumers are currently using this? We at nextbike have a "brand (system) -> city (region) -> place (station) -> bike" hierarchy and even if it became more and more common to have only one city (region) per brand (system) we publish this feed and have at least one region per system.

In my opinion it would be more consistent to allow specifying the `region_id` of each bike.

I see this issue being slightly related to https://github.com/NABSA/gbfs/issues/286 where the `system_id` was added.

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**

I propose adding the `region_id` as a property of the vehicle objects listed in the `bikes` array of the `free_bike_status` feed and making it conditionally required, if the `system_regions` feed is published. Eventually it could also be left out if `station_id` is given (thus the region could be derived from the `station_information`).

I furthermore propose changing the currently _optional_ `region_id` in `station_information` feed to `Conditionally REQUIRED`, as it seems unusual to specify (publish) system regions without telling which station belongs to which region.

Alternatively `region_id` of the `free_bike_status` should just added and made _optional_. I suppose this would make the change non-breaking and thus is the preferable option?

Or are we rather encouraged to leave alone the regions and put each city into an individual system? Thus we'd probably rather comply with
> Each distinct system or geographic area in which vehicles are operated should have its own system_id
 
Perhaps 
> and I've been planning to open a PR to move this from a recommendation (should) to a requirement (must) under 3.0.

(from https://github.com/NABSA/gbfs/issues/286) even forces us to do so in the future?

But then, what are the regions actually meant for? 

> Describe regions for a system that is broken up by geographic or political region.

Not sure how "geographic areas" (system) shall be broken up. I'd appreciate some examples of how this is intended.

### **Is your potential solution a breaking change?**
- [ ] Yes 
- [ ] No
[...]

PR #450, opened 2022-09-22
Update systems.csv

nbdh:
Thanks for merging https://github.com/NABSA/gbfs/pull/449!

> This change will not add any extra work to those wanting to add a system to the repository, MobilityData will ensure to add the validation report to any new PRs before merging.

In fact, [adding the validation URLs](https://github.com/NABSA/gbfs/pull/447
) caused (a little) extra work on our side, since we basically re-generate all lines of our systems in systems.csv, and I guess nobody would not have liked to add those URLs for all our systems every time we add/update a single one. 

So let's hope the hardcoded valiation URLs are of use to anybody, since they were trivial to construct from the auto discovery URL (which is now in the systems.csv basically twice). Thanks for your work!


mplsmitch:
@nbdh I'm happy to merge this as-is for now but I noticed that many your entries in the `Location` column don't follow the definition for that field because they only contain a country code:
<html>
<body>


Field Name | REQUIRED | Definition
-- | -- | --
Country Code | Yes | ISO 3166-1 alpha-2 code designating the country where the system is located.
Name | Yes | Name of the mobility system. This MUST match the name field in system_information.json
**Location** | Yes | **Primary city in which the system is located followed by country or state.**
...

</body>
</html>

Here's an example, the first one (Praha) has only the country code, the second follows the field definition of city followed by country
Country Code | Name | Location
-- | -- | --
CZ | nextbike Praha | "CZ"
CZ | nextbike Prostejov | "Prostejov, CZ"

nbdh:
Hi @mplsmitch,

thanks for the merge and your comment.

I looked into this and realized that even if we give the city, it sometimes does not even refer to an actual city. Unfortunately I'm unable to produce a correct "primary city" automatically, nor do we have it available in all cases.

If we leave it out in the `systems.csv`, we will publish the `system_regions.json` and give the multiple cities this system is located in. (I think we already discussed in https://github.com/NABSA/gbfs/issues/353 that our approach possibly is not the best, and ideally we should treat different cities as different systems in those cases - but it would mean a bigger change on our side.)

I hope we'll find a solution somehow, but I'm afraid our next updates of `systems.csv` will still come without properly filled `location` column.

mplsmitch:
Hi @nbdh 
Do you have any idea what's going on with the Nextbike Poland feeds? It seems like 95% of them are returning 404s. Here are some examples:


https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_bp/gbfs.json
https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_pd/gbfs.json
https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_jp/gbfs.json
https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_gp/gbfs.json
https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_pv/gbfs.json
https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_pk/gbfs.json
https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_kw/gbfs.json
https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_kr/gbfs.json

Thanks!

nbdh:
Hi @mplsmitch,

that's probably because most of the systems from Nextbike Polska do a "winter break" from December till March.

Unfortunately we never were able to reflect it the right way (keep publishing the feed, but specify the break in `system_calendar.json`). Instead the cities are being kind of deactivated internally for that period, thus we (automatically) remove the feed of the whole system.

Best regards
Daniel

Issue #401, opened 2022-01-13
Add reservation price


benwedge:
I work for [Joyride](https://joyride.city/), a whitelabel solutions company enabling small operators to launch micromobility fleets around the world. Many of our clients use GBFS. 

We would like to include a reservation price in the feeds, as some of our clients charge a fee to reserve the vehicle. This appears to be common with our competitors as well based on a market scan.

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**

1. Adding this information as a flat rate to system_pricing_plans as reservation_price
2. Adding this information as a variable rate in system_pricing_plans in the same manner as per_min_pricing
3. Some combination of the two options above for flexibility
4. Adding this to vehicle_types
5. Adding this to free_bike_status

### **Is your potential solution a breaking change?**
- [ ] Yes 
- [x] No
- [ ] Unsure

mplsmitch:
Hello @benwedge - welcome, it's wonderful to have your input. This a sounds very reasonable to me although I haven't seen much reservation pricing like this in the wild. Can you point to some examples you know of, either on the Joyride platform or elsewhere?
thanks!

benwedge:
This is an example from Bird in Ottawa (non-Joyride), where they charge $0.35 per minute to reserve, with a maximum of 30 min, which suggests we should do this in the manner of per_min_pricing 

<img src="https://user-images.githubusercontent.com/11489610/149831537-eaaf0f29-4607-4091-8c70-75aa238620a9.PNG" width=33% height=33%>

This is an example from [Unlimited Biking](https://www.unlimitedbiking.com/) (Joyride client), who charges 1 USD for up to 5 minutes. This would suggest that a flat rate pricing plan is needed.

<img src="https://user-images.githubusercontent.com/11489610/149833335-2eb736a5-76bd-4ec8-9ecc-19db9fe62bcd.jpg" width=33% height=33%>

For the Bird example, the rider would be charged per minute for the reservation, then an unlock fee, then a per minute fee on the ride. With the Unlimited Biking example, there is a flat rate to reserve, an unlock fee, and then a per minute fee once the ride begins.

I can do a more comprehensive search if needed to prove how popular this may be.

mplsmitch:
[...]

PR #456, opened 2022-10-14
V2.3 final prep

benwedge:
I work for [Joyride](https://joyride.city/), a whitelabel solutions company enabling small operators to launch micromobility fleets around the world. Many of our clients use GBFS.

We would like to include a reservation price in the feeds, as some of our clients charge a fee to reserve the vehicle. This appears to be common with our competitors as well based on a market scan, as discussed in [issue 401](https://github.com/MobilityData/gbfs/issues/401).

To split reservation pricing into a per minute fee and a flat rate fee to solve for varying use cases

Field | REQUIRED | Defines
-- | -- | --
reservation_price_per_min | OPTIONAL | The cost, described per minute, to reserve the vehicle prior to beginning a rental. This amount is charged for each minute of the vehicle reservation until the rental is initiated, or until the number of minutes defined in vehicle_types.json#default_reserve_time elapses, whichever comes first. When using this field, you MUST declare a value in vehicle_types.json#default_reserve_time. This field MUST NOT be combined in a single pricing plan with reservation_price_flat_rate.
reservation_price_flat_rate | OPTIONAL | The cost, described as a flat rate, to reserve the vehicle prior to beginning a rental. This amount is charged once to reserve the vehicle for the duration of the time defined by vehicle_types/json#default_reserve_time. When using this field, you MUST declare a value in vehicle_types/json#default_reserve_time.  This field  MUST NOT be combined in a single pricing plan with reservation_price_per_min.

To be added to: [system_pricing_plans.json](https://github.com/NABSA/gbfs/blob/master/gbfs.md#system_pricing_plansjson)

- [ ] Yes 
- [x] No
- [ ] Unsure

[system_pricing_plans.json](https://github.com/NABSA/gbfs/blob/master/gbfs.md#system_pricing_plansjson)


josee-sabourin:
Closing this as discussed in #401. All new discussion should be moved to #457

PR #460, opened 2022-10-28
Internationalization of free-form text

testower:

I am Tom Erik Støwer representing Entur AS.

This proposal solves several problems with internationalization of free-form text, see #262. In summary, this proposal removes the need to duplicate the feeds once for each supported language, and instead "inlines" translations within the free-form text fields themselves.

In essence, the proposal turns free-form text fields (text that is intended for humans to read), from a simple String field, into an array of objects, with one object per supported language. The object will have the keys `text` and `language`, where `text` is the translated string, and `language` is the IETF BCP 47 language code for the translation (i.e. the `Language`).

~~The fields `make` and `model` in `vehicle_types.json` are up for debate. I have included the field `color` as these seem uncontroversial (e.g. the examples are clearly in the English language), whereas make and model may not be translateable at all.~~

In addition, URLs that link to human readable content are subject to the same change, so that URLs can link to different translations. I have not included URLs that link to content intended for data consumers, and not end users.

The `gbfs.json` is modified so that the `data` field has a single field `feeds`, instead of one feed field per supported language (with the language as the key).

The `language` field in `system_information.json` is changed to `languages` - an array of supported languages.

The localization section is changed so that it describes the new pattern.

- [x] Yes 
- [ ] No
- [ ] Unsure

- `gbfs.json`
- `system_information.json`
- `vehicle_types.json`
- `station_information.json`
- `system_regions.json`
- `system_pricing_plans.json`
- `system_alerts.json`
- `geofencing_zones.json`

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -139,10 +139,9 @@ Publishers SHOULD implement auto-discovery of GBFS feeds by linking to the locat
 
 ### Localization
 
-* Each set of data files SHOULD be distributed in a single language as defined in `system_information.json`.
-* A system that wants to publish feeds in multiple languages SHOULD do so by publishing multiple distributions, such as:
-    * `https://www.example.com/data/en/system_information.json`
-    * `https://www.example.com/data/fr/system_information.json`
+* Each supported language MUST be listed in the `languages` field in `system_information.json`.
+* Translations MUST be provided for each supported language for all translateable fields.
+* URLs pointing to text intended for consumption by end-users MUST be provided for each supported language.
 
 ### Text Fields and Naming
 
@@ -178,7 +177,7 @@ Decimal places | Degrees | Distance at the Equator
 
 ### Data Latency
 
-The data returned by the near-realtime endpoints `station_status.json` and `vehicle_status.json` SHOULD be as close to realtime as possible, but in no case should it be more than 5 minutes out-of-date.  Appropriate values SHOULD be set using the `ttl` property for each endpoint based on how often the data in feeds are refreshed or updated. For near-realtime endpoints where the data should always be refreshed, the `ttl` value SHOULD be `0`. The`last_updated` timestamp represents the publisher's knowledge of the current state of the system at this point in time. The `last_reported` timestamp represents the last time a station or vehicle reported its status to the operator's backend.
+The data returned by the near-realtime endpoints `station_status.json` and `vehicle_status.json` SHOULD be as close to realtime as possible, but in no case should it be more than 5 minutes out-of-date.  Appropriate values SHOULD be set using the `ttl` property for each endpoint based on how often the data in feeds are refreshed or updated. For near-realtime endpoints where the data should always be refreshed, the `ttl` value SHOULD be `0`. The `last_updated` timestamp represents the publisher's knowledge of the current state of the system at this point in time. The `last_reported` timestamp represents the last time a station or vehicle reported its status to the operator's backend.
 
 ## Licensing
 
@@ -187,9 +186,9 @@ It is RECOMMENDED that all GBFS data sets be offered under an open data license.
 ## Field Types
 
 * Array - A JSON element consisting of an ordered sequence of zero or more values.
-* Boolean - One of two possible values, `true`or `false`. Boolean values MUST be JSON booleans, not strings (meaning `true` or `false`, not `"true"` or `"false"`). *(as of v2.0)*
+* Boolean - One of two possible values, `true` or `false`. Boolean values MUST be JSON booleans, not strings (meaning `true` or `false`, not `"true"` or `"false"`). *(as of v2.0)*
 * Country code - Country code following the [ISO 3166-1 alpha-2 notation](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2).
-* Date - A date in the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) Complete Date Extended Format: YYYY-MM-DD . Example: `2019-09-13` for September 13th, 2019.
+* Date - A date in the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) Complete Date Extended Format: YYYY-MM-DD. Example: `2019-09-13` for September 13th, 2019.
 * Datetime *(added in v2.3-RC2)*- Combination of a date and a time following [ISO 8601 notation](https://www.iso.org/iso-8601-date-and-time-format.html). Attributes : [year](https://docs.python.org/3/library/datetime.html#datetime.datetime.year), [month](https://docs.python.org/3/library/datetime.html#datetime.datetime.month), [day](https://docs.python.org/3/library/datetime.html#datetime.datetime.day), [hour](https://docs.python.org/3/library/datetime.html#datetime.datetime.hour), [minute](https://docs.python.org/3/library/datetime.html#datetime.datetime.minute), [second](https://docs.python.org/3/library/datetime.html#datetime.datetime.second), and timezone.
 * Email - An email address. Example: `example@example.com`
 * Enum (Enumerable values) - An option from a set of predefined constants in the "Defines" column. Enum values SHOULD be lowercase.
@@ -216,7 +215,6 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
 Example: `Asia/Tokyo`, `America/Los_Angeles` or `Africa/Cairo`.
 * URI - A fully qualified URI that includes the scheme (for example, `com.example.android://`). Any special characters in the URI MUST be correctly escaped. See the following https://www.w3.org/Addressing/URL/4_URI_Recommentations.html for a description of how to create fully qualified URI values. Note that URIs MAY be URLs.
 * URL - A fully qualified URL that includes `http://` or `https://`. Any special characters in the URL MUST be correctly escaped. See the following https://www.w3.org/Addressing/URL/4_URI_Recommentations.html for a description of how to create fully qualified URL values.
-
 
 ### Extensions Outside of the Specification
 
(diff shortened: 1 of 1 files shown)


hannesj:
> The fields `make` and `model` in `vehicle_types.json` are up for debate. I have included the field `color` as these seem uncontroversial (e.g. the examples are clearly in the English language), whereas make and model may not be translateable at all.

They might be represented in a different writing system, eg Latin and Cyrillic scrips.

josee-sabourin:

cmonagle:
+1 from Transit, thanks @testower

testower:
+1 from Entur (not sure my vote counts since its my proposal)

gbfs.md
```diff
@@ -388,10 +383,25 @@ Field Name | REQUIRED | Type | Defines
   "version": "3.0",
   "data": {
     "system_id": "example_cityname",
-    "language": "en",
-    "name": "Example Bike Rental",
-    "short_name": "Example Bike",
-    "operator": "Example Sharing, Inc",
+    "languages": ["en"],
```
bdferris-v2, line 386: Due to a limitation of GitHub, I can't seem to comment on unmodified text that appears in collapsed sections of the diff, but...

There are still a couple of example `system_information.json` snippets in the file that still reference the old singular "language" field (e.g line 253 in the updated file).  Might be good to update them as well?

gbfs.md
```diff
@@ -139,10 +139,9 @@ Publishers SHOULD implement auto-discovery of GBFS feeds by linking to the locat
 
 ### Localization
 
-* Each set of data files SHOULD be distributed in a single language as defined in `system_information.json`.
-* A system that wants to publish feeds in multiple languages SHOULD do so by publishing multiple distributions, such as:
-    * `https://www.example.com/data/en/system_information.json`
-    * `https://www.example.com/data/fr/system_information.json`
+* Each supported language MUST be listed in the `languages` field in `system_information.json`.
+* Translations SHOULD be provided for each supported language for all translateable fields.
```
bdferris-v2: Because supported languages are a MUST but individual translations are a SHOULD, do you have any thoughts on preferred behavior when a string translation is missing for a particular language?  Similarly, should data be considered invalid if a translation is provided for an unsupported language?

gbfs.md
```diff
@@ -345,11 +328,17 @@ The following fields are all attributes within the main `data` object for this f
...
-`language` | Yes | Language | The language that will be used throughout the rest of the files. It MUST match the value in the [gbfs.json](#gbfsjson) file.
-`name` | Yes | String | Name of the system to be displayed to customers.
-`opening_hours`<br/>*(added in v3.0-RC)*| Yes | String | Hours and dates of operation for the system in [OSM opening_hours](https://wiki.openstreetmap.org/wiki/Key:opening_hours) format. *(added in v3.0-RC)*
-`short_name` | OPTIONAL | String | OPTIONAL abbreviation for a system.
-`operator` | OPTIONAL | String | Name of the system operator.
+`languages` | Yes | Array | List of languages used in translated strings. Each element in the list must be of type Language.
+`name` | Yes | Array | Name of the system to be displayed to customers. An array with one object per supported language with the following keys:
+\-&nbsp; `text` | Yes | String | The translated text
+\-&nbsp; `language` | Yes | Language | IETF BCP 47 language code
```
bdferris-v2, line 334: It is kind of a bummer that you have to repeat the same text + language snippet for every translated field.  In the spirit of DRY, would you consider introduced an `Array<Localized String>` type (or something equivalent) and defined the fields of `Localized String` above in the "Field types" section?

gbfs.md
testower: Good catch, I can see some examples still need to be updated.

gbfs.md
testower: I agree. This should be a MUST as well.

gbfs.md
testower: Yeah, we went a few rounds on this but ended up going with what seems to be the overall style of the spec. This could always be changed without it being breaking.

gbfs.md
testower: "Similarly, should data be considered invalid if a translation is provided for an unsupported language?" not sure if that is necessary - what purpose would it serve to specify this?

testower:
I have updated the PR with two changes of SHOULD to MUST in the Localization section, as well as corrected a few examples that I missed previously.

josee-sabourin:

gbfs.md
bdferris-v2: The purpose would be to make it explicit that there should be 1:1 mapping between values in the `languages` field in `system_information.json` and individual localized strings in translatable fields.  I think changing this to MUST helps, but I'd go further (see comment below)

gbfs.md
bdferris-v2: I wanted to recommend that you also update `language` description to say:

> An IETF BCP 47 language code that MUST match one of values specified in the `languages` field in `system_information.json`.

However, it seems silly to repeatedly duplicate that guidance for every translatable field.

testower: Yes I agree. I have played around with some varieties of this but all felt clunky to me and kind of collided with the style of the rest of the spec. However, if you have a concrete proposal on how this would look would you mind drafting it up?

testower: E.g. the notation `Array<Type>` is not used elsewhere in the spec

testower:
It would be a shame to make a new major version without this change in, as it would have to wait for a v4 then, and the state of internationalization now is not great. Can the voting period be extended until we are all happy with this proposal?

rootsixtysix:
We discussed this proposal at nextbike by TIER and came up with the following pros and cons:

- (+) this would reduce some backend-to-backend-traffic for big aggregators/consumers
- (-) small consumers might prefer a single feed/dataset in one language
- (-) the feed would become harder to consume. You already need some sort of programming / processing (e.g it wouldn’t work anymore to simply use the dot-operator as data.vehicle_types[0].name)

In general we don’t see that the advantages are big enough for a breaking change. However we would implement the new standard. So no vote / abstentation on this proposal.

josee-sabourin:
@testower as the advocate you can choose to cancel the vote and continue the discussion until you feel the proposal is in a good place and consensus has been reached. Once at that point you can re-open the vote. There have been changes pending for v3 for some time now, waiting an extra week or so for this to be resolved will not be a problem. Additionally it would allow us to open the vote on 474 for v3 as well.

bdferris-v2:
+1 from Google

All my word-smithing aside, I think this is a more natural data representation that avoids duplicating the same semantic data cross multiple localized feeds.

To the comments from @rootsixtysix, I think dot-operator syntax would still work for a valid feed.  Namely, if a string field was already a required field, then there must be a translation specified (otherwise field is invalid), so you could safely reference the first element for a single-language feed.  For multiple language feeds, maybe we could add a "The primary language of the feed should be specified first"?

gbfs.md
```diff
@@ -345,11 +328,17 @@ The following fields are all attributes within the main `data` object for this f
...
-`language` | Yes | Language | The language that will be used throughout the rest of the files. It MUST match the value in the [gbfs.json](#gbfsjson) file.
-`name` | Yes | String | Name of the system to be displayed to customers.
-`opening_hours`<br/>*(added in v3.0-RC)*| Yes | String | Hours and dates of operation for the system in [OSM opening_hours](https://wiki.openstreetmap.org/wiki/Key:opening_hours) format. *(added in v3.0-RC)*
-`short_name` | OPTIONAL | String | OPTIONAL abbreviation for a system.
-`operator` | OPTIONAL | String | Name of the system operator.
+`languages` | Yes | Array | List of languages used in translated strings. Each element in the list must be of type Language.
```
bdferris-v2, line 331: This field is Required (that's good!), but does that imply that the array must have at least one value?

gbfs.md
bdferris-v2: Here's an example of what the syntax could look like:

https://github.com/MobilityData/gbfs/compare/master...bdferris-v2:gbfs:localized-string?expand=1
https://github.com/bdferris-v2/gbfs/blob/2b98e0280279ce705de746a3366ebf6795209dfc/gbfs.md

gbfs.md
testower: That is a very valid point, since "Array" is defined in the spec as "A JSON element consisting of an ordered sequence of zero or more values". I'm curious if that's intentional. Is there any field in the spec that would carry different meanings for an empty array as opposed to omitting the field altogether? If not then maybe the right thing to do is to require (optional?) arrays to have at least one element?

gbfs.md
testower: I'm not against this, but it will take a while to update the proposal - and the vote may have to be postponed.

@mplsmitch what do you think about this style? It certainly makes things much less verbose.

gbfs.md
mplsmitch: Maybe this could be clarified in the `languages` array definition but wouldn't there always be at least one value in the array? For example if you don't offer any translations and only offer feeds in `en` then that's the array value.

testower: Yes exactly - there should always be at least one value, but the spec leaves this open. I.e. providing an empty array would still be valid according to the spec.

benwedge:
This looks like a wise proposal without creating too much work for single-language fields, so +1 from Joyride

gbfs.md
testower: On the other hand - changing this separately will not be breaking...

mplsmitch: Less verbose is better but maybe we should try to pass this as-is, and then open a separate vote on changing to the localized string format. It seems like it would be an easy sell with some example JSON showing how much lighter it would be.

futuretap:
+1 from whereto.app

ezmckinn:
+1 from Superpedestrian as well.

josee-sabourin:

Votes in favor:
Transit (consumer)
Google Maps (consumer)
Superpedestrian (producer)
Joyride (producer)

Abstentions: 
Nextbike (producer)
 
There were no votes against.

josee-sabourin (approved):

PR #471, opened 2022-12-07
adds van as vehicle type

teebot:
Note: If your PR is to add, remove, or correct a system in systems.csv, please delete this template. 

Hi, 
I'm Thibaut from Poppy Mobility. 

We're a shared mobility company offering shared cars, bikes, scooters and also vans. This latter category is not in vehicle types. I think it deserves its own type as it's quite a different form factor and usage than cars. 

- [ ] Yes 
- [x] No
- [ ] Unsure

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -436,7 +436,7 @@ Field Name | REQUIRED | Type | Defines
 ---|---|---|---
 `vehicle_types` | Yes | Array | Array that contains one object per vehicle type in the system as defined below.
 \- `vehicle_type_id` | Yes | ID | Unique identifier of a vehicle type. See [Field Types](#field-types) above for ID field requirements.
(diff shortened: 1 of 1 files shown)


mplsmitch:
@teebot I think it would be helpful to distinguish cargo vans from passenger vans. It looks like what Poppy offers are cargo vans.
 This could  be done by simply adding an enum for each:

- `van_cargo`
- `van_passenger`

NicolasFrasie:
I agree that vehicles designed for the carriage of passengers and vehicles designed for the carriage of goods should be distinguished. This refers to M and N vehicle types defined in EU classification : https://alternative-fuels-observatory.ec.europa.eu/general-information/vehicle-types. 

I think "utility vehicle" would more adequate than "van".

However, it doesn't apply the same for bike as a cargo bike can generally move people as well as things (see Cargoroo https://cargoroo.nl/en/ or Cambio https://www.cambio.be/fr-bxl/e-cargo-bike or Citiz https://lpa.citiz.coop/velos-cargo).

josee-sabourin:
Hi Thibaut! We were hoping to call on a vote on this today since there is interest, any chance you would be able to re-open this PR?

Issue #404, opened 2022-01-27
Geofencing refactor


mplsmitch:
MobilityData has been working on a refactor of the `geofencing_zones.json` file. The geofencing endpoint was added to the specification starting with v2.1 in early 2020. Since that time we've heard that it's has been a source of confusion for both data producers and consumers. Before opening a PR, we'd like to solicit feedback on a proposed solution. Because this has been a difficult topic for people to get their heads around, we've outlined our proposal in this explainer video.  Please have a look and let us know what you think.

https://user-images.githubusercontent.com/15235861/151430511-eb360e3c-9bbf-4dfd-8955-362db917e3c7.mp4

If there's general agreement to go forward with this approach, we'll open a PR and try to roll it into v3.0. If that's not that case we'll go ahead with the v3.0 release as-is and push the refactor of geofencing to a later release.

### **Is your potential solution a breaking change?**
- [x] Yes 
- [ ] No
- [ ] Unsure

mtborges:
I would like to offer a few suggestions to for the Geofencing refactor effort. Below you'll find our ideas, please feel free to ask for more details if needed.

1. **Geofence Zones assumptions:** currently the specification states that `"By default, no restrictions apply everywhere. Geofencing zones SHOULD be modeled according to restrictions rather than allowance"`. This creates situations where data producers are forced to create zones where rules apply to the external area of their polygons, for instance when defining their zone of operation. We would like to suggest that if geofencing_zones.json is present on the GBFS feed, all areas not defined by the feed are considered disallowed. This would allow providers to define their operational area first and then restrictions that might apply and it is needed for the next suggestion.

1. **Remove polygons applying rules to external areas:** we would also like to suggest that all zones defined have rules that apply to the internal area of their polygons. Allowing rules to apply to the exterior area of a polygon makes it harder for consumers to properly use the data, creates confusion to data producers and also creates edge cases that can break the geofence zone definition.

   - One such edge case can be seen on the example shown on the slide at the 5:35 min mark on the video. The service area defined here allows for users to ride vehicles outside the defined geofence (not park, just ride), which makes it possible for users to go outside the provider business area by any distance (I could potentially ride to a different city and come back) and also allows users to potentially break ride through rules in a neighbor city (since a user could potentially get a vehicle from Rotterdam and ride to Hague - which is close by - where the ride through rules might simply not be defined by the provider, since it's outside the Rotterdam service area).

[...]

PR #481, opened 2023-01-23
Geofencing refactor

bdferris-v2:
Hey all, a bit late to the party, but I'm Brian Ferris at Google.  I've got a long history in the GTFS space, but I've been recently asked to help out on the GBFS side of things.  With that in mind, I figured my first PR should be small one that is unlikely to cause much discussion :) ...

The Geofencing refactor from #404.

As discussed in #404 and captured in discussions at other forums (e.g. MobilityData developer workshop last summer), there have been a number of pain points around the existing `geofencing_zones.json` schema mechanism.  Specifically, the combination of interior/exterior polygon semantics, overlapping polygons, and additive rules led to confusion amongst both producers and consumers of data about what restrictions actually applied in a given area.

This proposal attempts to capture the sentiment of various stakeholders, including those from the [developers workshop](https://github.com/MobilityData/gbfs/issues/404#issuecomment-1147796880), with the following high-level changes:

* Polygons should define the interior (never the exterior) for a given restriction.
* If geofencing is not present, there are no restrictions. If geofencing *is* present, rides are restricted by default any only enabled explicitly via defined geofence zone.
* If ride-through is allowed outside the main service zone, the provider should provide a polygon corresponding with the maximum possible reach of the vehicle.
* In the case of overlapping polygons, rules are generally not additive, but instead strict rules precedence is defined.  (Note that there is still complexity on this point with optional rules fields.  Feedback appreciated).

- [X] Yes 
- [ ] No
- [ ] Unsure

`geofencing_zones.json`

I know the GBFS 3.0 release process is well underway, but given that amount of discussion geofencing has received in the past year from ecosystem stakeholders, it seems worth trying to get this in the 3.0 release considering it is a breaking change.


gbfs.md
```diff
@@ -699,10 +699,11 @@ Field Name | REQUIRED | Type | Defines
...
 \-&nbsp;`parking_type` <br/>*(added in v2.3)* | OPTIONAL | Enum | Type of parking station.<br /><br />Current valid values are:<ul><li>`parking_lot` _(Off-street parking lot)_</li><li>`street_parking` _(Curbside parking)_</li><li>`underground_parking` _(Parking that is below street level, station may be non-communicating)_</li><li>`sidewalk_parking` _(Park vehicle on sidewalk, out of the pedestrian right of way)_</li><li>`other`</li></ul>
 \-&nbsp;`parking_hoop`<br/>*(added in v2.3)* | OPTIONAL | Boolean | Are parking hoops present at this station?<br /><br />`true` - Parking hoops are present at this station.<br />`false` - Parking hoops are not present at this station.<br /><br />Parking hoops are lockable devices that are used to secure a parking space to prevent parking of unauthorized vehicles.
 \-&nbsp;`contact_phone`<br/>*(added in v2.3)* | OPTIONAL | Phone number | Contact phone of the station.
+>>>>>>> master
```
schnuerle: May need to clean this line up from a prior conflict.

bdferris-v2: fixed, thanks!

mplsmitch:
A few questions and some comments:

>If ride-through is allowed outside the main service zone, the provider should provide a polygon corresponding with the maximum possible reach of the vehicle.

I think I understand the problem you’re trying to solve with this but I’m still a little skeptical of the maximum possible reach measure. If the vehicle is electrified, the “_maximum possible reach_” could be a factor of the vehicle’s range contained in `max_range_meters` but how would this be done for a vehicle that's not range limited like a pedal powered bike, or a shared car that could travel hundreds of miles from where the trip began? 

If an operator wanted to prevent users from ending a trip outside of the city where they are permitted to operate they might put a `ride_end_allowed`=`false` rule on the outer “_maximum possible reach_” polygon. If that polygon intersected with another service where the rule should not be applied would that cause a problem for you? It seems like it creates the same problem as applying rules to the outside of polygons.

If there’s an area of limitation within a service area like `ride_through_allowed`=`false`, should this be done by creating a hole in the service area polygon where rides are permitted (I think that’s in the current example JSON) or should it be done with a separate overlapping polygon earlier in the JSON, before the service area polygon? 


Some other stuff:

- The definition of `geometry` is written in terms of limitations but your new scheme depends on allowances so maybe the `geometry` definition should be changed to reflect that?

- The first sentence in the `rules` definition is repeated

- The definitions for `ride_start_allowed` and `ride_end_allowed` both refer exclusively to Undocked (“free floating”) vehicles. Since these are now REQUIRED if the `rules` array is defined it seems like they need to apply to all vehicles, both dockless and docked (station based).

bdferris-v2:
Paraphrasing:

> How should an operator model a maximum-possible-reach zone when the effective range of a vehicle (electric bike, car) could be quite large?

I'll admit this will probably be specific to each operator and could be influenced by a number of factors: vehicle range, operating licenses, restrictions on insurance, other factors.  I wouldn't want to be too prescriptive on how operators model this, as it's ultimately their call.  That said, I'd love to see some specific comments from operators if they have feedback on this.

> What happens when service area polygons from different systems overlap?

So there are a couple of dimensions to this.

1) Perhaps it's worth adding this explicitly to the text of the spec, but for two separate systems defined in separate feeds, the geofences of the first system should never be applied to the second system, whether they have overlap or not.  Put another way, geofences are scoped to a single feed.

2) Things are more complicated for aggregated feeds, where multiple effective systems are combined into a single feed.  Here, we acknowledge the possibility that two separate systems from nearby cities or regions could potentially have overlapping geofences.  However, in practice, we believe this overlap is rare.  Second, in practice, many of the aggregated feeds we have seen use separate vehicle types to distinguish the vehicles of one system from another.  Based on our resolution rules proposed in this PR, polygons with separate vehicle types would not cause interactions with each other.

That said, if this becomes an issue down the road (e.g. aggregator doesn’t want to specify separate vehicle types), I can imagine a number of approaches for adding system-level scoping of geofence polygons that would not break backwards-compatibility for existing feed producers.

> Should restricted areas within a service area be modeled as “holes” in the service area polygon or as individual polygons that overlap (and come before) the service area polygon?

Because we are using GeoJSON as our polygon definition spec, I think we are on the hook for supporting "hole" functionality (unless we explicitly forbid it?).  That said, as a best practice, I think we’d encourage providers to model restricted areas as separate polygons for clarity.  That should could be called out in the spec as a SHOULD if you think it's appropriate.

> Other stuff

Fixed as noted.  Though I would point out that the `ride_start_allowed` and `ride_end_allowed` are just a refactor of the existing `ride_allowed` field for clarity, which was already REQUIRED.  However, agreed that it makes sense to remove the reference to a specific vehicle type.

ezmckinn:
Thanks for this detailed proposal, @bdferris-v2! A few thoughts:

To me, the idea of `maximum_possible_reach` introduces more confusion than clarity. @mplsmitch  interpreted this as the vehicle's maximum possible range — but did you mean it as the boundary at which a vehicle will be physically stopped? Either way, it would have different implications across modes (i.e. a car will have longer range than a scooter, and also a car  driver may just be beeped at when they exit the zone, whereas a scooter rider would slow to a stop). I think it also runs contrary to the spirit of the law; most cities intend their operational zones to be the maximum outer reach, and in my view, formalizing a buffer zone invites operators to exceed that.

Clarifying that geofences should always define the inner polygon, rather than the exterior, seems fine.

I think it's fine to assume that if there are no geofences, there are no restrictions. However, I have concerns about this provision: "If geofencing is present, rides are restricted by default any only enabled explicitly via defined geofence zone."  

Mostly, this latter clause would be difficult to implement. If a MaaS platform were to adopt this assumption before all the providers had updated their geofences to include an explicit "ride allowed" base layer, then large areas of many markets would suddenly appear un-rideable to riders.  It would likely worsen the overlapping geofences problem, as it would only work if the "ride allowed" operations zone were the lowest layer, with all the other more restricted layers on top of it. Lastly, it would lead to cluttered user interfaces; would Google Maps overlay a colored polygon to show riders they can travel in a certain zone? If not, then what's the value of making that explicit through a geofence, rather than just leaving it as an open area? Riders would have to start a trip from operators' native app, anyway, where the operational zone would be displayed.

To my mind the simplest thing is to keep the assumption of "if there's no geofence here, you can ride" — regardless of how many geofences the system has as a whole.

bdferris-v2:
Thanks for the comments and questions @ezmckinn.  Let me respond with more comments and more questions:

> To me, the idea of maximum_possible_reach introduces more confusion than clarity.

I think it's worth pointing out that the spec as written makes no mention of `maximum_possible_reach` as an explicit concept.  Using the old spec, I believe operators could define polygons for both an operating zone and a larger maximum-possible-reach zone, and they can continue to do so with the new spec if they so choose.  The main distinction is that hopefully they are less likely to implement it incorrectly with the new semantics :)

The question of whether providers should be specifying maximum-possible-reach zones at all seems like more of a policy / best-practices question.  I'll concede it's an important question for the community to address, but I also wonder if our thoughts on that subject actual change the details of this proposal.  I'm not saying that in attempt to derail the discussion, just genuinely curious if you have specific thoughts.  But maybe you sort of hinted at your stance with your next question.

> However, I have concerns about this provision: "If geofencing is present, rides are restricted by default..."

Let's get into it :)

> If a MaaS platform were to adopt this assumption before all the providers had updated their geofences to include an explicit "ride allowed" base layer, then large areas of many markets would suddenly appear un-rideable to riders.

Conceivably, a provider would only get the new semantics if they were specifying a new GBFS v3 feed (assuming this proposal makes it), which will likely require other refactoring for geofencing and other breaking changes.  I'm hoping it wouldn't take them by surprise.  Critically, we aren't proposing to apply these changes to existing v2 feeds.  And yes, as a data consumer, you'll have to figure out how to potentially combine v2 and v3 geofence feeds in the same system.  I acknowledge that technical burden, though I hope it doesn't stop us from trying to improve the spec.

> It would likely worsen the overlapping geofences problem, as it would only work if the "ride allowed" operations zone were the lowest layer, with all the other more restricted layers on top of it.

There is no way around it.  Providers /will/ need to take care in how they order their polygons in order to get the behavior they desire.  However, we think the polygon ordering semantics are clearer and less error-prone than the status quo of the existing geofence spec.

> Lastly, it would lead to cluttered user interfaces; would Google Maps overlay a colored polygon to show riders they can travel in a certain zone?

I can't comment too much on what Google might or might not do from a user experience (prepare to be underwhelmed either way :), but I do have some general thoughts.

First, I hope you'd agree that whether we specify allow-by-default or restrict-by-default, providers could specify a large polygon (as large a city, state, country, continent, globe) specifying the alternate behavior (restrict in allow-by-default, allow in restrict-by-default), so an app needs to be prepared to handle either scenario in terms of map rendering (or not rendering), regardless of whether they decide to color areas where you can or cannot ride (or both).  In that sense, I don't think the spec choice here changes the complexity of the display problem for data consumers.

Second, if you agree with me that the rendering challenges are the same (maybe not?), then for me it comes down to what is the more reasonable default?  Namely, are there more operators who want an allow-by-default policy (such that they have to specify an explicit restrict polygon as a base if they want to override) or who want restrict-by-default (such that they have to specify a base polygon for their operational area).  My hunch is that it's the latter (restrict-by-default).

gbfs.md
```diff
@@ -1310,30 +1310,36 @@ Field Name | REQUIRED | Type | Defines
...
+
+The following fields are all attributes within the main `data` object for this feed.
 
 Field Name | REQUIRED | Type | Defines
 ---|---|---|---
 `geofencing_zones` | Yes | GeoJSON FeatureCollection | Each geofenced zone and its associated rules and attributes is described as an object within the array of features, as follows.
 \-&nbsp;`type` | Yes | String | “FeatureCollection” (as per IETF [RFC 7946](https://tools.ietf.org/html/rfc7946#section-3.3)).
 \-&nbsp;`features` | Yes | Array | Array of objects as defined below.
 &emsp;\-&nbsp;`type` | Yes | String | “Feature” (as per IETF [RFC 7946](https://tools.ietf.org/html/rfc7946#section-3.3)).
-&emsp;\-&nbsp;`geometry` | Yes | GeoJSON MultiPolygon | A polygon that describes where rides might not be able to start, end, go through, or have other limitations. A clockwise arrangement of points defines the area enclosed by the polygon, while a counterclockwise order defines the area outside the polygon ([right-hand rule](https://tools.ietf.org/html/rfc7946#section-3.1.6)). All geofencing zones contained in this list are public (meaning they can be displayed on a map for public use).
```
leonardehrenfried, line 1322: I'm very much in favour of dropping the clockwise/counterclockwise semantics. They are confusing and I've not actually seen them in the wild.

leonardehrenfried:
I have spent the last few weeks implementing the current geofencing spec in OpenTripPlanner so I feel qualified to weigh in.

Some comments:

- I like the fact that you're proposing to remove the hole semantics. The geometry order is a better, easier to understand solution.
- I like the fact that the presence of geofencing zones makes areas outside the zones unusable. This is something that  I am currently inferring from the available data. In fact it addresses the important question of how to model a general operating area. Right now you'd probably have to use the difficult to understand clockwise/counterclockwise semantics or create a "ring of fire" . What the feeds that I looked at did instead though was to have a large area with ride_allowed=true and ride_through_allowed=true. I interpreted this as "don't allow leaving this area" which is contrary to the spec which says "By default, no restrictions apply everywhere".

In summary, I think what you're proposing is what my implementation assumes the spec _should_ say by looking at the available data.

ezmckinn:
[bdferris-v2](https://github.com/bdferris-v2) Thanks for the follow-up! It's helpful to understand your thinking in a bit more detail. You raise a good point about operators defining what version they are providing (e.g. `v3`), and also that there will be display issues whether we use "allowed" or "restricted" as the default setting. I'd be curious to know what other operators on this thread think. I'm certainly willing to be convinced either way.

mplsmitch:
I was an operator of both docked and dockless bikes from 2010 to 2019. During that time we never had any limitations on where a bike could travel. Users were free to take bikes anywhere they chose for up to 24hrs so long as they paid the fees. Sometimes this meant putting them on the back of their cars and taking them far outside the city limits to a place they wanted to ride. To end the rental and stop being charged they had to return the bike to the service area.

This isn’t something that was limited to my bikeshare system, there are many examples, like [this guy](https://www.dailymail.co.uk/news/article-7776847/Cyclist-rides-three-gear-Boris-Bike-snakelike-Tour-France-mountain-stage.html). 

I think this is generally true if you were to rent a bike from a bike shop, just as it's true if you rent a car.  Limitations on where these vehicles can travel are set via local traffic law. 

The limitations that we were subject to as an operator had to do with where we were permitted to offer vehicles for rental, not where users could take those vehicles.  If a bike was left by a user outside our service area, we would have to go and retrieve it since we weren't permitted to offer rentals in adjoining suburbs. In the current spec, anything outside the service area polygon would be:
`ride_allowed : FALSE, ride_through_allowed : TRUE`.

I've also seen the same rules applied to free floating car sharing. When Car2Go was operating in the US, you could drive the car as far as you wanted outside the service area and park the car for as long as you wanted but you couldn't end your rental (and stop being charged) until you returned to the service area. 

This is all a long way of saying that many operators may not place restrictions on how far vehicles can travel outside the service area, but there is a need to place limitations on where vehicles can be rented and where rentals can end. 

I would like to find a way to accomplish this that doesn't require operators to place limitations on their services in order to comply with the specification. The specification should model operators' business practices, not the other way around.

bdferris-v2:
@mplsmitch trying to get to the heart of the matter, it sounds like you are in-favor of removing the "restrict by default if geofences are defined" clause?  Namely:

> When `geofencing_zones.json` is defined, rides may not start, end, or travel through any area of the system unless explicitly enabled by a geofence zone and its associated rules.

Instead, we could replace it with two options:

1) More restrictive.  Change the semantics of "travel through":

> When `geofencing_zones.json` is defined, rides may not start or end in any area of the system unless explicitly enabled by a geofence zone and its associated rules.  Ridges may travel through any area of the system unless explicitly disabled by a geofence zone and its associated rules.

2) Less restrictive.  Enable everything by default:

> When `geofencing_zones.json` is defined, there are no default restrictions on any area of the system not covered by a geofence zone.  As such, rides can start, end, or travel through all areas of the system by default unless explicitly restricted by a geofence zone and its associated rules.

(We can wordsmith the details here)

As discussed above, a provider still has the power to globally restrict rides outside of their operating zone in either case.  They just have to be explicit about it.

Thoughts?

leonardehrenfried:
@bdferris-v2 with your latest suggestions how would you model a general operating area that you cannot leave?

A ring around it with a ride_allowed=false?

bdferris-v2:
@leonardehrenfried yeah, I believe that's the idea.  Though "ring" isn't maybe the right mental model, even though that's how it might be rendered:

```
┌─Poly─A──────────────────┐
│                         │
│                         │
│    ┌──Poly─B──────┐     │
│    │              │     │
│    │              │     │
│    │              │     │
│    │              │     │
│    │              │     │
│    └──────────────┘     │
│                         │
│                         │
└─────────────────────────┘
```

In practice, we'd have a larger Polygon A that defines the overall restricted area (it could be as large as the world, conceivably).  We'd then layer a smaller Polygon B over top (by specifying it first) that defines the active operating area.

Does that make sense?

mplsmitch:
> In practice, we'd have a larger Polygon A that defines the overall restricted area (it could be as large as the world, conceivably).
 
If this outer area polygon could be "as large as the world", then maybe there's no need for a polygon. What I mean is if you could declare rules that had a global scope,  and then override those rules using the polygons, that would solve all the issues.  So the default  TRUE/FALSE state for each rule used in the file is set first and then changed using the polygons.

Something like:
```

"global_rules" : [
 {
          "vehicle_type_id": ["bike"],
          "ride_start_allowed": false,
         "ride_end_allowed": false,

        }
],
  {
    "geometry": { "#": "... Polygon A ..." },
    "properties": {
      "rules": [
        {
          "vehicle_type_id": ["bike"],
          "ride_start_allowed": true,
         "ride_end_allowed": true,

        }
      ]
    }
  },`
```

bdferris-v2:
Would global rules be required?  Would they be required to cover all vehicle types?  If the answer is "no" to either of those questions, what should the default behavior be?

leonardehrenfried:
> Does that make sense?

It does make sense to me. I have to think about whether that's easier or harder to understand than the clockwise/counterclockwise behaviour.

mplsmitch:
> Would global rules be required? 

No - I don't think they need to be required, see below on default behavior


> Would they be required to cover all vehicle types?

No - I would use the same language that's in the current `rules` array: "If vehicle type IDs are not specified, then restrictions apply to all vehicle types." 

> If the answer is "no" to either of those questions, what should the default behavior be? 

The default behavior in any area should always be no restriction.  Similar to the question above on vehicle IDs, any area including global that doesn't have a rule specified contains no restrictions.

bdferris-v2:
> The default behavior in any area should always be no restriction.

@mplsmitch I hear and acknowledge your argument that there should be no restrictions on travel by default.  I'm generally ok with making unrestricted travel the default, but I'll admit I'm slightly more worried about defaulting to no restrictions on trip start/end, as my intuition is that there are few operators that truly allow their trips to start/end anywhere.

What are your thoughts on adopting your `global_rules` amendment but upgrading it to a required field?  I'm generally in favor of anything that makes the travel policy explicit, instead of relying on a default that might not be intuitive to a particular provider.

mplsmitch:
@bdferris-v2 When I said no restrictions on travel by default I was thinking what's the right way to handle a file where these fields are either not included or null. In other words if  `global_rules` was not defined or empty you wouldn't apply a restriction by default. I'm okay with making the `global_rules` required, but maybe it should be conditioned on the fields where it makes sense. For example if  `ride_start`, `ride_end` or `ride_through` are applied to any polygons, then you'd be required to also set them globally.  
 If none of these are defined and the only thing in the file is a polygon with `station_parking` or `maximum_speed_kph` then maybe you don't need to set them globally? If there's no speed limit outside the polygon, what number do you set `maximum_speed_kph` at?

bdferris-v2:
@mplsmitch `ride_start`, `ride_end` or `ride_through` are all required fields within a rule, so my general idea is that by making `global_rules` required, you'd be forced to specify those fields.  In practice, it's a bit more complicated than that because of the `vehicle_type_id` semantics, but we could theoretically write the spec in such a way that `global_rules` is required in `geofencing_zones.json` and must cover all vehicle type ids in the feed.

mplsmitch:
I'm sorry, I had forgotten that those were already required. In that case it seems pretty straight forward. Thanks- I think we're getting close to something that will work for everyone.

bdferris-v2:
@mplsmitch I updated the PR to include a required `global_rules` field.  Since rules are now used in two different fields, I took the liberty of breaking out the definition of Rule objects in its own section to avoid duplication, which is a bit of a departure for the spec.  I'd be interested in the proposed semantics along with the formatting of the documentation.

sven4all:
I like this proposal now the global rules are added. We would like to use this proposal to start communicating service areas between operators and municipalities, to give municipalities the opportunity to check if operators are complying with the permitted service areas. 

What is needed to start a vote and make this proposel part of GBFS 3.0 RC?

gbfs.md
sven4all: I agree with this comment, if we drop this requirement I think it's much easier to reason about polygons if you don't have to think about something that's not 'visiual'

gbfs.md
```diff
@@ -1310,30 +1310,54 @@ Field Name | REQUIRED | Type | Defines
...
+&emsp;&emsp;\-&nbsp;`rules` | OPTIONAL | Array&lt;[Rule](#geofencing-rule-object)&gt; | Array of [Rule](#geofencing-rule-object) objects defining restrictions that apply within the area of the polygon.  See [Geofencing Rule Precedence](#geofencing-rule-precedence) for details on semantics of overlapping polygons, vehicle types, and other precedence rules.
+`global_rules` | Yes | Array&lt;[Rule](#geofencing-rule-object)&gt; | Array of [Rule](#geofencing-rule-object) objects defining restrictions that apply globally in all areas as the default restrictions, except where overridden with an explicit geofencing zone.  See [Geofencing Rule Precedence](#geofencing-rule-precedence) for more details.<br/<br/>A rule or list of rules, as appropriate, must be specified in the global rules list covering all vehicle types in the feed.
 
-**Example:**
+#### Geofencing Rule Object
+
+A `Rule` object defines the set of restriction in place for a particular zone.  `Rule` objects are used as a value type in multiple places in `geofencing_zones.json`.  A `Rule` object has the following fields:
```
mplsmitch: typo: set of restriction(s)

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+
+```json
+"geofencing_zones.features": [
+  {
+    "geometry": { "#": "... Polygon A ..." },
+    "properties": {
+      "rules": [
+        {
+          "vehicle_type_id": ["bike"],
+          "ride_through_allowed": true,
```
mplsmitch: extra comma after : true

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+    }
+  },
+  {
+    "geometry": { "#": "... Polygon B ..." },
+    "properties": {
+      "rules": [
+        {
+          "vehicle_type_id": ["bike"],
+          "ride_through_allowed": false,
+          "maximum_speed_kph": 20,
```
mplsmitch: extra comma after : 20

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+        }
+      ]
+    }
+  }
+],
+"global_rules": [
+  {
+    "vehicle_type_id": ["bike"],
+    "ride_through_allowed": false
```
mplsmitch: missing comma after : false
extra comma after : 10 on next line

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+
+```json
+"geofencing_zones.features": [
+  {
+    "geometry": { "#": "... Polygon A ..." },
+    "properties": {
+      "rules": [
+        {
+          "vehicle_type_id": ["bike"],
+          "ride_through_allowed": true,
```
mplsmitch: extra comma after : true

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+      ]
+    }
+  },
+  {
+    "geometry": { "#": "... Polygon B ..." },
+    "properties": {
+      "rules": [
+        {
+          "vehicle_type_id": ["scooter"],
+          "ride_through_allowed": false,
```
mplsmitch: extra comma after : false

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+```json
+"geofencing_zones.features": [
+[
+  {
+    "geometry": { "#": "... Polygon A ..." },
+    "properties": {
+      "rules": [
+        {
+          "vehicle_type_id": ["bike", "scooter"],
+          "ride_through_allowed": true,
```
mplsmitch: extra comma after : true

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+      ]
+    }
+  },
+  {
+    "geometry": { "#": "... Polygon B ..." },
+    "properties": {
+      "rules": [
+        {
+          "vehicle_type_id": ["scooter"],
+          "ride_through_allowed": false,
```
mplsmitch: extra comma after : false

gbfs.md
```diff
@@ -1423,11 +1448,175 @@ Field Name | REQUIRED | Type | Defines
...
+    ]
   }
 }
 ```
 
+##### Polygon Overlap Examples
+
+The following polygon diagram will be used in a number of examples below of geofencing rule resolution.
+
+![Geofencing Example Diagram with two overlapping polygons A and B, and areas a, b, and ab.](geofencing_example.svg)
```
mplsmitch: I tried to look at this using a dark mode browser and the text in the diagrams was illegible because of the transparent background. Would be better if they were on white or the text color was changed. Also, we should probably set up an /images directory to keep things neat and clean as more images are added to the repo in the future.

mplsmitch (changes requested):
just a few things that need cleaning up - the diff shows something wrong around line 1605 but I can't see what it is, maybe correcting the JOSN will fix it

gbfs.md
bdferris-v2: To keep things simple, I switched the image to a white background.  I also moved it to a new images/ sub-directory.

bdferris-v2:
Ok, I've made the updates suggested by @mplsmitch.  As there seems to be positive engagement on where we've ended up at with this proposal, I'm going to call a vote.

bdferris-v2:
I hereby call a vote on this proposal. Voting will be open for 10 full calendar days until 11:59PM UTC on Thursday, March 9th, 2023.

benwedge:
Joyride (feed provider) votes in favour. 

We do not currently use the geofencing file, but may do so in the future.

testower:
Entur (aggregator) votes in favour.

We will implement support for this proposal.

sven4all:
CROW (aggregator for municipalities) votes in favour.

If this vote passes we will try to implement this next month within our dashboard https://dashboarddeelmobiliteit.nl/.

leonardehrenfried:
OpenTripPlanner votes in favour although it's not clear yet if and when it will be implemented.

ezmckinn:
Superpedestrian votes in favor. No timeline for implementation yet, but would implement in the future.

josee-sabourin:

Votes in favor:
Joyride (producer)
Entur (consumer)
CROW (consumer)
OpenTripPlanner (consumer)
Superpedestrian (producer)
 
There were no votes against.

Thank everyone for coming together on this! This is a massive accomplishment.

josee-sabourin (approved):

PR #485, opened 2023-01-27
Added Otto to systems.csv

arehlt:
Adds `otto` as an entry in `systems.csv`


josee-sabourin (approved):

josee-sabourin:
Welcome Are! Thanks for this contribution, if I'm not mistaken this is the first carsharing feed in the catalog. Can I ask why you create one feed for all regions rather than separate feeds?

arehlt:
Thanks 🙌 

Both [MoveAbout](https://github.com/MobilityData/gbfs/blob/master/systems.csv#L418) and [Hertz](https://github.com/MobilityData/gbfs/blob/master/systems.csv#L415) are carsharing.

Regarding regions: We briefly spoke with Entur about this and decided to keep it one feed now because of some business logic on our end. But we'll split them into separate feeds in the future because of their requirements:
> Data for several geographic regions (i.e. cities) should be delivered in separate feeds.

josee-sabourin:
Oh those two slipped passed me! Thanks for pointing them out. 

Looking forward to getting all those separated feeds into the catalog when ready. The new file manifest.json might be interesting to you. Read about it at #462. 

Additionally, consider joining our [Slack community](https://bit.ly/mobilitydata-slack) to stay in the loop on GBFS news.

Issue #434, opened 2022-06-06
Add easier interface for contributions to system.csv


CourchesneA:
### **Have you read our [FAQ](https://gbfs.mobilitydata.org/faq)? It’s possible your question can be answered there!**
yes
### **If you are new to the specification, please introduce yourself (name and organization/link to GBFS). It’s helpful to know who we're chatting with!** 
Anthony Courchesne - [IVI](https://www.ivisolutions.ca/en/)
Attendee at IMDS / Developer workshop

### **What is the issue and _why_ is it an issue?**
In one of our discussions, we mentioned that we might get more contributions to system.csv if the contributors did not require interaction with the github repo. Currently, people need to know how to open an issue or a PR.

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**
One solution would be to have a google form with fields possibly corresponding to columns in the system.csv, possibly increasing the amount of contributions to the database.

### **Is your potential solution a breaking change?**
- [ ] Yes 
- [X] No
- [ ] Unsure

josee-sabourin:
Closing this since it has been solved by #487

PR #487, opened 2023-01-31
Add contribution form for systems.csv in README.md

josee-sabourin:
As discussed in #434, there is now a Google form to contribute to systems.csv which provides a Github-less opportunity to contribute. 

Note: This addition does not require a vote.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Each new addition to the spec adds complexity. We want to avoid additions to the
 
 ## Systems Catalog - Systems Implementing GBFS
 There are now over 600 shared mobility systems publishing GBFS worldwide. This list contains all known systems publishing GBFS feeds and is maintained by the GBFS community. This is an incomplete list. If you have or are aware of a system that doesn’t appear on the list please add it.
-If you would like to add a system, please fork this repository and submit a pull request. Please keep this list alphabetized by country and system name.
+If you would like to add a system, please fork this repository and submit a pull request. Please keep this list alphabetized by country and system name. Alternatively, fill out [this contribution form](https://forms.gle/WSXFuXx9k4PSTfbC9) for a Github-less contribution. 
 * [systems.csv](systems.csv)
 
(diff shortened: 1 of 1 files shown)


emmambd (changes requested):
2 comments on the Google form: 

1 nit: it's cool that you can validate that the ISO code is in fact an ISO code! I would give some thought to how strict you want to be about that when our main goal is to encourage submissions. Your call, but it's worth considering just leaving it as text. 

1 bug: the Name field for the system requires an email right now.

josee-sabourin:
You're right, I've removed the data validation from that question. 

Also changed the format for name! 

thanks for the review :)

futuretap:
Do we actually need all these fields? At least in theory, filling out the GBFS feed URL should be all that’s needed since all the fields could be automatically retrieved from system_information.json. (Not sure though if/how such an automation could be done with Google forms.)

Having the validation URL as a separate entry in systems.csv is superfluous imo since it can be directly derived from the feed URL. This might be even more correct since currently, systems.csv has some broken validation URLs for Link, Helbiz and Donkey Republic.

josee-sabourin:
Hi @futuretap, currently in the form the only required field is the auto-discovery URL, if the contributor would like to provide more they can, otherwise we will add the missing information. When dealing with a large number of feeds to add, any and all extra information helps. We currently don't have any automations to harvest this data, although we could create some in the future, but for now this is what we are working with.

In terms of validation URL, there are definitely some bugs and we are working internally to identify out the best way forward to provide validation in an easy way.

emmambd (commented):
@josee-sabourin nit: the location field is required currently in the google form, as well as the auto-discovery URL.

josee-sabourin:
@emmambd location fixed, auto-discovery URL remains required since we cannot add a new entry without it! All of the information not required needs to be harvested from the auto-discovery URL.

emmambd (approved):

PR #505, opened 2023-04-20
Add Dott in various countries to systems.csv

viestat:
Adds Dott public feed for various cities to systems.csv.


josee-sabourin:
Hi Andres, 

Thanks for this! 39 feeds, such an amazing contribution. 

It seems every entry has the same auto-discovery URL `https://gbfs.api.ridedott.com/public/v2/gbfs.json`,in this endpoint if see multiples of the various endpoints but for the different cities. 

Do you have auto-discovery URLs for each city? Ex  `https://gbfs.api.ridedott.com/public/bordeaux/v2/gbfs.json`

This also seems like a great first step towards `manifest.json` (see #462), is there any intention from Dott to publish this endpoint in the future? 

Thanks!

viestat:
Hello @josee-sabourin !

Thank you for your review!
Here are some answers to your questions:

> It seems every entry has the same auto-discovery URL `https://gbfs.api.ridedott.com/public/v2/gbfs.json`,in this endpoint if see multiples of the various endpoints but for the different cities.
> 
> Do you have auto-discovery URLs for each city? Ex `https://gbfs.api.ridedott.com/public/bordeaux/v2/gbfs.json`

~~At the moment this is not the case, which is why they all share the same entry point. Once this happens we should also have  auto-discovery URLs for each city.~~

Edit: We have made it possible to have an auto-discovery URL per region, will update the PR.

> This also seems like a great first step towards `manifest.json` (see #462), is there any intention from Dott to publish this endpoint in the future?

We were not aware of this file's existence since it seems to be part of the [v3.0-RC](https://github.com/MobilityData/gbfs/tree/v3.0-RC) but it seems to be valuable for us. We currently have no plans to publish it but I will make a case for it internally.

josee-sabourin (approved):

PR #510, opened 2023-05-23
Update gbfs.md : last_updated description

fchabouis:
My name is Francis Chabouis, I work for [transport.data.gouv.fr](transport.data.gouv.fr), the National Access Point of France.

Some of the French GBFS producers seem to update the `last_updated` field rarely in some files, typically `gbfs.json`. This file is not expected to change very often, but an old `last_updated` is difficult to interpret for a user. Does it mean there has been no update on that file for a long time (but it is still up-to-date)? Or is the file stale?

While trying to understand why they published their data that way, I came to read the description of the field in the documentation. And found it slightly misleading. It says "Indicates the last time data in the feed was updated.".  I think this could be interpreted as "Indicates the last time data in the feed was **changed**". This interpretation would allow a data producer to keep an old `last_updated` value, as long as the file content doesn't change.

The second sentence is more explicit: "This timestamp represents the publisher's knowledge of the current state of the system at this point in time."

I propose to modify the first sentence to remove any ambiguity. 

"Time at which the data in the feed was guaranteed to be up-to-date. This timestamp represents the publisher's knowledge of the current state of the system at this point in time."

Feel free to propose any other wording, as English is not my primary language.

- [ ] Yes 
- [x] No
- [ ] Unsure

None

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -245,7 +245,7 @@ Every JSON file presented in this specification contains the same common header 
 
 Field Name | REQUIRED | Type | Defines
 ---|---|---|---
-`last_updated` | Yes | Timestamp | Indicates the last time data in the feed was updated. This timestamp represents the publisher's knowledge of the current state of the system at this point in time.
+`last_updated` | Yes | Timestamp | Time at which the data in the feed was guaranteed to be up-to-date. This timestamp represents the publisher's knowledge of the current state of the system at this point in time.
 `ttl` | Yes | Non-negative integer | Number of seconds before the data in the feed will be updated again (0 if the data should always be refreshed).
 `version` | Yes | String | GBFS version number to which the feed conforms, according to the versioning framework.
 `data` | Yes | Object | Response data in the form of name:value pairs.


mplsmitch:
@fchabouis  Does this really change anything? I don't see how it would force anyone to publish files more frequently. If a producer only publishes a file like gbfs.json when they make a modification to it, they would still be able to "guarantee" that the file is up-to-date as of the time stamp. 

I agree it's confusing when you see files with a timestamp way in the past and a `ttl` of `0`.  Files that change infrequently should probably have a `ttl` of 12-24hrs.

fchabouis:
Hi @mplsmitch ,
It doesn't change anything, it's just a clarification of the purpose of the field.
When I read the the field description "Indicates the last time data in the feed was updated." I thought "maybe the producers are right then, if the file has not been updated for 3 months, it's expected to have an old `last_updated` value."

And I went to the MD slack and ask for [clarification](https://mobilitydata-io.slack.com/archives/CNXA9ASBV/p1684830007353919) and people seemed to agree that the current description was ambiguous.

I hope that with a clearer description, producers won't make the mistake to think it's ok to serve the same file for months without updating the `last_updated` field.

mplsmitch:
I'm not opposed to this, I just question whether it solve your problem.  I agree there's some ambiguity in the current `last_updated` description. It could represent the last time the file was published, or the last time data in the file was modified. A third possibility would be that the time stamp represents when the request for the file was made to the server if the file was created on the fly.

> I hope that with a clearer description, producers won't make the mistake to think it's ok to serve the same file for months without updating the last_updated field.

 I think as it's currently defined , **it is ok** to serve the same file for months and I don't see how the language you're proposing changes that.  In either case, whether I take it to mean last published or last modified, I could also say that data in the feed was "guaranteed to be up-to-date" as of the time stamp regardless of how far in the past the timestamp is.

Issue #519, opened 2023-07-06
Use human-readable timestamp (like ISO-8601)


leonardehrenfried:
### **What is the issue and _why_ is it an issue?**

Right now GBFS uses seconds since the epoch as a timestamp, for example in the `last_updated` field. While this practice is pretty wide-spread, it makes debugging feeds a little harder than necessary.

Every time I need to check the `last_updated` field I need to copy the epoch seconds, google "epoch to time converter", paste the seconds to get a usable date time.

While this isn't an impossible task, I think there are alternatives that are both human-readable _and_ unique so that machines can identify a universal point in time.

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**

I suggest that GBFS uses [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) date times.  An example would be `2023-07-06T11:35:33+0200`.

This is much easier to interpret for a human and can be parsed easily by many programming languages.

This is of course a major breaking change so it can only be introduced for a major version jump.

### **Is your potential solution a breaking change?**
- [x] Yes, very much so
- [ ] No
- [ ] Unsure

mplsmitch:
For what it's worth, we had a lengthy discussion on this early on in #9.  I think the consensus was that we should switch to RFC 3339  with a timezone of "Z" (i.e. UTC) but at the time we lacked versioning and since it would have broken existing feeds we decided to  table it.

sven4all:
Would appreciate this change, let's introduce it in 3.0.

leonardehrenfried:
If I understand it correctly then RFC3339 is a subset of ISO 8601 which reduces the number of allowed formats, isn't it? That would be great since ISO gives quite a bit of flexibility.

With two people in favour I believe this might have a chance of making it into GBFS 3.

Who else do I need to convince?

leonardehrenfried:
About only allowing UTC as the timezone: the Java parser, which I would use, can deal with any offset so I don't have a strong opinion either way but it's probably better to be restrictive early on and then allow more flexibility later, if there is a need.

sven4all:
UTC makes sense to me, there is no need for the local timezone because this data is not user focussed.

josee-sabourin:
Hi all! A reminder of the governance process, especially if we want to move quickly to roll this into v3.0 (which would be v3.0-RC2, also including #497). A Pull Request must be open for 7 calendar days before we can open a vote, so if we open a PR today, we can call a vote on Thursday, July 13th. 

[...]

PR #522, opened 2023-07-17
Use RFC3339 for Timestamp

leonardehrenfried:

I'm a core developer of OpenTripPlanner, a multimodal transit routing engine with GBFS support.

Right now GBFS uses seconds since the epoch as a timestamp, for example in the last_updated field. While this practice is pretty wide-spread, it makes debugging feeds a little harder than necessary.

Every time I need to check the last_updated field I need to copy the epoch seconds, google "epoch to time converter", paste the seconds to get a usable date time.

While this isn't an impossible task, I think there are alternatives that are both human-readable and unique so that machines can identify a universal point in time.

Replacing POSIX time with RFC3339 makes a much more pleasant debugging experience for humans and doesn't add ambiguity for computers.

An example of RFC3339 time is `2023-07-17T13:45:29+02:00`.

- [X] Yes 
- [ ] No
- [ ] Unsure

All of them. :flushed: 

#### Timezone

~It was suggested in #519 that we should limit the offset/timezone to UTC, so an offset of 0 or a `Z` at the end.~

~Personally, I'm not in favour of this restriction as you should use a proper date parsing library which handles all offsets, but as time is tight I would like to keep the discussion to a minimum and not let this point block the proposal.~

EDIT: Consensus was reached to include the offset/timezone in this PR.


josee-sabourin:
Tagging those who were active in #519 
@mplsmitch @sven4all @benwedge @testower @futuretap

benwedge:
> Timezone
It was suggested in https://github.com/MobilityData/gbfs/issues/519 that we should limit the offset/timezone to UTC, so an offset of 0 or a Z at the end.

>Personally, I'm not in favour of this restriction as you should use a proper date parsing library which handles all offsets, but as time is tight I would like to keep the discussion to a minimum and not let this point block the proposal.

IMO we should just get it right the first time. If it's going to be a breaking change and every language has a library that handles this nicely why not just do it?

leonardehrenfried:
> IMO we should just get it right the first time. If it's going to be a breaking change and every language has a library that handles this nicely why not just do it?

That would of course be okay for me.

testower:
I agree with @benwedge let us not do it half way.

futuretap:
I support the proposal, including variable offsets.

leonardehrenfried:
Thanks all for chiming in. I will wait until the end of today and if no one shows up voicing opposition to variable offsets I will update the PR.

leonardehrenfried:
Since no-one has shown up who is against variable offsets, I have updated the PR.

gbfs.md
```diff
@@ -1275,8 +1275,8 @@ Field Name | REQUIRED | Type | Defines
         ],
         "times": [
           {
-            "start": 1604448000,
-            "end": 1604674800
+            "start": "2023-07-17T13:34:13+02:00",
+            "end": "2023-07-18T13:34:13Z"
```
mplsmitch: local offset  for "end" should be +2:00

gbfs.md
```diff
@@ -1426,8 +1426,8 @@ See examples below.
                 "language": "en"
               }
             ],
-            "start": 1593878400,
-            "end": 1593907260,
+            "start": "2023-07-17T13:34:13+02:00",
+            "end": "2024-07-17T13:34:13Z",
```
mplsmitch: local offset  for "end" should be +2:00

gbfs.md
leonardehrenfried: You're right, that's a search & replace missing this case. Corrected in 348929606c7541cbcfac90490700c6178ddcd4d0.

gbfs.md
leonardehrenfried: You're right, that's a search & replace missing this case. Corrected in 348929606c7541cbcfac90490700c6178ddcd4d0.

mplsmitch (approved):

josee-sabourin:
I (as proxy for @leonardehrenfried who is on vacation) hereby call a vote on this proposal. Voting will be open for 10 full calendar days until 11:59PM UTC on August 5th.

testower:
Entur votes in favor of this proposal and will implement it

futuretap:
We (Where To?) are in favor, too, and will implement it.

AntoineAugusti:
transport.data.gouv.fr is in favour and will implement it (for feeds we consume)

benwedge:
Joyride (provider) is in favour and will implement this when we move to 3.x

leonardehrenfried:
Not sure if my vote counts but OpenTripPlanner will implement this via Entur's Java library.

josee-sabourin:

Votes in favor:
Entur (consumer)
Whereto.app (consumer)
Transport.data.gouv (consumer)
Joyride (producer)
 
There were no votes against.

richfab:
Datetime is currently `ISO 8601` notation in the spec. Should Datetime be replaced by Timestamp (`RFC3339`)? 

Datetime definition in the spec:
https://github.com/MobilityData/gbfs/blob/master/gbfs.md?plain=1#L195

Datetime only usage is in vehicle_status.json for `available_until` field:
https://github.com/MobilityData/gbfs/blob/master/gbfs.md?plain=1#L956

testower:
Good catch @richfab - I think for most practical purposes they are the same, but the spec should definitely be consistent. The ISO is very broad, so I would be in favor of restricting it to the RFC (which is a profile of the ISO standard?)

leonardehrenfried:
My suggestion is to combine the two, so all of them will be of type Timestamp or DateTime.

I have a slight personal preference for the name DateTime but I'm fine with both.

leonardehrenfried:
@josee-sabourin What is the process of merging this PR?

I would prefer to merge this and then make a follow up where I combine the two types since that would be a non-breaking change.

josee-sabourin:
@leonardehrenfried currently we are waiting to merge pending a few more potential breaking changes to limit the number of v3.0 release candidates we have.

hbruch:
Note that [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) allows `T` and `Z` also to be written in lowercase, `T` might also be left out. Should the spec prescribe uppercase `T` and `Z` or allow these variations? In the latter case, the regex in gbfs-json-schema of course should reflect this.

richfab:
Hello @hbruch,

Great question. For reference, this is the extract from the [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) spec:

>NOTE: Per [[ABNF](https://www.rfc-editor.org/rfc/rfc3339#ref-ABNF)] and ISO8601, the "T" and "Z" characters in this syntax may alternatively be lower case "t" or "z" respectively.
This date/time format may be used in some environments or contexts that distinguish between the upper- and lower-case letters 'A'-'Z' and 'a'-'z' (e.g. XML).  Specifications that use this format in such environments MAY further limit the date/time syntax so that the letters 'T' and 'Z' used in the date/time syntax must always be upper case.  Applications that generate this format SHOULD use upper case letters.
NOTE: ISO 8601 defines date and time separated by "T". Applications using this syntax may choose, for the sake of readability, to specify a full-date and full-time separated by (say) a space character.

What is everyone's opinion about:
1. Restricting `T` and `Z` to upper case or not?
2. Allowing `T` to be left out or not?

Thank you!

leonardehrenfried:
I would be fine with not supporting the full flexibility of the spec and require them both to be mandatory and upper case. (Z can of course be replaced with an offset).

futuretap:
We're strongly preferring to restrict `T` and `Z` to upper case and require it not to be left out. Technical reason: Swift `Decodable` requires this strict format.

richfab (approved):
LGTM

gbfs.md
```diff
@@ -221,7 +221,7 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
 * Phone Number *as of v3.0-RC* - Phone number in [E.164](https://www.itu.int/rec/T-REC-E.164-201011-I/en) format. The phone number MUST start with a "+". The characters following the "+" MUST be integers and MUST NOT contain any hyphens, spaces or parentheses.
 * String - Can only contain text. Strings MUST NOT contain any formatting codes (including HTML) other than newlines.
 * Time - Service time in the HH:MM:SS format for the time zone indicated in `system_information.json` (00:00:00 - 47:59:59). Time can stretch up to one additional day in the future to accommodate situations where, for example, a system was open from 11:30pm - 11pm the next day (23:30:00-47:00:00).
-* Timestamp - Timestamp fields MUST be represented as integers in POSIX time (representing the number of seconds since January 1st 1970 00:00:00 UTC).
+* Timestamp - Timestamp fields MUST be represented as strings in [RFC3339 format](https://www.rfc-editor.org/rfc/rfc3339), for example `2023-07-17T13:34:13+02:00`.
```
richfab: As of v3.0-RC2

```suggestion
* Timestamp - Timestamp fields MUST be represented as strings in [RFC3339 format](https://www.rfc-editor.org/rfc/rfc3339), for example `2023-07-17T13:34:13+02:00`. *(as of v3.0-RC2)*
```

PR #532, opened 2023-08-25
fix(systems): replace offenburg by new ortenau feed

tobsesHub:
GBFS Feed for Nextbike Offenburg does not exist anymore. Instead, there is a new feed for Nextbike Ortenau.
Therefore replaced them in systems.csv


josee-sabourin (approved):

josee-sabourin:
Welcome @tobsesHub and thanks for this contribution! It's great to see `system_regions.json` being put to good use.

PR #545, opened 2023-09-20
Specify IDs characters range restrictions

### LINKED ISSUE #541: GBFS ID type definition
### **What is the issue and _why_ is it an issue?**

The current definition of `ID` in https://github.com/MobilityData/gbfs/blob/master/gbfs.md#field-types is:
- "a string" that "MUST NOT contain spaces"

This definition is the same since the first draft in 2016 https://github.com/MobilityData/gbfs/pull/5 :
> must not contain spaces

There are multiple problems with that definition: 
- A precision problem: What is a space? It's not defined here, and as any character that JSON can encode is valid, the definition is not straightforward:
  - is it `0x20` (ASCII space)
  - is it any Unicode whitespace? (https://en.wikipedia.org/wiki/Whitespace_character)
  - What about tabulation and other Unicode control characters? https://en.wikipedia.org/wiki/List_of_Unicode_characters#Control_codes

- A purpose problem: Why spaces are forbidden? What was the goal?

- The fact that allowing any characters (except spaces) may cause compatibility problems:
  - Extended Unicode characters needs to be represented differently in JSON, with two escaped character `\uXXXX\uXXXX`
  - Two different code point sequences may represent the same thing depending on the encoding path, and may need normalization to compare them  (https://en.wikipedia.org/wiki/Unicode_equivalence)
  - Non-ASCII characters may cause compatibility problem during transport and storage
  - Manual debugging for producers and integrators may be harder (hard to read, write, copy without errors)

### Current usage of IDs (based on `system.cvs`)

Of the 826 system declared, I've analyzed 798 (others are dead or broken)

IDs from `free_bike_status`, `station_status`, `vehicle_types` and `system_pricing_plans` were analyzed.

The IDs of those systems are composed of:

- `.` (dot) for 1 system (`sharedmobility.ch`)
- `@` (AT) for HelBiz 2 systems (`"bike_id"` like `"P8S0C2149C0011@SD02"` in https://gbfs.helbiz.com/v2.2/miamilakes/free_bike_status.json)
- ` ` (space !!!) for 7 systems (mostly in vehicles types)
  - 5 `publicbikesystem.net` in `vehicle_type` (ex. : `"vehicle_type_id": "E Scooter"`)
  - `openov.nl` in `station_information` (one station has the id `"GR002 "`, which is probably an error as all other station follow the same pattern without space in https://gbfs.openov.nl/ovfiets/station_information.json)
  - `sharedmobility.ch` in `plan_id` such as `emobility:83106 return"` and `emobility:2306 oneway`
- `:` for 34 systems (`sharedmobility.ch` and `entur`)
- All other systems only use `0-9`, `a-z`, `A-Z`, `-` and `_`

### **Please describe some potential solutions you have considered (even if they aren’t related to GBFS).**

The specification should clearly state what constitute a valid ID. Some solutions are presented below:

- Any Unicode character
  - Advantages: least restrictive
  - Drawbacks: comparison, encoding and compatibility issues
- Any Unicode character expects a specific list (Spaces? Non-printable characters?)
  - Advantages: most compatible with the current specification
  - Drawbacks: comparison, encoding and compatibility issues
- Any printable ASCII character (https://en.wikipedia.org/wiki/ASCII#Printable_characters)
  - Advantages: high compatibility
  - Drawbacks: a few escaping issues, restrictive compared to the current specification
- A restricted list of printable ASCII character
  - Advantages: high compatibility
  - Drawbacks: a few escaping issues, restrictive compared to the current specification
- `[A-Z][a-z][0-9]_-.:@`
  - Advantages: very high compatibility (791/798 : 99.1%, not 100% because it excludes the 7 that are using spaces)
  - Drawbacks: restrictive compared to the current specification
- `[A-Z][a-z][0-9]_-:` (789/798 : 98.8%)
  - Advantages: high compatibility, no encoding or escaping issue
  - Drawbacks: restrictive compared to the current specification
- `[A-Z][a-z][0-9]_-` (756/798: 94.7%)
  - Advantages: high compatibility, no encoding or escaping issue
  - Drawbacks: restrictive compared to the current specification

As 3.0 is a major breaking version, it is the perfect moment to restrict the character range of IDs.
Restricting them to the strictest proposition `[A-Z][a-z][0-9]_-` seams to be a reasonable choice:
  - It makes it simpler to store, transmit and compare IDs without any compatibility problem
  - They could be used in URIs without any issues (both web and Apps)
  - Systems with non-compatible IDs could simply encoded them in Base64url to be compatible
A least restrictive choice, regarding current usage, could be `[A-Z][a-z][0-9]_-.:@` or something in between. (`:` and `@` are mostly safe in most situation, but may be interpreted differently in URIs).

### Comparisons with IDs in other systems:
- In GTFS: 
https://gtfs.org/schedule/reference/
> ID - [...] is a sequence of any UTF-8 characters. Using only printable ASCII characters is recommended. [...]

- RFC 7946: The GeoJSON Format
https://datatracker.ietf.org/doc/html/rfc7946#appendix-B.1
> A Feature object's "id" member is a string or number

- HTML IDs
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id
> Technically, the value for an id attribute may contain any character, except [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) characters. However, to avoid inadvertent errors, only [ASCII](https://developer.mozilla.org/en-US/docs/Glossary/ASCII) letters, digits, '_', and '-' should be used, and the value for an id attribute should start with a letter.

### References:
- JSON strings: https://datatracker.ietf.org/doc/html/rfc8259#section-7
> All Unicode characters may be placed within the
   quotation marks, except for the characters that MUST be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).

- https://fr.wikipedia.org/wiki/Base64#base64url
Base64 encoding require `[a-z][0-9]+/=`
Base64url require `[a-z][0-9]_-`

### **Is your potential solution a breaking change?**
- [x] Yes

tdelmas:
This PR clarify and restrict the characters allowed in IDs. See #541 for additional details.

Those edits have multiple goals:
- Clarify the current specification that is too vague
- Ensure interoperability, it MUST be possible to store and compare IDs easily and without pitfall
- Ensure that IDs are easy to manipulate by humans, because even if the format is design for machines, IDs are often used by humans to debug, so they SHOULD be easy to read, write, copy and compare, regardless of the keyboard layout or system used.

With that specification, all IDs present in `system.cvs` should stay compliant (excepts those using spaces `0x20`, who are already not compliant with the current specification).

Also, all others shared mobility systems that I have seen (GBFS or not) are compatible with both restrictions.

Does anyone think that IDs should be restricted further, such as `A-Za-z0-9_-:`? (And consider that existing system that are using `.@/` are compliant because "SHOULD" means "that there may exist valid reasons in particular circumstances to ignore a particular item", and at least "legacy" may fit into that.

Fix #541

- [x] Yes

Most of them

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -202,8 +202,8 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
 * ID - Should be represented as a string that identifies that particular entity. An ID:
     * MUST be unique within like fields (for example, `station_id` MUST be unique among stations)
     * Does not have to be globally unique, unless otherwise specified
-    * MUST be in the ASCII printable character range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters
-    * SHOULD be restricted to `A-Z`, `a-z`, `0-9` and `.@:/_-`
+    * MUST be in the ASCII printable character range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters *(as of v3.0-RC2)*
+    * SHOULD be restricted to `A-Z`, `a-z`, `0-9` and `.@:/_-` *(as of v3.0-RC2)*
     * MUST be persistent for a given entity (station, plan, etc.). An exception is `vehicle_id`, which MUST NOT be persistent for privacy reasons (see `vehicle_status.json`). *(as of v2.0)*
 * Language - An IETF BCP 47 language code. For an introduction to IETF BCP 47, refer to https://www.rfc-editor.org/rfc/bcp/bcp47.txt and https://www.w3.org/International/articles/language-tags/. Examples: `en` for English, `en-US` for American English, or `de` for German.
 * Latitude - WGS84 latitude in decimal degrees. The value MUST be greater than or equal to -90.0 and less than or equal to 90.0. Example: `41.890169` for the Colosseum in Rome.
--- a/gbfs.md
+++ b/gbfs.md
@@ -202,7 +202,7 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
 * ID - Should be represented as a string that identifies that particular entity. An ID:
     * MUST be unique within like fields (for example, `station_id` MUST be unique among stations)
     * Does not have to be globally unique, unless otherwise specified
-    * MUST be in the ASCII printable caracter range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters
+    * MUST be in the ASCII printable character range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters
     * SHOULD be restricted to `A-Z`, `a-z`, `0-9` and `.@:/_-`
     * MUST be persistent for a given entity (station, plan, etc.). An exception is `vehicle_id`, which MUST NOT be persistent for privacy reasons (see `vehicle_status.json`). *(as of v2.0)*
 * Language - An IETF BCP 47 language code. For an introduction to IETF BCP 47, refer to https://www.rfc-editor.org/rfc/bcp/bcp47.txt and https://www.w3.org/International/articles/language-tags/. Examples: `en` for English, `en-US` for American English, or `de` for German.
--- a/gbfs.md
+++ b/gbfs.md
@@ -202,7 +202,8 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
 * ID - Should be represented as a string that identifies that particular entity. An ID:
     * MUST be unique within like fields (for example, `station_id` MUST be unique among stations)
     * Does not have to be globally unique, unless otherwise specified
-    * MUST NOT contain spaces
+    * MUST be in the ASCII printable caracter range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters
+    * SHOULD be restricted to `A-Z`, `a-z`, `0-9` and `.@:/_-`
     * MUST be persistent for a given entity (station, plan, etc.). An exception is `vehicle_id`, which MUST NOT be persistent for privacy reasons (see `vehicle_status.json`). *(as of v2.0)*
 * Language - An IETF BCP 47 language code. For an introduction to IETF BCP 47, refer to https://www.rfc-editor.org/rfc/bcp/bcp47.txt and https://www.w3.org/International/articles/language-tags/. Examples: `en` for English, `en-US` for American English, or `de` for German.
 * Latitude - WGS84 latitude in decimal degrees. The value MUST be greater than or equal to -90.0 and less than or equal to 90.0. Example: `41.890169` for the Colosseum in Rome.


josee-sabourin:
tagging others for visibility: @cmonagle @viestat @ArashMansouri @futuretap @AntoineAugusti @ezmckinn @fbouchPBSC @Carl-NM @bdferris-v2 @kulovan

futuretap:
+1 looks good

AntoineAugusti:
+1 as well

PierrickP:
(+1) Tom is my colleague, and we work on it together

cmonagle:
+1 From Transit

testower:
+1 from Entur

gbfs.md
```diff
@@ -202,7 +202,8 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
 * ID - Should be represented as a string that identifies that particular entity. An ID:
     * MUST be unique within like fields (for example, `station_id` MUST be unique among stations)
     * Does not have to be globally unique, unless otherwise specified
-    * MUST NOT contain spaces
+    * MUST be in the ASCII printable caracter range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters
```
futuretap: caracter => character

Cj-Malone:
+1

gbfs.md
tdelmas: Fixed, thanks!

tdelmas:
(to conform with https://github.com/MobilityData/gbfs/blob/master/governance.md)

I hereby call a vote on this proposal. Voting will be open for 10 full calendar days until 11:59PM UTC on Oct 12.

ArashMansouri:
+1 from Lyft

tdelmas:
@futuretap @AntoineAugusti @cmonagle @testower @Cj-Malone Thank you for the +1. I'm told that to as official as possible, votes should be done after the call to vote.

I'll be grateful if you could confirm your vote before 11:59PM UTC on Oct 12 **and include the organization for which you are voting in your comment**.

AntoineAugusti:
+1, transport.data.gouv.fr

futuretap:
+1 from Where To? / FutureTap

ArashMansouri:
+1 from Lyft

josee-sabourin:

testower:
+1 from Entur

bdferris-v2:
+1 from Google

josee-sabourin:

Votes in favor:
Transport.data.gouv (consumer)
Whereto.app (consumer)
Lyft (producer)
Entur (consumer)
Google (consumer)
 
There were no votes against.

This change will be added to v3.0-RC2, which will be released in the coming weeks.

richfab (approved):
LGTM

gbfs.md
```diff
@@ -202,7 +202,8 @@ Example: The `rental_methods` field contains values `creditcard`, `paypass`, etc
 * ID - Should be represented as a string that identifies that particular entity. An ID:
     * MUST be unique within like fields (for example, `station_id` MUST be unique among stations)
     * Does not have to be globally unique, unless otherwise specified
-    * MUST NOT contain spaces
+    * MUST be in the ASCII printable character range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters
+    * SHOULD be restricted to `A-Z`, `a-z`, `0-9` and `.@:/_-`
```
richfab: As of v3.0-RC2

```suggestion
    * MUST be in the ASCII printable character range, space excluded (0x21 to 0x7E) https://en.wikipedia.org/wiki/ASCII#Printable_characters *(as of v3.0-RC2)*
    * SHOULD be restricted to `A-Z`, `a-z`, `0-9` and `.@:/_-` *(as of v3.0-RC2)*
```

PR #565, opened 2023-11-14
systems.csv: Removes Louvelo

cmonagle:
System shut down a couple weeks ago: https://www.wave3.com/2023/10/04/louisvilles-bike-share-program-louvelo-will-shut-down-end-october/


richfab (approved):
Thank you @cmonagle!

This is sad news for people of Louisville.

PR #567, opened 2023-11-16
Add Authentication Info URL column to systems.csv

richfab:
## Problem
Some GBFS feeds require authentication using a free token provided by the producer. Consuming applications need to know how to obtain this token.

## Solution
Added an `Authentication Info URL` column to the systems.csv file to indicate to the consumer how to obtain the free token from the producer.

If authentication is required, the `Authentication Info URL` field contains a URL to a human-readable page describing how the authentication should be performed and how credentials can be created.


josee-sabourin (approved):

mplsmitch:
I'm against including auth info in systems.csv. This feels like an invitation for producers to add auth to their feeds which cuts against the open & public ethos of the specification.  I also think increasing the  complexity of this file with additional columns raises the bar for contributions. For example, how would the validation url function in  cases where auth is required?

futuretap:
It's a tough question. Indeed, authenticated feeds should not be encouraged. On the other hand, I'd prefer to include a few authenticated feeds rather than not to include them at all. However, it should be emphasized that authenticated feeds only qualify if credentials can be freely obtained. 

offtopic: Imo, the validation URL could be omitted. The validation URL can (for non-auth feeds) be directly derived from the feed URL. So there's no additional value in that field. Otoh there's potential for mismatches: What should be done if the feed and the validation URL don't correspond with each other? IIRC, we already had some of these problems in the past. A successful validation should be required for inclusion in `systems.csv` but the validation URL itself should not be included.

richfab:
Hello @mplsmitch,
Thank you for your comment. I understand your concern. We listened to the opinions of the GBFS consumer community to make this decision. The validation report column has been removed from systems.csv to avoid the case you mentioned and because it can be derived from the feed URL. Rest assured that we will adjust if we observe producers moving away from the open and public format encouraged by GBFS.
We are happy to listen if you have any other questions or comments.
Thank you!

PR #570, opened 2023-11-16
Remove Validation Report column

richfab:
## Problem
As mentioned in https://github.com/MobilityData/gbfs/pull/567, the validation URL will not work in cases where auth is required.

## Solution
This PR removes the Validation Report column from systems.csv.

- The validation URL can (for non-auth feeds) be directly derived from the feed URL.
- It makes the file less complex so it lowers the bar for contributions.

Screenshot 2023-11-16 at 15 39 34


josee-sabourin (changes requested):
We will need to also update the readme to remove validation report from the systems.csv blurb

josee-sabourin (approved):

PR #571, opened 2023-11-17
Request to update systems.csv

iniad-bessho:
We are working for Association for Open Data of Public Transportation in Japan.
We'd like you to add the following one feed.

- docomo bike share service

Last year, we released data on the docomo bike share service in Tokyo area.
We have recently released a feed that covers areas outside of Tokyo, so we would like you to add this.
(Old feed will also be retained for compatibility.)


richfab (approved):
Hello @iniad-bessho,
Thank you for this contribution.
It's great to see that you now cover areas outside of Tokyo!
I will merge now.
Best,
Fabien

Issue #520, opened 2023-07-12
manifest.json: include region


futuretap:
### **What is the proposal?**
The introduction of `manifest.json` greatly facilitates the distribution of feed changes. It would be even more useful if a course geographic specification of the feed area would be specified for each feed. This would allow consumers to select relevant feeds according to the location of the user.

The geographic area could be specified by an array of lat/lon rectangles (`areas` in the example). In most cases, one rectangle per feed should suffice. Multiple rectangles could be useful for disjunct operating areas in a single feed.

Optionally, specifying an ISO 3166-1 alpha-2 `country_code` could help selecting relevant feeds per country.

Example:
```
"data":{
  "datasets":[
    {
      "system_id":"flamingo_wellington",
      "versions":[
        {
          "version":"2.3",
          "url":"https://data.rideflamingo.com/gbfs/wellington/gbfs.json",
        },
        {
          "version":"3.0-RC",
          "url":"https://data.rideflamingo.com/gbfs/3/wellington/gbfs.json",
        }
      ],
      "areas": [
        {
          "north": -41.20,
          "east": 174.86,
          "south": -41.36,
          "west": 174.71
        }
      ],
      "country_code": "nz"
    }
    …
  ]
}
```

### **Alternative solutions**

In theory, this information could also be retrieved by fetching `geofencing_zones.json` (if existing). However, the rule mechanism makes it relatively hard to determine if a feed is "relevant" for a given coordinate and it requires to fetch 2 feeds (`gbfs.json` and  `geofencing_zones.json`) per tested region. The idea really is to specify only as much information on the manifest level as needed to filter relevant feeds. This would help reduce unnecessary fetching of sub-feeds.

### **Is your potential solution a breaking change?**
- [ ] Yes 
- [x] No, since `manifest.json` is new in 3.0.
- [ ] Unsure

### **Which files are affected by this change?**

`manifest.json`

benwedge:
I have a few concerns with this idea:

1. There was extensive discussion about the geofence rule objects being introduced. In short - for many users of GBFS there is no "area" so to speak. You could rent a car in France and drive to China if the fleet's rules permit it. To the extent that an area should be defined, it is likely for the purposes of the `ride_start_allowed` rule. This brings me to my second point:
2. There are different classes of feed consumer. The area for a MaaS app would be any area where the vehicle can be used, while some mapping applications will only care about the start position of the vehicles.
[...]

PR #572, opened 2023-12-04
manifest.json: include area, country_code

futuretap:
As described and discussed in #520, this PR adds new `area` and `country_code` fields allowing consumers to select relevant feeds according to the user location. `area` is proposed as a GeoJSON MultiPolygon (instead of plain rectangles) since most stake holders seemed to prefer this format.

Add `area` and `country_code` to `manifest.json`.

- [ ] Yes 
- [x] No
- [ ] Unsure

`manifest.json`


gbfs.md
```diff
@@ -316,7 +316,9 @@ Field Name | REQUIRED | Type | Defines
 \-&nbsp;`versions` | Yes | Array | Contains one object, as defined below, for each of the available versions of a feed. The array MUST be sorted by increasing MAJOR and MINOR version number. 
 &emsp;&emsp;\-`version` | Yes | String | The semantic version of the feed in the form `X.Y`.                               
 &emsp;&emsp;\-`url` | Yes  | URL | URL of the corresponding `gbfs.json` endpoint.
- 
+&emsp;&emsp;\-`area` | OPTIONAL | GeoJSON MultiPolygon | A GeoJSON MultiPolygon that describes the operating area. If `area` is supplied, then the record defines the area in which vehicles can be rented and returned.
+&emsp;&emsp;\-`country_code` | OPTIONAL | String | The country code of the operating area. The field MUST not be specified if the operating area spans multiple countries.
```
richfab: Tiny edit:

```suggestion
&emsp;&emsp;\-`country_code` | OPTIONAL | String | The country code of the operating area. The field MUST NOT be specified if the operating area spans multiple countries.
```

mplsmitch: I'm not an expert on validation but it seems like not specifying a code would make things difficult. To validate an empty `country_code` you would need to verify that `area` does in fact span international borders. Maybe it would be better to require the use of a user-assigned 3166 alpa-2 code. We could choose one from the list and require it's use for any case where a system spanned international borders.  Details on how that works here: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#User-assigned_code_elements

futuretap: Both fields are OPTIONAL so I don’t see an issue if one of the fields is left out. Imo, an empty country code should be avoided. Instead it should be left out. 

We could also allow multi-country polygons together with a `country_code`. In this case, both conditions have to be met, so the operating area equals the multi polygon intersected with the country area. Would you prefer this? It’s really an edge case that should rarely ever occur but the language should of course be clear.

mplsmitch:
> Both fields are OPTIONAL so I don’t see an issue if one of the fields is left out. Imo, an empty country code should be avoided. Instead it should be left out.

I totally agree.  What I'm suggesting is that if I'm publishing a `manifest.json` list that includes a number of systems, each with it's own `country_code`, it might make sense to have a reserved code, say `XX`. This would be used to indicate a system that spans multiple countries in stead of having to write an exception for that one list item where the `country_code` field should be left out. That way we would avoid the possibility of an empty `country_code` and the file would still be valid.

> We could also allow multi-country polygons together with a country_code.

As `area` is currently defined, I would assume this would already be allowed. There's nothing to indicate that `area` should be limited to a single country.  So then the question is, if you have a multi-country polygon, which country code would you use with it?  

This is all very much an edge case so I'm fine with leaving it as it's currently written. If it ever comes up in the future we can easily fix it.

futuretap:
> What I'm suggesting is that if I'm publishing a `manifest.json` list that includes a number of systems, each with it's own `country_code`, it might make sense to have a reserved code, say `XX`. This would be used to indicate a system that spans multiple countries in stead of having to write an exception for that one list item where the `country_code` field should be left out. That way we would avoid the possibility of an empty `country_code` and the file would still be valid.

We could allow JSON `null` for the `country_code`. I assume most parser implementations would resolve this to the same as leaving it out. Would be cleaner imo than specifying `XX` or `multi` or something similar.

richfab:
A potential solution to avoid the conditional requirement could be to name the field `country_codes` and make it an array. So, for a system in a single country we would have "country_codes": ["DE"] and for a system that covers several countries "country_codes": ["DE", "FR"]. Looking forward to hearing your opinion.

futuretap:
I’m open about making it an array. The only gripe I have is when a system is worldwide or covers a very large number of countries (are there such systems?).
Not sure though if it’s actually a problem. In `systems.csv` we have a single country code for ages and it seems to work well.

richfab:
I hereby call a vote on this proposal. Voting will be open for 10 full calendar days until 11:59PM UTC on Monday, January 15, 2024.

richfab (approved):

tdelmas:
+1 From Fluctuo

richfab:
Voting on this PR closes in 3 calendar days, at 11:59PM UTC on Monday, January 15, 2024. Please vote for or against the proposal, and include the organization for which you are voting in your comment.

BredeD:
+1 from Entur

richfab:
The vote remains open until quorum is met. Tagging recent contributors for visibility: @testower, @jkurzanski, @PierrickP, @cmonagle, @simonsolnes, @ezmckinn, @AntoineAugusti, @Cj-Malone, @ArashMansouri, @bdferris-v2, @HannesOlszewski.

Thank you for your involvement in the spec 🙏

gbfs.md
```diff
@@ -326,7 +326,9 @@ Field Name | REQUIRED | Type | Defines
 `datasets[].versions` | Yes | Array&lt;Object&gt; | Contains one object for each of the available versions of a feed. The array MUST be sorted by increasing MAJOR and MINOR version number. 
 `datasets[].versions[].version` | Yes | String | The semantic version of the feed in the form `X.Y`.                               
 `datasets[].versions[].url` | Yes  | URL | URL of the corresponding `gbfs.json` endpoint.
- 
+`datasets[].area` | OPTIONAL | GeoJSON MultiPolygon | A GeoJSON MultiPolygon that describes the operating area. If `area` is supplied, then the record defines the area in which vehicles can be rented and returned.
+`datasets[].country_code` | OPTIONAL | String | The country code of the operating area. The field MUST NOT be specified if the operating area spans multiple countries.
```
bdferris-v2: Should the field type be "Country Code" to match the convention of the existing `vehicle_types[].eco_labels[].country_code` field?

gbfs.md
```diff
@@ -326,7 +326,9 @@ Field Name | REQUIRED | Type | Defines
 `datasets[].versions` | Yes | Array&lt;Object&gt; | Contains one object for each of the available versions of a feed. The array MUST be sorted by increasing MAJOR and MINOR version number. 
 `datasets[].versions[].version` | Yes | String | The semantic version of the feed in the form `X.Y`.                               
 `datasets[].versions[].url` | Yes  | URL | URL of the corresponding `gbfs.json` endpoint.
- 
+`datasets[].area` | OPTIONAL | GeoJSON MultiPolygon | A GeoJSON MultiPolygon that describes the operating area. If `area` is supplied, then the record defines the area in which vehicles can be rented and returned.
```
bdferris-v2: the record defines the area in which vehicles can be rented and returned

Is intended that this area be used as an additional geofencing mechanism?  Or is it only descriptive?

gbfs.md
futuretap: Yes, it should be the country code in ISO 3166-1 alpha-2 notation that is used in `systems.csv`. 

A vehicle type could be reused by multiple systems and there is a list of eco labels. So in theory, a vehicle type could list all eco labels of multiple countries, even though the system is provided for a particular city only. At least that’s how I'd interpret the spec.

bdferris-v2: So what I'm specifically saying: can you change `String` here to `Country Code` instead?

gbfs.md
futuretap: I don’t see it as an additional geofencing mechanism. It should rather describe the maximum operating area where stations are located and vehicles can be rented or returned. The geofencing mechanism may then provide more fine-grained areas including slow zones, no-parking zones etc.
It is also optional which would make sense for (the unlikely case of) multi-country systems.

gbfs.md
futuretap: Yes that makes sense, thanks. Pushed [776d780](https://github.com/MobilityData/gbfs/pull/572/commits/776d7808b91306b5932a60271e92fcf269b0203d).

gbfs.md
bdferris-v2: My main point is that the text as-is written seems slightly ambiguous to me, in that it's not clear if the area is purely descriptive or if it should be used in controlling operations in some way.  For example, if a referenced system had stations or geofenced operating regions outside the bounds of the specified area, would that be considered an error?

tdelmas: Suggestion:

> If `area` is supplied, then the record defines the area that covers where vehicles can be rented, returned or used. This is for discovery and not geofencing. Ex. : if vehicles can be rented and renturned only in a city but used anywhere, that area can be the city. On the contrary, if they can only be used inside a city limit but rented in specific parkings, that area could also be the city.

futuretap: Thanks for the suggestion, sounds good to me. Does that address your concerns @bdferris-v2?

bdferris-v2: I'd propose something like:

> If `area` is supplied, then the record describes the general operating area of the system for the purpose of discovery.  Geographic details of the system's operating restrictions must be explicitly specified using station locations and geofencing zones, where appropriate.

futuretap: Thanks, using that language.

bdferris-v2:
I appreciate the updates based on my feedback.  Unfortunately, Google is a +0 on this proposal.  Which is to say, I don't think we have an specific qualms with it (especially after the updates), but we don't anticipate using it either.  So deferring votes to those who actually would.

richfab:
Given that there are at least 1 vote each from a producer (ENTUR) and a consumer (Fluctuo), the third vote can come from the Advocate (@futuretap).

[reference](https://github.com/MobilityData/gbfs/blob/master/governance.md#change-process)

futuretap:
Thanks, so +1 from Where To? / FutureTap.

richfab:
With some delay on my side, this vote has now closed, and it passes!

Votes in favor:
Fluctuo (consumer)
Entur (producer)
Where To? (consumer)
 
There were no votes against.

This change will be part of v3.1-RC planned to be released in May, as per the version release cycle in the [governance](https://github.com/MobilityData/gbfs/blob/master/governance.md#version-release-cycles).

Thank you for your involvement in the GBFS spec 🙏

PR #574, opened 2023-12-08
Fix-examples

tdelmas:

Fix the examples to makes them valid JSON and conform to the GBFS specification.

- [ ] Yes 
- [x] No
- [ ] Unsure

Only the examples.


richfab:
Great catches @tdelmas! Thank you for this contribution.

I noticed that we missed a few more typos on our side. 

I created a new PR (https://github.com/MobilityData/gbfs/pull/575) including your changes so that you don't need to spend more time on these typos.

Thank you!

PR #576, opened 2023-12-09
Improve readability

tdelmas:

This PR tries to improve the readability of some points:
- Objects and array of objects, which are too often described as "An array of objects, each containing the keys below."
- REQUIRED field inside OPTIONAL objects, which are marked "Conditionally REQUIRED" with the description "REQUIRED if `(name of the object)` is defined". 

Those make the specification too much verbose, thus blurs the important things. For example, when a field is "Conditionally REQUIRED", it's important to clearly understand why, and having too many fields defined as such without a reason make the specification unnecessarily complicated

- Describe the type object, to say once at the top that the keys are defined below.
- Clarify that a REQUIRED field in an OPTIONAL object is only required when the object is present

- [ ] Yes 
- [x] No
- [ ] Unsure


tdelmas:
Ping @richfab : Can I have your opinion on those changes?

gbfs.md
```diff
@@ -276,7 +277,7 @@ The `gbfs.json` discovery file SHOULD represent a single system or geographic ar
 
 Field Name | REQUIRED | Type | Defines
 ---|---|---|---
-`feeds` | Yes | Array | An array of all of the feeds that are published by this auto-discovery file. Each element in the array is an object with the keys below.
+`feeds` | Yes | Array<Object> | An array of all of the feeds that are published by this auto-discovery file.
```
richfab: When the draft PR becomes a Pull Request, let's please replace `<` and `>` with `&lt;` and `&gt;` for display.

```suggestion
`feeds` | Yes | Array&lt;Object&gt; | An array of all of the feeds that are published by this auto-discovery file.
```

gbfs.md
```diff
@@ -311,9 +312,9 @@ The following fields are all attributes within the main `data` object for this f
 
 Field Name | REQUIRED | Type | Defines
 ---|---|---|---
-`datasets` | Yes | Array | An array of objects, each containing the keys below.
+`datasets` | Yes | Array<Object> |
```
richfab: I prefer when all fields have a definition, even though it can be redundant for the fields which are containers for objects (this is a personal preference).

Curious to know what @josee-sabourin and others think.

```suggestion
`datasets` | Yes | Array<Object> | An array of `datasets` produced by a publisher.
```

gbfs.md
```diff
@@ -396,7 +397,7 @@ The following fields are all attributes within the main `data` object for this f
 Field Name | REQUIRED | Type | Defines
 ---|---|---|---
 `system_id` | Yes | ID | This is a globally unique identifier for the vehicle share system. Each distinct system or geographic area in which vehicles are operated MUST have its own unique `system_id`. It is up to the publisher of the feed to guarantee uniqueness and MUST be checked against existing `system_id` fields in  [systems.csv](https://github.com/MobilityData/gbfs/blob/master/systems.csv) to ensure this. This value is intended to remain the same over the life of the system. <br><br> System IDs SHOULD be recognizable as belonging to a particular system as opposed to random strings - for example, `bcycle_austin` or `biketown_pdx`.
-`languages` <br/>*(added of v3.0-RC)* | Yes | Array | List of languages used in translated strings. Each element in the list must be of type Language.
+`languages` <br/>*(added of v3.0-RC)* | Yes | Array<Language> | List of languages used in translated strings.
```
richfab: I think that making the type contained in the arrays is a good idea (eg: `Array<Object>`, `Array<Language>`, etc) 👍

gbfs.md
```diff
@@ -414,23 +415,23 @@ Field Name | REQUIRED | Type | Defines
...
 \- `brand_image_url_dark` <br/>*(added in v2.3)*  | OPTIONAL |  URL | A fully qualified URL pointing to the location of a graphic file representing the brand for the service for use in dark mode applications.  File MUST be in SVG V1.1 format and MUST be either square or round.
 \- `color` <br/>*(added in v2.3)*  | OPTIONAL |  String |  Color used to represent the brand for the service expressed as a 6 digit hexadecimal color code in the form #000000.
 `terms_url` <br/>*(as of v3.0-RC)* | OPTIONAL | Array&lt;Localized String&gt; | A fully qualified URL pointing to the terms of service (also often called "terms of use" or "terms and conditions") for the service.
-`terms_last_updated` <br/>*(added in v2.3)* |Conditionally REQUIRED | Date | REQUIRED if `terms_url` is defined. The date that the terms of service provided at `terms_url` were last updated. 
+`terms_last_updated` <br/>*(added in v2.3)* | REQUIRED | Date | The date that the terms of service provided at `terms_url` were last updated.
```
richfab, line 436: I like that the proposed solution is more compact and less verbose. I am slightly concerned that the proposed solution may be less obvious for the newcomers to the spec.

Curious to know what @josee-sabourin and others think.

richfab (commented):
Thank you @tdelmas for this contribution! We all love precise and concise documentation 💯 

I gave my personal opinion as requested. Curious to know what others think!

gbfs.md
tdelmas: > I prefer when all fields have a definition

I agree, as long as it does add context/information vs "Array of datasets". So I do like your suggestion here.

gbfs.md
tdelmas: I have the opposite feeling : conditionals are hard to read, and having a required field in an optional object is, in my opinion, easier to understand, especially for newcomers. But I want to see if others are feeling something else.

gbfs.md
tdelmas: I've push a commit in that direction: https://github.com/MobilityData/gbfs/pull/576/commits/5d6ad389a1be917eb4215e75a5847bb77f52f74a

gbfs.md
tdelmas: I think this could be more (in this PR or in a follow-up one), I didn't do it yet to make this objective of this PR more visible.

gbfs.md
tdelmas: Fixed by https://github.com/MobilityData/gbfs/pull/576/commits/0d45ea25dcaaca351cd474a01cf787652655f726

gbfs.md
richfab: Tagging recent contributors to get your opinion: @mplsmitch, @testower, @cmonagle, @simonsolnes, @ezmckinn, @AntoineAugusti, @ArashMansouri, @jkurzanski.

What is your view on the proposal to make the specification less verbose for REQUIRED fields inside OPTIONAL objects?

Before | After
-- | --
Conditionally REQUIRED with the mention "REQUIRED if object is defined" | REQUIRED

testower: I agree with @tdelmas. If the object itself is optional, there's no need to repeat that information on each property of the object.

josee-sabourin:
My two cents: 
Personally, I prefer the more explicit version of Conditionally Required / Required if simply because I don't think the "-" before what is contained in an object, or the tab for subsequent objects within that object are super clear.  See image below: 
Screenshot 2023-12-20 at 12 59 42 PM

If we can make these dashes more explicit, then I would be in favour of this proposal.

gbfs.md
futuretap: +1 for the changes!

tdelmas:
> If we can make these dashes more explicit, than I would be in favour of this proposal.

@josee-sabourin  I 100% agree with that analysis. Any suggestion to make it clearer? Maybe one dash per level?

```md

|----------------|---|---|
|`rental_apps`   |   |   |
| - `android`    |   |   |
| - - `store_uri`|   |   |
```

josee-sabourin:
A dash per level would be more clear. We could also amend the Object field definition to say "The key value pairs contained within an object are identified by a dash (-). In the instance of an Object within an Object, an extra dash (-) is present."? 

The last sentence needs some more work. Open to suggestions.

mplsmitch:
1. the definition for `object` at line 191 should be combined with, or replace the existing `object` definition in the alphabetized list at line 225
2. I'm all for improved readability but I agree with the 2 cents from @josee-sabourin. I think it's pretty easy to misinterpret the "-" and tabs currently used in the Field Name column.  One dash per level is an improvement. Another option would be to use the format that Google uses in their [GBFS definitions](https://developers.google.com/micromobility/reference/gbfs-definitions), for example :
`plans`
`plans[].per_km_pricing`
 `plans[].per_km_pricing[].start`

richfab:
+1 for using the format that Google uses in their [GBFS definitions](https://developers.google.com/micromobility/reference/gbfs-definitions).

richfab:
Closing in favor of https://github.com/MobilityData/gbfs/pull/583

PR #585, opened 2024-01-05
gbfs.md: Fix language about iOS links

futuretap:
The sentence `Please use iOS Universal Links if possible so viewing apps do not need to manually manage the redirect of the user to the app store if the user does not have the application installed` in `gbfs.md` is not correct. Universal Links do not fallback to the App Store if the app is not installed. Instead they open the website in the browser. While that website could in turn redirect to the App Store, this is not a feature of Universal Links per se. 

So I recommend writing: "Please use iOS Universal Links if possible so viewing apps provide a fallback web experience if the application is not installed."


richfab (approved):
Thank you @futuretap for this important precision.
Merging now 🚀

PR #587, opened 2024-01-10
Location formatting recommendation in systems.csv

richfab:
This PR adds a recommendation in README.md about the formatting of the system location in systems.csv:
>Primary city in which the system is located, followed by the 2-letter state code for US systems. The location name SHOULD be in English if the location has an English name (eg: `Brussels`).


futuretap:
Isn't the country code redundant, since it’s already in the first column? For US systems, the state is currently used behind the city name which we should keep imo.

richfab:
@josee-sabourin Is there a reason for adding the 2-letter country code after the city name, since it's already present in the country column?

josee-sabourin:
I don't see the added value, no

richfab:
Thank you @futuretap for your feedback.

I updated the location formatting recommendation as suggested and removed the country codes from systems.csv.

PR is ready for review. Thank you!

josee-sabourin (approved):

PR #588, opened 2024-01-11
systems.csv: Removed nextbike_fo from systems.csv

merjakaj:
Removed nextbike_fo as the bike share system in Oulu, Finland has not been in use since 2019.


richfab (approved):
Happy first PR to this repo @merjakaj! 🎉 
And thank you for keeping the catalog up to date 💯

PR #590, opened 2024-01-11
systems.csv: Updated location names in systems.csv

merjakaj:
Updated locations names:
- If the city has an English name, use it
- Replaced some location names with the actual location 
- "Multiple cities" replaced with the country name


richfab (approved):
LGTM 👍 Merging now 🚀 
Thank you @merjakaj! 🙏

PR #591, opened 2024-01-12
systems.csv: Updated JP location name

merjakaj:
Updated JP location name in systems.csv


richfab:
Hi @merjakaj,
I would suggest to keep Japan as the location as the geographic area of this feed is spread across the country. See [this visualization](https://transport.data.gouv.fr/tools/gbfs/analyze?url=https%3A%2F%2Fapi-public.odpt.org%2Fapi%2Fv4%2Fgbfs%2Fhellocycling%2Fgbfs.json) (the map takes a minute to load).
Thank you!

image

PR #593, opened 2024-01-22
Syncs systems.csv with transport.data.gouv

richfab:
This PR syncs systems.csv with GBFS data sources from [transport.data.gouv.fr](https://transport.data.gouv.fr/datasets?q=gbfs).

- Removed Bird Marseille ([ceased operations](https://www.20minutes.fr/marseille/4057243-20231011-marseille-plus-deux-operateurs-trottinette-autorises))
- Added levélo Marseille
- Updated LoVélo Rouen (endpoint was system_information instead of gbfs)
- Removed Vélivert Saint Étienne duplicate and updated website URL
- Added Vélo'Baie,Saint-Brieuc


systems.csv
```diff
@@ -345,13 +344,14 @@ FR,Dott Paris,Paris,dott-paris,https://ridedott.com,https://gbfs.api.ridedott.co
 FR,Dott Tignes,Tignes,dott-tignes,https://ridedott.com,https://gbfs.api.ridedott.com/public/v2/tignes/gbfs.json,
 FR,Dott Val-d’isere,Val-d’isere,dott-val-d’iser,https://ridedott.com/,https://gbfs.api.ridedott.com/public/v2/val-d%E2%80%99isere/gbfs.json,
 FR,LE vélo STAR,Rennes,le_velo_star,https://www.star.fr/le-velo/nos-offres/vls/,https://eu.ftp.opendatasoft.com/star/gbfs/gbfs.json,
+FR,Levélo,Marseille,levelo_inurba_marseille,https://levelo.ampmetropole.fr,https://api.omega.fifteen.eu/gbfs/2.2/marseille/en/gbfs.json?&key=MjE0ZDNmMGEtNGFkZS00M2FlLWFmMWItZGNhOTZhMWQyYzM2,
```
isabelle-dr: This format with a key would be a first in systems.csv I believe

isabelle-dr: According to the files listed, the auto-discovery URL is https://api.omega.fifteen.eu/gbfs/2.2/marseille/en/gbfs.json and users need a token to access the data. Should we include the token in the last column instead?

systems.csv
```diff
@@ -374,8 +374,8 @@ FR,Tier GRENOBLE,Grenoble,tier_GRENOBLE,https://www.tier.app,https://data.mobili
 FR,VCub Bordeaux,Bordeaux,vcub,https://www.vcub.fr/,https://transport.data.gouv.fr/gbfs/vcub/gbfs.json,
 FR,Vélam,Amiens,amiens,http://www.velam.amiens.fr/,https://transport.data.gouv.fr/gbfs/amiens/gbfs.json,
 FR,Vélib' Metropole,Paris,Paris,https://www.velib-metropole.fr/,https://velib-metropole-opendata.smovengo.cloud/opendata/Velib_Metropole/gbfs.json,
-FR,Vélivert,Saint-Etienne,Velivert_FR_Saint-Etienne,https://www.velivert.fr/,https://saint-etienne-gbfs.klervi.net/gbfs/gbfs.json,
 FR,VélO2,Cergy-Pontoise,cergy-pontoise,http://www.velo2.cergypontoise.fr/,https://transport.data.gouv.fr/gbfs/cergy-pontoise/gbfs.json,
+FR,Vélo'Baie,Saint-Brieuc,saintbrieuc,https://www.saintbrieuc-armor-agglo.bzh/velobaie,https://gateway.prod.partners-fs37hd8.zoov.site/gbfs/2.2/saintbrieuc/en/gbfs.json?key=YmE1ZDVlNDYtMGIwNy00MGEyLWIxZWYtNGEwOGQ4NTYxNTYz,
```
isabelle-dr: Same comment about the tokens

isabelle-dr (commented):
Two of these feeds need a token, should we remove the key from the URL and provide it in the last column?

systems.csv
richfab: Currently the column `Authentication Info URL` is to provide a **URL** to a human-readable page describing how the authentication should be performed ([README.md](https://github.com/MobilityData/gbfs?tab=readme-ov-file#systems-catalog---systems-implementing-gbfs)). 

I could not find a page explaining how to retrieve the token. This public token is directly appended to the URL on [transport.data.gouv](https://transport.data.gouv.fr/datasets/velos-a-assistance-electrique-en-libre-service-levelo-sur-marseille) and on [ampmetropole.opendatasoft.com](https://ampmetropole.opendatasoft.com/explore/dataset/gbfs-donnees-gbfs-levelo-a-marseille/api/).

I suggest to rename the column `Authentication Info URL` to `Authentication Info`, update the description in the [README.md](https://github.com/MobilityData/gbfs?tab=readme-ov-file#systems-catalog---systems-implementing-gbfs) and move the token from the URL to the `Authentication Info` column.

richfab: I added commits to this PR to make the change.

README.md
```diff
@@ -126,7 +126,7 @@ Location | Yes| Primary city in which the system is located, followed by the 2-l
 System ID | Yes | ID for the system. This MUST match the `system_id` field in `system_information.json`.
 URL | Yes | URL for the system from the `url` field in `system_information.json`. If the `url` field is not included in `system_information.json` this SHOULD be the primary URL for the system operator.
 Auto-Discovery URL | Yes | URL for the system's `gbfs.json` auto-discovery file.
-Authentication Info | Conditionally REQUIRED | If authentication is required, this MUST contain a URL to a human-readable page describing how the authentication should be performed and how credentials can be created, or directly contain the public key-value pair to be added to the URL.
+Authentication Info | Conditionally REQUIRED | If authentication is required, this MUST contain a URL to a human-readable page describing how the authentication should be performed and how credentials can be created, or directly contain the public key-value pair to append to the feed URLs.
```
mplsmitch, line 129: This wouldn't be a key-value pair. I think what you want to say here is "contain the API key or token to append to the feed URLs"

richfab: Thanks. How can we make explicit that the name of the parameter is required too?

mplsmitch: I think this is getting overly complicated - I don't see any reason to separate the key from the URL in the auto-discovery column. What's the value in that? We've got lots of URLS with parameters for example: `https://qickscooters.rideatom.com/gbfs/v2_2/en/gbfs?id=104` the only difference is that in this case it's an API key not a `system_id` but they function in the same way since the API key doesn't change. It's my understanding that the last column was added to support Auth schemes that use tokens or other forms of Auth that are subject to change and therefore needed a URL to provide additional info on how to get access.

futuretap: The problem here is that `gbfs.json` references the sub-feeds without the key. It needs to be appended manually to them. So I suggest keeping the "key value pair" language.

mplsmitch: I'm sorry - I misunderstood the problem. Ortwin is correct. I didn't realize that the URLS in `gbfs.json`  lacked the key. However I would also argue that this is broken because  `feeds[].url` in `gbfs.json` should contain a valid URL pointing to the file. Maybe we should consider updating that definition and requiring valid URLs. I've seen this lots of times the past and I just left the feeds out of `systems.csv` until they fixed the `gbfs.json` file.

richfab: Thank you both for your input!

>I would also argue that this is broken because `feeds[].url` in `gbfs.json` should contain a valid URL pointing to the file.

I agree. I informed the producer that it would be best to remove the public key or append it to all URLs in `gbfs.json`.

>Maybe we should consider updating that definition and requiring valid URLs.

+1

isabelle-dr (approved):
LGTM, thanks for the input @mplsmitch

PR #623, opened 2024-04-10
Correction systems.csv

mplsmitch:
File is not displaying correctly - missing csv comma line 816


richfab (approved):
LGTM
Thank you for checking @mplsmitch 🙏

PR #632, opened 2024-04-29
Clarify vehicle_id persistence rule

richfab:
This PR:
- Makes the rule about vehicle_id persistence more explicit: rotation is between rentals (not every time the feed is loaded)
- Updates a forgotten reference to free_bike_status.json (now vehicle_status.json)

Thanks @testower for flagging!


isabelle-dr (approved):
LGTM!

PR #644, opened 2024-05-28
Remove an extra space in systems.csv

richfab:
This PR removes an extra space in systems.csv which caused the failure of the workflow `[MobilityData/gbfs] Node.js CI`.

Thanks to @Alessandro100 for noticing this 🙏 

Before | After
-- | --
image | image


Alessandro100 (approved):
😎

PR #646, opened 2024-06-01
Fix "Detroit" spelling.

mjmsmith:
s/Detriot/Detroit


richfab (approved):
Good catch! Thank you for the fix @mjmsmith 🙏

PR #649, opened 2024-06-11
Ming: Clean up the unavailable GBFS feeds

gongmingqm10:
Remove some GBFS feeds for Neuron Mobility. Neuron Mobility GBFS feeds have been under the abnormal traffic attack in the last couple months. And as a result, we have to stop the public GBFS for now. For accessing some of the data, we will need to use a simple API token to temporarily give access with.


richfab (approved):
LGTM
Sorry to hear that Neuron Mobility GBFS feeds have been under abnormal traffic attack.
I hope this gets resolved and that you can add the feeds back in the catalog.
Thank you 🙏

PR #666, opened 2024-07-22
De-duplicate invalid `System ID` in `systems.csv`

richfab:
## Context
This PR is in preparation of https://github.com/MobilityData/gbfs/issues/664 to add a column named `ID` in `systems.csv`.

## What's Changed
- 7 edits: Updated the value of `System ID` in `systems.csv` for 7 feeds from LINK to match the value of `system_id` in `system_information.json` in the corresponding GBFS feed.
- 16 deletions:
    - Removed 15 LINK systems as they shutdown operations in the U.S. 
    - Removed a duplicate feed for Bicing Barcelona. Chose to keep the one with the correct value in the 'Name' column. The content of both feeds is identical. Only the URL slightly changes (both are from PBSC).


mplsmitch:
As far as I know all of the US based LINK systems shutdown on Dec 31st, 2023 - unfortunately they did not stop publishing so it looks like these are now zombie feeds. When  I look at `free_bike_status`, they're empty or nearly empty.  I think this may also be true of the  European LINK systems.

richfab:
Thank you @mplsmitch. I removed all LINK feeds in the US.

Shutdown of LINK operations in the U.S. on Dec 31 2024 ([TechCrunch](https://techcrunch.com/2023/12/15/scooter-startup-superpedestrian-shutting-down-us-operations-explores-sale-of-europe-business/)).

isabelle-dr (approved):
lgtm!

PR #688, opened 2024-10-08
Deprecation of v1.1

mplsmitch:
Ending support for v1.1, 180 days after the official release of v3.0

 From the documentation:
 

> Supported versions SHALL NOT span more than two MAJOR versions. Past versions that are beyond the two most recent MAJOR versions will be deprecated 180 days after the latest MAJOR version becomes official.

It's now been 180 days since v3.0 became the official release so v1.1 should be deprecated.

- [x ] No

README and version v1.1


richfab (approved):
LGTM. Thank you @mplsmitch!

PR #695, opened 2024-10-21
Remove duplicate feed in systems.csv for Bergen City Bike

richfab:
## Problem
There is a duplicate GBFS feed in [systems.csv](https://github.com/MobilityData/gbfs/blob/master/systems.csv) for Bergen City Bike.

## What's Changed
Removed the feed from the aggregator (Entur) to keep only the upstream feed from the operator (Urban Sharing).

Thank you @merjakaj for raising this issue 🙏


isabelle-dr (approved):
LGTM

PR #699, opened 2024-10-23
Best practice regarding URLs and version upgrades

### LINKED ISSUE #655: Good practice regarding URLs and version upgrades
Hello,

We're wondering about the recommended practice regarding GBFS versions in the URL and version upgrades.

Imagine you have `example.com/gbfs/gbfs.json` at 3.0 with `gbfs_versions` containing links to other versions (such as `example.com/gbfs/2.3/gbfs.json`).

You now upgrade your main GBFS version to 3.1. Reusers can be surprised by your move and it now breaks their app.

- Should `example.com/gbfs/gbfs.json` switch to 3.1 silently?
- Should producers always include the version in the URL (and avoid URLs like `example.com/gbfs/gbfs.json`)
- Should producers have URLs like `example.com/gbfs/latest/gbfs.json` which can lead to silent version upgrades

richfab:
Fixes https://github.com/MobilityData/gbfs/issues/655

#### **What problem does your proposal solve?**
When a GBFS producer upgrades their feed to a MAJOR version without changing the feed URL, it can break consuming applications.

Add a best practice in the spec to recommend that feed URLs contain the MAJOR version number. eg: https://www.example.com/gbfs/v3/gbfs.json

- [ ] Yes 
- [x] No
- [ ] Unsure

It does not affect the content of the files, only the URL of the files.

Thank you @AntoineAugusti for raising this issue 🙏

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -79,7 +79,11 @@ geofencing_zones.json <br/>*(added in v2.1)* | OPTIONAL | Geofencing zones and t
 
 ## Accessibility
 
-Datasets SHOULD be published at an easily accessible, public, permanent URL. (for example, https://www.example.com/gbfs/v3/gbfs.json). The URL SHOULD be directly available without requiring login to access the file to facilitate download by consuming software applications.
+Datasets SHOULD be published at an easily accessible, public, permanent URL. 
+
+The URL SHOULD contain the MAJOR version number. If upgrading to a MINOR version, the URL SHOULD NOT change. (for example, https://www.example.com/gbfs/v3/gbfs.json). 
+
+The URL SHOULD be directly available without requiring login to access the file to facilitate download by consuming software applications.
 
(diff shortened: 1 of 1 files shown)


isabelle-dr (approved):
LGTM. 

Regarding not including the minor version in the URL; I see many feeds that don't follow this in systems.csv. 
Should we do the following:
- contact them to let them know if this new Best Practice
- add this to the validator (acknowledging this might be a bit of a clunky validation)

richfab:
Good point. 10% of the feeds in systems.csv include the MINOR version number in their URL.

I added both strategies to the MobilityData product backlog. Thanks @isabelle-dr 🙏

isabelle-dr:
@richfab 
I am also wondering if this section should be renamed Dataset Publishing & General Practices
[example for GTFS](https://gtfs.org/documentation/schedule/reference/#dataset-publishing-general-practices) and maybe have it above the "Files" section to increase its visibility.

richfab:
I added this idea to the MobilityData product backlog as well. Good point! Thanks

PR #721, opened 2025-01-17
Fix requirement of lat/lon for virtual stations

### LINKED ISSUE #719: Inconsistency on virtual station
Greetings,

I noticed what seems to be an inconsistency in the definition of station_information.json.
For the field `is_virtual_station` we have: 

> `true` - The station is a location without smart docking infrastructure. the station may be defined by a point (lat/lon) **and/or** station_area (below).

But we also have for the field `station_area`:

> If lat/lon and station_area are both defined, the lat/lon is the significant coordinate of the station

But it is currently not possible to set `lat/lon` to null. This mean that even if we want to define only `station_area`, we are forced to define `lat/lon` that become the significant coordinate.

In my opinion, the solution is to be able to set `lat/lon` to null, because even if it possible to define a virtual station with only a single coordinate it should also be possible to defined a virtual station with only a set of coordinates.

You'll also need to make sure that either `lat/lon` or `station_area` is supplied and not null.

UPDATE: The consensus is to remove lat/lon OPTIONAL in `is_virtual_station` field because consumers rely on lat/lon.

- [ ] Yes 
- [x] No
- [ ] Unsure

richfab:
#### **What problem does your proposal solve?**
Fixes https://github.com/MobilityData/gbfs/issues/719

The `is_virtual_station` field indicates that `lat/lon` are OPTIONAL when the `station_area` field is defined.

It causes an inconsistency with `lat/lon` fields in the same file which are REQUIRED values.

Remove  `lat/lon` OPTIONAL in `is_virtual_station` field.

- [ ] Yes 
- [x] No
- [ ] Unsure

There is no vote needed since this is a clarification of the spec, not a change to a field.

station_information.json

Note that the [JSON schemas](https://github.com/MobilityData/gbfs-json-schema/blob/master/v3.0/station_information.json#L270) are not affected by the inconsistency. Which means they don't need to be updated and neither do the [validator](https://gbfs-validator.mobilitydata.org/) and the [models](https://github.com/MobilityData/gbfs-json-schema/tree/master/models) since they are based on the JSON schemas.

Thank you @Yaronn44 for raising this issue 🙏


davidgamez (approved):

fredericsimard (approved):
LGTM!

PR #765, opened 2025-06-05
Rename 'levélo - Marseille' and 'Citiz Métrople Lyon'

richfab:
## Context
Related to https://github.com/MobilityData/mobility-feed-api/issues/1238: When searching a GBFS feed by city, some feeds are not returning.

For example, searching for "Marseille" should return "Levélo" feed. This is a known issue in our admin levels we have for France in our database.

A closer look reveals that the correct name for "Levélo" in [system_information.json](https://gbfs.omega.fifteen.eu/gbfs/2.2/marseille/en/system_information.json) is "levélo - Marseille", which would allow the feed to be returned when searching by city.

## What's Changed
Edited 2 feeds with the correct values from their system_information.json files:

- Renamed "Levélo" feed to "levélo - Marseille" (matches value in [system_information.json](https://gbfs.omega.fifteen.eu/gbfs/2.2/marseille/en/system_information.json))
- Renamed "Citiz LPA" feed to "Citiz Métrople Lyon" and updated the `system_id` to "citiz_metropole_lyon" (matches values in [system_information.json](https://backend.citiz.fr/public/provider/3/gbfs/v3.0/system_information.json))


fredericsimard (approved):
LGTM!

PR #770, opened 2025-06-30
Fix vehicle availability example (Datetime)

richfab:
### Context
The endpoint `vehicle_availability.json` was added in [v3.1-RC2](https://github.com/MobilityData/gbfs/releases/tag/v3.1-RC2) but there was a typo in the example.

### What's Changed
Fixed the values in the example of the new endpoint `vehicle_availability.json` to be Datetime.

Before: "2024-12-24T08:15Z"
After: "2024-12-24T08:15:00Z" (added missing seconds :00 before Z)

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -1178,11 +1178,11 @@ Field Name | REQUIRED | Type | Defines
         ],
         "availabilities": [
           {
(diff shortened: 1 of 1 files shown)


fredericsimard (approved):
LGTM!

PR #794, opened 2025-08-06
Move Dott-as to Norway

evertheylen:
While there is a city in Belgium named "As", the Dott bikes are actually in Norway (also in a city named "As").


richfab (approved):
Hello @evertheylen,
Thank you for this update 🙏 Merging now.
Fabien (MobilityData)

PR #798, opened 2025-09-25
nextbike - Update systems.csv

zimmerling:
Hello, 
after a long break due to a broken automation, we are updating our endpoints again. 
As before, we have two separate commits. The first one only sorts the existing systems and the second one updates our own systems.


mplsmitch:
@zimmerling I'm seeing a lot of errors, it looks like your automation predates changes to the columns in the systems.csv file. You can find all the current field definitions [here](https://github.com/MobilityData/gbfs?tab=readme-ov-file#systems-catalog---systems-implementing-gbfs)

zimmerling:
@mplsmitch, I'm sorry. I've fixed the errors and hope everything is OK now.

mplsmitch (approved):

PR #806, opened 2025-10-21
Revert "Add Lime Seattle v1.0 feed"

sargunv:
Reverts MobilityData/gbfs#805

My bad; I saw `"version": "1.0"` in the feed and assumed it was a valid GBFS v1 feed, but it's actually not conformant to v1.

For example: https://data.lime.bike/api/partners/v1/gbfs/seattle/station_status

```jsonc
{
    "last_updated": 1761008760,
    "ttl": 0,
    "version": "1.0",
    "data": {
        "stations": [
            {
                "station_id": "seattle",
                "num_vehicles_available": 11849, // should be num_bikes_available in v1
                "num_docks_available": 999999,
                "is_installed": true, // should be 1|0 in v1
                "is_renting": true,
                "is_returning": true,
                "last_reported": 1761008760
            }
        ]
    }
}
```


mplsmitch:
Unfortunately, none of Lime's feeds follow the specification.  If you look at their v2 feeds you will see the same garbage data that's in your example in every city. They don't seem to understand that `station_information.json` and `station_status.json` should not be included in feeds from dockless systems that don't have any stations. This has been the case for years.

PR #807, opened 2025-10-21
Update opening_hours documentation

mplsmitch:
This PR corrects an inconsistency in the documentation around `opening_hours`.  @hbruch noted that in the documentation under **Hours and Dates of Operation** it says:

> Hours and dates of operation **SHOULD** be published even in cases where services are continuously available 24/7.
 
but `system_information.opening_hours` is REQUIRED so the line quoted above should be changed from SHOULD to MUST. 

This is kind of a gray area but my view is that this is an editorial change to the documentation and does not require a vote. If anyone disagrees with this we can put it to a vote. That would probably also make it a breaking change which would mean it would not take effect until v4.0.  If no one has an issue it will go in the next minor version.

- [ ] Yes 
- [ ] No 
- [ X ] Unsure

gbfs.md

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -109,7 +109,7 @@ Beginning with v3.0, hours and dates of operation are described using the Open S
 * [OSM opening_hours validation tool](https://openingh.openstreetmap.de/evaluation_tool/)
 * [OSM opening_hours project and code libraries](https://github.com/opening-hours)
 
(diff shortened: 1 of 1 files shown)


richfab:
Thank you @hbruch for raising this inconsistency in the spec.

We deemed that this is an editorial change so it does not require a vote.

I'll merge the change into master now (v3.1-RC2).

richfab (approved):
LGTM!

PR #808, opened 2025-10-21
Update link for shared mobility resources

mplsmitch:
Fixes bad link to gbfs.org/tools in the README

- [ ] Yes 
- [x] No
- [ ] Unsure

README


fredericsimard (approved):
LGTM! I also added a redirection for all 3 languages on the server itself, increases chances that users will not fall on a 404.

PR #811, opened 2025-11-12
Remove feeds from Qick & Bolt

mplsmitch:
Removes feeds from Qick and Bolt that are no longer available

- [ ] Yes 
- [x] No
- [ ] Unsure

systems.csv


fredericsimard (approved):
LGTM!

PR #817, opened 2025-12-09
Add / Update to systems.csv

mplsmitch:
Adds a number of new feeds to sytems.csv, provides alternate sources for some that were recently lost to Auth and updates some existing feeds to v3.0


fredericsimard (approved):
LGTM!

PR #819, opened 2026-01-08
chore: remove offline regions

dersvenhesse:
The dott regions of Ellwangen and Greifswald are responding with `{"error":"Forbidden: ERR_REGION_OFFLINE"}` for quite some time, hence the removal. There are also no rides listed within their own app.


mplsmitch (approved):
Looks good - Thank you!

PR #820, opened 2026-01-10
add Flamingo Christchurch (again)

applecuckoo:
they have relaunched in Christchurch since I removed them from systems.csv


mplsmitch (approved):
Looks Good - thanks!

PR #821, opened 2026-01-12
Improve performance: ETag/If-None-Match

### LINKED ISSUE #708: Improve performance: Push for the usage of ETag/If-None-Match
### What is the issue and _why_ is it an issue?

In order to reduce the load both on producers and consumers, caching mechanisms could be improved.

### Please describe some potential solutions you have considered (even if they aren’t related to GBFS).

Following the discussion during the International Mobility Data Summit (https://github.com/MobilityData/gbfs/discussions/703), I would like to suggest this addition to the specification in the "File Distribution" section:

- HTTP responses SHOULD include the [`ETag` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag).
- Clients receiving `Etag`s SHOULD use the [`If-None-Match` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) for subsequent requests.
- HTTP responses to requests including the `If-None-Match` HTTP header SHOULD respect it by returning a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) when appropriate.

- [ ] Yes 
- [X] No
- [ ] Unsure

tdelmas:

Allow consumers to easily cache anwsers with minimal burden on producers (Fix #708)

The HTTP ETag/If-None-Match mechanism

- [ ] Yes 
- [X] No
- [ ] Unsure

Not a breaking change, as the "SHOULD" requirement allow some flexibility both for producers and consumers.

### DIFF
--- a/gbfs.md
+++ b/gbfs.md
@@ -126,6 +126,9 @@ Hours and dates of operation SHOULD be published even in cases where services ar
     * All endpoints MUST use HTTPS
     * REQUIRED files MUST NOT 404. They MUST return a properly formatted JSON file as defined in [Output Format](#output-format).
     * OPTIONAL files MAY 404. A 404 of an OPTIONAL file SHOULD NOT be considered an error.
+* HTTP responses SHOULD include the [`ETag` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag).
+* Clients receiving `Etag`s SHOULD use the [`If-None-Match` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) for subsequent requests.
+* HTTP responses to requests including the `If-None-Match` HTTP header SHOULD respect it by returning a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) when appropriate.
 
 ### Version Endpoints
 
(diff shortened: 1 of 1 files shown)


gbfs.md
```diff
@@ -126,6 +126,9 @@ Hours and dates of operation SHOULD be published even in cases where services ar
     * All endpoints MUST use HTTPS
     * REQUIRED files MUST NOT 404. They MUST return a properly formatted JSON file as defined in [Output Format](#output-format).
     * OPTIONAL files MAY 404. A 404 of an OPTIONAL file SHOULD NOT be considered an error.
+* HTTP responses SHOULD include the [`ETag` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag).
+* Clients receiving `Etag`s SHOULD use the [`If-None-Match` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) for subsequent requests.
+* HTTP responses to requests including the `If-None-Match` HTTP header SHOULD respect it by returning a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) when appropriate.
```
mplsmitch: I think this would be clearer if it said  `HTTP responses to requests containing...`  in stead of `requests including`

mplsmitch (changes requested):

mplsmitch (approved):

richfab:
Thanks @tdelmas for this proposal, which follows the discussions held at the Mobility Data Summit on Oct 2024.

We deem that there is no need for a vote because this change is a best practice on the web, not specific to GBFS.

I'll merge the proposal into master now (v3.1-RC2).

richfab (approved):
LGTM!

PR #827, opened 2026-02-11
Update systems.csv: Add esel.ac

1Maxnet1:
This is a Community-operated bike sharing network in Aachen.


mplsmitch (approved):

mplsmitch:
@1Maxnet1 I'm going to merge this but it appears that this is a purely free floating system without stations. If that is the case the `station_information` and `station_status` files should not be included in this feed.
Thanks!

PR #828, opened 2026-02-19
Update systems.csv: Add MOL Bubi (Budapest)

bndeff:
This is a Nextbike system running under its own branding. It has a limited shelf life (it will be replaced by a system run by Inurba Mobility in the coming months) but while it's still operational it makes sense to have it in the feed.


mplsmitch (approved):
Looks Great - Thank you!

PR #853, opened 2026-04-20
Remove unofficial BiciMAD feed

richfab:
## Context
There are 2 data sources for BiciMAD in Madrid 🇪🇸
Having a unique data source for each system makes feed integration easier for consumers.

## What's Changed
In this PR:
- Removed unofficial BiciMAD feed


fredericsimard (approved):
All good!

PR #865, opened 2026-05-05
Add 149 feeds from Getaround in AT, BE, DE, ES and FR

richfab:
### Context
Getaround (carshare) publishes 6 manifest files for these countries: AT, BE, DE, ES, FR and NO:
ie: https://getaround.com/gbfs/manifest?country_code=AT

### What's Changed
- Added 149 feeds from Getaround in AT, BE, DE, ES and FR
- Replaced all 34 Entur URLs for Getaround Norway with the operator URLs

### Method
- Only added the feeds with at least one vehicle available (many cities in the manifests are actually empty feeds).
- Only added cities with > 50,000 population (many feeds lists only vehicles that are actually in the neighbouring city).


emmambd (approved):
LGTM! What's the rationale behind using the operator URLs directly instead of Entur's?

richfab:
Although we know Entur's high data quality standards, we apply the same rule for all feeds which is: using the operator data source is less prone to errors than relying on a third party aggregator.

PR #875, opened 2026-06-02
Remove 3 Getaround feeds with duplicate system ids

richfab:
### Context
Getaround has duplicate system_ids for cities with duplicate names in different countries. For example Beaumont, FR and Beaumont, BE.

### What's Changed
This PR removes 3 Getaround feeds that had duplicate system ids.


emmambd (approved):
LGTM!

PR #878, opened 2026-06-12
Fix Car-Ship entry

1Maxnet1:
The feed is located in Germany and the provider has it's own URL instead of the one where the feed is re-published.


richfab (approved):
LGTM!
Thanks for this fix @1Maxnet1.

PR #880, opened 2026-06-15
Fix two minor issues

1Maxnet1:
- Region of Welo is not Welo (that's the name of the service) but rather Bonn
- Region of StadtRadLüneburg is Lüneburg (typo)


richfab (approved):
LGTM
Thank you @1Maxnet1!

PR #884, opened 2026-07-07
Add 4 feeds from Dott, Ford and Lime in BE, DE and PL

richfab:
### Context
Dott, Ford and Lime published new feeds 🎉

### What's Changed
This PR adds:
- 1 feed from Dott in PL
- 2 feeds from Ford in DE
- 1 feed from Lime in BE


fredericsimard (approved):
LGTM

PR #887, opened 2026-07-14
chore: remove offline regions

dersvenhesse:
The removed dott regions are responding with `ERR_REGION_OFFLINE` or `ERR_REGION_NOT_FOUND`. There are also no rides listed within their own app.


richfab (approved):
LGTM. Thanks @dersvenhesse!

cc @miselaytes-anton for visibility

PR #889, opened 2026-07-15
chore: remove region

dersvenhesse:
I found another region after https://github.com/MobilityData/gbfs/pull/887.


richfab (approved):
Thanks @dersvenhesse! LGTM

cc @miselaytes-anton

PR #895, opened 2026-07-29
Add Lyft Bike, rebrand of the existing Bay Wheels system

fkh:
Updating feed details for Lyft Bike, which supersedes the Bay Wheels feed as of today (old feeds remain active thru end of 2026 but no need to wait on sharing the new feed with consumers).


fkh:
Requesting a review, @fredericsimard - thank you!

fredericsimard (approved):
LGTM!

PR #896, opened 2026-08-03
Add 1 feed for MOL Bubi 3.0 in Budapest, HU

bela333:
Bubi relaunched, now with Inurba instead of Nextbike


emmambd (approved):
Thanks for this submission @bela333!

PR #899, opened 2026-08-25
chore: remove offline region

dersvenhesse:
The removed dott region is responding with ERR_REGION_NOT_FOUND. There are also no rides listed within their own app.


emmambd (approved):
Thanks @dersvenhesse for this clean up!

PR #900, opened 2026-09-07
Add missing zero after version number

richfab:
### Context
Follow up of https://github.com/MobilityData/gbfs/pull/897

A few feeds added are missing a zero.

### What's Changed
Added missing zero after version number.


fredericsimard (approved):
LGTM!

PR #906, opened 2026-09-14
feat: remove obsolete dott locations in germany

LarsBehl:
# Context
The removed locations started to response with 'region not found' in the last 7 days. Since the initial 'region not found' response, they did not provide any information again.

# What's Changed
This PR removes the 13 locations which started to responde with 'region not found'.


richfab:
Thanks @LarsBehl for sharing more information.

I've asked Dott for more information about the feed authentication.

If Dott confirms that these feeds require auth, we may keep them in systems.csv with information on how to obtain credentials.

Thank you,
Fabien

